PR #2, opened 2015-12-31
use 1.x version on draft spec so as not to confuse

jeffhammond:
The current draft is certainly not the 1.1 spec, so it should not have version 1.1 on the cover.


jeffhammond:
Sorry. It seems you are keeping the 1.1 spec on master.  Please don't do that.  No one uses master for the oldest version of whatever they have in version control.  Please create a git tag for the 1.1 specification and put the latest working version on master.  We should only have a 1.3 branch on it is in semifinal draft form and we only intend to commit bugfixes to it.

PR #3, opened 2016-06-01
Update volatile qualifiers in wait and lock API

jdinan:
Ref: redmine ticket #211 (http://www.openshmem.org/redmine/issues/211).

### DIFF
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -351,6 +351,17 @@ specification.
 
 \chapter{Changes to this Document}\label{sec:changelog}
 
+\section{Version 1.4}
+
+\begin{itemize}
+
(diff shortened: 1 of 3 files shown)


jdinan:
Please re-open.  The reading was successful and this proposal is moving forward for a vote.  It's better to ask the owner of the PR to close it, if you think it's no longer needed.

dkhaldi:
Jim,
I think this happened after the deletion of the branches which I did few minutes ago. I just restored the branches since apparently, there are pull requests that are still related to them which I don't understand since the branches are supposed to be released. No development or PR should be attached to 1.3, right?

jdinan:
Oh, I see what happened.  Sorry, the closure message was confusing.  Please go ahead and delete the 1.2/1.3 branches.  I'll fix the pull request on my end.

PR #5, opened 2016-07-18
Remove volatile qualifiers in wait and lock APIs

jdinan:
Alternative to #3 that completely removes the volatile qualifiers added in OpenSHMEM 1.3


jdinan:
Re-opening against osh_spec_next

PR #9, opened 2016-08-16
Fix duplicate ref and add reference from text

jdinan:


jdinan:
I noticed another issue of the refs, beyond what is fixed here.  In Annex A, the references from the text refer to all of the Listings (A.1-A.1) as "Listing A".  Does anyone know how to correct this, so we get text like "see Listing A.3" instead of "see Listing A"?  "Listing A" is not specific enough to guide a reader to the right listing.

dkhaldi:
The refs that are not printed correctly (points to Annex instead of the actual listing) is because there is no caption there. We need to add a caption in order to fix this (eg. add a caption before label of "label=openshmem-hello-f90").

swpoole:
We can certainly create a proposal for such a thing.

Steve...

On 8/16/16 08:35, Dounia Khaldi wrote:
> The refs that are not printed correctly (points to Annex instead of the actual listing) is because there is no caption there. We need to add a caption in order to fix this (eg. add a caption before label of "label=openshmem-hello-f90"). 
>

-- 
Regards,
Steve...

"Don't mistake my kindness for weakness. I am kind to everyone,
but when someone is unkind to me, weak is not what you are going
to remember about me." -- Al Capone.

jdinan:
Thanks, @dkhaldi!  I added a commit to fix the captions and the references are now generated correctly.  Typically, changes like this would be made by the section committees and rolled up into a single vote during final preparation.  I think what we may need to do, because we are making these corrections early in the process, is to push this through the voting process.

There's a discussion on the mailing list about how to incorporate these changes, so let's wait to see what conclusion we reach over there.

swpoole:
Agreed. Until we can get the section heads in place, we should all vote
on these items.

Steve...

On 8/16/16 09:10, James Dinan wrote:
> Thanks, @dkhaldi!  I added a commit to fix the captions and the references are now generated correctly.  Typically, changes like this would be made by the section committees and rolled up into a single vote during final preparation.  I think what we may need to do, because we are making these corrections early in the process, is to push this through the voting process.
>
> There's a discussion on the mailing list about how to incorporate these changes, so let's wait to see what conclusion we reach over there.
>

-- 
Regards,
Steve...

"Don't mistake my kindness for weakness. I am kind to everyone,
but when someone is unkind to me, weak is not what you are going
to remember about me." -- Al Capone.

dkhaldi:
@jdinan Sounds good to me.

jdinan:
@swpoole, bear in mind that there is still voting required for the work done by the section committees and this this only occurs during the release process.  The release process is quite heavyweight, so it's unlikely to be a good solution for this scenario.

swpoole:
Agreed.

Best wishes
Steve...

jdinan:
Document read on 10/20/2016: [main_spec.pdf](https://github.com/openshmem-org/specification/files/542868/main_spec.pdf)

jdinan:

PR #10, opened 2016-08-16
[NO MERGE] -- Fix the examples

jdinan:
This PR includes a Makefile to build and run the C/Fortran examples, as well as a few fixes.

This is turning out to be a much bigger task than I initially thought, so I'm putting this out there to see if there's someone else who can pick it up.  It appears that many of the examples (especially the Fortran examples) were corrupted between OpenSHMEM 1.1 and 1.3, and there are many other lurking bugs and errors.

Any takers?


manjugv:
@dawelch (at ORNL) had worked on this particularly the C/C++ examples, 
and IIRC, C/C++ examples should be fine. If it is not, probably he 
has already had some changes and it is not in the tree.

For Fortan, I’m not sure of the state.

@dawelch Can you please sync up with Jim.

jdinan:
One of the C examples didn't compile -- that's fixed in d066e57 on this PR.  Note that this patch is reverting the example to what it was in OpenSHMEM 1.1.

dawelch:
That test was changed to reflect the update adding c11 generics as the new (preferred?) way of performing operations in commit e2766e628ba8a1475e884549b7388ce7584fc9c7.  To compile and run that test, you need both a 1.3 compliant implementation as well as a compiler that supports c11 generics.  With those two things, the shmem_add in that example is simply a macro that translates to shmem_int_add.  Without c11 generics, that won't work, but that was the point of the example - I'm not sure if we should be specifically downgrading the examples in the newer specification version to support the syntax and semantics of older versions.

jdinan:
In that case, the bug is that it should have been ```shmem_fadd``` instead of ```shmem_add```.

dawelch:
You're right, I missed that part in the diff, and shoot, I do in fact have a few changes that somehow never made it out.  For c/c++ it looks like maybe 7 examples including the add example, the changes for which seem to be mostly concerning issues with the c11 api update.  I'll have to retest it all against the latest version.

I don't appear to have had any outstanding changes to any fortran examples, so either they worked for me at the time or, more likely, they were not included in my prior testing.  I'll also check those out while I'm at it.

swpoole:
Where are his edits?

Best wishes
Steve...

jdinan:
The Fortran tests don't compile (see patches on this PR for two examples), so it's unlikely that you tested them.  Some of the Fortran issues are bugs introduced since 1.1 (or since the SGI manpages where they originated) and some are old bugs.

I'd like to integrate testing the examples with Travis to avoid these sort of issues in the future.  There's a simple Makefile in 6b79a9b that ought to be usable for this purpose.  The model would be that all examples are compatible with running on 2 PEs and they have nonzero exit code if there is a failure.

swpoole:
Most of the Fortran programs are incomplete and few of them show good programming practices.
It is not "required" to have a "program" statement at the beginning, but is good practice. Most of the
examples do not have shmem_finalize() at the end. Most do not have "end" statements.

jeffhammond:
I'll rewrite all the Fortran examples to not suck if encouraged to do so.

swpoole:
I am working on them. I will send them to you when I finish. They are
totally hosed over.
I am running SOS, your version of OpenSHMEM and the reference version to
make sure
it is just not the code (fortran) that bites. Many of them seem just to
be suggestions of how
to use a single call in fortran. Not too useful as examples.

Steve...

On 8/28/16 14:50, Jeff Hammond wrote:
> I'll rewrite all the Fortran examples to not suck if encouraged to do so. 
>
>

-- 
Regards,
Steve...

"Don't mistake my kindness for weakness. I am kind to everyone,
but when someone is unkind to me, weak is not what you are going
to remember about me." -- Al Capone.

jdinan:
Any update, @swpoole ?

jdinan:
This was split into #22 (makefile), #23 (C examples), and jdinan:proposal/fortran-examples.

PR #12, opened 2016-08-18
Add .gitignore

nspark:
Add .gitignore to ignore generated files in the tree


jdinan:
:+1:

shamisp:
@dkhaldi for the record, I would recommend to +1 and only then merged. Otherwise it is not clear if the path was +2.

dkhaldi:
+1
@shamisp Sure

PR #13, opened 2016-08-18
Expand RMA and AMO type support

nspark:
This PR expands RMA and AMO type support.

This is Redmine ticket 189. This ticket had a formal reading at the August 2016 meeting, but this PR includes subsequent changes (as well as change highlighting). Merge is pending special ballot and formal ballot.

**Special Ballot**: diff ([6fcdf02..a7b4673](https://github.com/openshmem-org/specification/pull/13/files/6fcdf02414113492ddb1c9c822a31a4986c4092d..a7b4673049be29210f8a9cae10df95293837bacd)) and [full PDF](https://github.com/openshmem-org/specification/files/838032/proposal-extended-generics-20170313.pdf)

**To Do**:

- [x] Pull a cherry-pick of ef06661 into this branch to remove `[u]intptr_t`
- [x] Address `stdint.h` and C99/C11 dependence


content/atomics_intro.tex
```diff
@@ -52,18 +49,26 @@
...
+      unsigned int       & uint       \\ \hline
+      unsigned long      & ulong      \\ \hline
+      unsigned long long & ulonglong  \\ \hline
+      int32\_t           & int32      \\ \hline
+      int64\_t           & int64      \\ \hline
+      uint32\_t          & uint32     \\ \hline
+      uint64\_t          & uint64     \\ \hline
+      size\_t            & size       \\ \hline
+      ptrdiff\_t         & ptrdiff    \\ \hline
+      wchar\_t           & wchar      \\ \hline
```
jdinan: Adding ```wchar_t``` to the AMO types makes me nervous because it can be smaller than ```int```, thereby introducing a new requirement to the AMO interface.  There should not be performance impact to the other AMO routines, because we isolate now by type.  Please correct me if I'm wrong, but adding the ```wchar_t``` type seems to be of little value to users and an unnecessary distraction to implementors.

content/rma_intro.tex
```diff
@@ -16,33 +16,59 @@
...
+      uintptr\_t         & uintptr    \\ \hline
+%}
     \end{tabular}
     \caption{Standard \ac{RMA} Types and Names}
     \label{stdrmatypes}
   \end{center} 
 \end{table}
+
+\footnotetext{\newtext{In order to maintain compatibility between C and C++,
+the C interface requires the \VAR{bool} macro as defined in stdbool.h.}}
```
jdinan: I suggest that we introduce a C99 baseline requirement for the C bindings and change this footnote to indicate that implementations may choose to use the type ```_Bool``` to *avoid* a dependency on ```stdbool.h```.  I have no idea how to support a ```bool``` type (nor do I know how to support a ```complex``` type) portably prior to C99.  We can further indicate that implementations may optionally choose to support earlier versions of the C language by disabling bindings for types not defined prior to C99.

If we want ```bool``` and ```complex``` to work in C++, then we should really define proper C++ bindings.  As defined, C++ users are calling C functions and would need to supply arguments compatible with the C types.

content/atomics_intro.tex
nspark: Correct; it was mostly there for brevity (to simply reference the integer types of stddef.h). Will update.

content/rma_intro.tex
nspark: @jdinan I agree with the overall strategy, though I think it should be a separate ticket/issue. For now, I've put in a placeholder in Redmine and will work on proposed text. Do you consider such a ticket as a happens-before change for this ticket?

content/atomics_intro.tex
```diff
@@ -29,19 +29,16 @@
 \ac{AMO} routines each support the ``standard \ac{AMO} types’’ listed in Table \ref{stdamotypes}, 
 except for \FUNC{shmem\_fetch}, \FUNC{shmem\_set}, and \FUNC{shmem\_swap}, which supports the ``extended \ac{AMO} types’’ listed 
 in Table \ref{extamotypes}.
+\newtext{Implementations may optionally support additional types.}
```
jdinan: This is in conflict with Sec. 5, paragraph 1: ```The OpenSHMEM header files for C and Fortran must contain only the interfaces and constant names defined in this specification.```

But, also, I don't think this statement adds anything of substance to the specification.

content/backmatter.tex
```diff
@@ -357,6 +357,9 @@ \section{Version 1.5}
 
 \item The \VAR{SHMEM\_SYNC\_SIZE} constant was added.
 \\See Section \ref{subsec:library_constants}.
+%
+\item \newtext{Expanded the type support for RMA and AMO operations.
+\\ See Sections \ref{sec:rma} and \ref{sec:amo}.}
```
jdinan: I think it would be a good idea to list all of the types that were added in the changelog entry.  In the future, if a user is trying to figure out whether their OpenSHMEM library should support a given type, this will be useful information.

content/rma_intro.tex
```diff
@@ -16,31 +16,65 @@
 one-sided communication interfaces via \Celev{} generic selection 
 (\Celev{} \S6.5.1.1\footnote{Formally, the \Celev{} specification is ISO/IEC 9899:2011(E).})
 for block, scalar, and block-strided put and get communication. 
-Such type-generic routines are supported for the ``standard \ac{RMA} types’’
-identified in Table \ref{stdrmatypes}.
+Such type-generic routines are supported for the
+\newtext{\Celev{}
+``standard integer types'' (\Celev{} \S6.2.5.7),
+``real floating types'' (\Celev{} \S6.2.5.10),
```
jdinan: real floating +point+ types?

content/rma_intro.tex
```diff
@@ -16,31 +16,65 @@
...
+``exact-width integer types'' (\Celev{} \S7.20.1.1),
+``integer types capable of holding object pointers'' (\Celev{} \S7.20.1.4),
+and the integer types \textit{size\_t} and \textit{ptrdiff\_t} defined in
+stddef.h (\Celev{} \S7.19), which are collectively called the}
+``standard \ac{RMA} types''
+\newtext{and are listed} \oldtext{identified} in Table \ref{stdrmatypes}.
+\newtext{Implementations may optionally support additional types.}
```
jdinan: Same comment as above; I don't think line 29 adds anything of substance and it is potentially in conflict with restrictions from other parts of the spec.

content/rma_intro.tex
```diff
@@ -16,31 +16,65 @@
...
+
+\textbf{\newtext{Note to implementors:}}
+\begin{itemize}
+\item \newtext{%
+  In order to provide compatibility between C and C++, implementors may elect
+  to include a dependence on stdbool.h, which provides the macro \textit{bool}.
```
jdinan: Do you get better compatibility with C++ by including the C header?  This link seems to suggest not, http://www.cplusplus.com/reference/cstdbool/.

content/rma_intro.tex
nspark: No. The [C specification](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf) refers to "real floating types":

> There are three _real floating types_, designated as **`float`**, **`double`**, and **`long
double`**. [...]

content/rma_intro.tex
```diff
@@ -16,31 +16,65 @@
...
+  Alternatively, implementors may elect to avoid this dependence and use the
+  C99\footnote{Formally, the C99 specification is ISO/IEC 9899:1999(E), including amendments and corrections.}
+  \textit{\_Bool} type explicitly.
+  }
```
jdinan: How are these types are handled in older versions of the C language?  Can we also expand this to cover the same issue with ```complex```?

content/rma_intro.tex
nspark: With `stdbool.h`, the C preprocessor expands `bool` to `_Bool`. With or without `cstdbool`, the C++ preprocessor has no effect on `bool` types. The compatibility is principally lexical; including `stdbool.h` and using `bool` as the type (instead of `_Bool`) will make the code parse-able by C and C++ compilers. Object-representation compatibility for the C `_Bool` and C++ `bool` types is not defined across the standards, as far as I can tell.

jdinan: Right, but I can also have a C++ case in my header file that uses the ```bool``` type instead of ```_Bool```.  What is a bit concerning, is that if the C/C++ objects are not compatible then I need to have a proper C++ binding to support this.

content/rma_intro.tex
jdinan: Fair enough, but we have exactly the same issue with ```complex``` and it seems likely that this isn't the last time we will run into incompatibility across language versions.  It would be great if this text could be moved to the language bindings section, and accompanied by a table indicating the types with limitations on where they are supported.  For starters, we should indicate that ```bool``` and ```complex``` are only supported in C99 or later.

content/rma_intro.tex
nspark: That's fair, but, according to Section 5 "Language Bindings and Conformance", there is no C++ binding (and our use of "C/C++" is ambiguous and sloppy). Further, one could argue that the reference to "ISO C" is most technically referring to C11 (as of this writing), because C11 formally supersedes C99.

Alternatively, we work with the fact that there is no "C" -- there are only C89/C90, C99, and C11 -- and declare the points of incompatibility between OpenSHMEM and all three C standards. This, I think is much bigger than this ticket.

jdinan: Aren't the fixed-width integer types also C99?

content/rma_intro.tex
jdinan: I think what I'm trying to ask for here is that we define what happens when you pass ```shmem.h``` through a C89 compiler.  We now have multiple types used in the spec, but not defined prior to C99.  So, what should happen?  Am I allowed to disable these bindings in C89 mode so that this will work?  Or, should we require C99 as a baseline?

content/rma_intro.tex
```diff
@@ -16,31 +16,40 @@
...
+      int8\_t            & int8       \\ \hline
+      int16\_t           & int16      \\ \hline
+      int32\_t           & int32      \\ \hline
+      int64\_t           & int64      \\ \hline
+      uint8\_t           & uint8      \\ \hline
+      uint16\_t          & uint16     \\ \hline
+      uint32\_t          & uint32     \\ \hline
+      uint64\_t          & uint64     \\ \hline
+      intptr\_t          & intptr     \\ \hline
+      uintptr\_t         & uintptr    \\ \hline
```
jdinan: The ```inptr_t``` and ```uintptr_t``` types kind of beg the question -- why would one be sending pointers across the network in the first place (and for that matter pointers converted to ints)?  Especially since we caution users against sharing pointers elsewhere in the spec.  I don't have much of an argument against, I just can't imagine any legitimate usage for these types through the SHMEM API.  Maybe someone can help me out?

BryantLam:
I'm not sure we need all (any?) of the `stdint.h` or `stddef.h` types to provide RMA support for them. A lot of those `int_t` flavored types are `typedef`'d to one of the other primitive types (Is there a platform that this isn't true?).

In my skeleton implementation of this expanded-types support, I get the following errors using `clang`:
```
error: type 'int8_t *' (aka 'signed char *') in generic association compatible with previously specified type 'signed char *'
error: type 'int16_t *' (aka 'short *') in generic association compatible with previously specified type 'short *'
...
```
Similar errors appear with GCC:
```
error: '_Generic> selector matches multiple associations
error: '_Generic' specifies two compatible types
...
```

I don't have a clever way to fix this without some tooling support (e.g., autotools) and a ghastly configure-time check.

Alternatively, an implementation could just simply not implement the `stdint.h` types for RMA and be "compliant" by way of `typedef`, yet the standard could still specify support as requierd.

jdinan:
Yikes, I reproduced this issue in SOS..  This is indeed a restriction on C11 generic selection -- no two type names in the list can specify compatible types.

nspark:
I noted this on [Redmine](http://www.openshmem.org/redmine/issues/189#note-15) a while ago; @jdinan referred to this as "horrifying". I think the practical implication is that an implementation will provide generic associations for the standard integer types (e.g., `int`, `unsigned long`) and compatible fixed-width types (e.g., `int32_t`, `uint64_t`) will automatically be supported. I haven't seen a platform where the fixed-width integer types _weren't_ compatible with a standard integer type. See [this gist](https://gist.github.com/nspark/39355c64eeddd49e7c57dfe547939912) to try it out for your platform.

Here's what I see (Linux x86-64, GCC 5.4.0):
```
         int8_t is typedef'd to signed char
        int16_t is typedef'd to short
        int32_t is typedef'd to int
        int64_t is typedef'd to long
        uint8_t is typedef'd to unsigned char
       uint16_t is typedef'd to unsigned short
       uint32_t is typedef'd to unsigned int
       uint64_t is typedef'd to unsigned long
        wchar_t is typedef'd to int
         size_t is typedef'd to unsigned long
      ptrdiff_t is typedef'd to long
       intptr_t is typedef'd to long
      uintptr_t is typedef'd to unsigned long
       intmax_t is typedef'd to long
      uintmax_t is typedef'd to unsigned long
    max_align_t is typedef'd to <not a standard integer type>
```

The restriction against compatible types in generic selection means that, if we _only_ adopted the generic API, an implementer would not need to provide both `shmem_int_put()` and `shmem_int32_put()`.

content/rma_intro.tex
nspark: @jdinan I understand the aversion to RMAing pointer values between PEs, but as noted below about compatible types, the capability will certainly exist. That said, I'm not opposed to removing explicit listing of support for `intptr_t` and `uintptr_t`.

nspark:
> I'm not sure we need all (any?) of the `stdint.h` or `stddef.h` types to provide RMA support for them. [...]

@BryantLam The point of the table is to identify the types that are supported. If two types are compatible types and supported by the same generic association, I think it _potentially_ simplifies the number of interfaces needing implementation. Still, having names such as `shmem_int32_put()` and `shmem_uint64_put()` is helpful for users.

jdinan:
Expanding on Nick's gist, it should be workable to use a program like [this](https://gist.github.com/jdinan/dd1c6c2389a2c01761ff6eaae35ba876) to have the build system capture the types that are not compatible with the C basic types supported by OpenSHMEM.  This gist generates M4 that I would use autoconf to substitute into my [C bindings generator](https://github.com/Sandia-OpenSHMEM/SOS/blob/master/src/shmem_bind_c.m4).  But, you could also use this technique to create a program that emits the full complement of C11 generic bindings.

nspark:
I basically think this PR is where I want it to be, so please review and comment. I'll be bringing this for reading at the February specification face-to-face.

The one change from about a month ago is that I added floating-point support for atomic add and fetch-and-add. Given our atomicity model, it does not affect the performance of integer AMOs for platforms without hardware FP AMOs, and some hardware does have FP AMOs.

jdinan:
@nspark Could you resolve the merge conflict before submitting the proposal for a reading?

content/rma_intro.tex
jdinan: I'm not going to vote against the proposal if you want to keep these, but this feels like we are adding functions that will never get used.

nspark: I think I just forgot to take these out. I'll remove `intptr_t` and `uintptr_t`.

nspark:
@jdinan Yes, I can, but it will come up again if another PR with a tail-appended change-log entry gets merged before this one. I don't know of an easy way to resolve for the general case.

content/rma_intro.tex
jdinan: Were these removed?  They still show up in the diff.

nspark: I thought I took them out, but I'm realizing that I pulled this PR into #31 and made the changes there, which has the "final" text for both. That said, I still managed to forget to remove `intptr_t` and `uintpt_t` from the RMA type-support table.

BryantLam:
Notes from F2F
- Language conformance for C99/C11 and pre-C99 support for `<stdint.h>`

nspark:
@BryantLam @jdinan What do you think of the following text for addressing the `stdint.h` dependence?

> The standard RMA types include the exact-width integer types defined in _stdint.h_ by
_C99_ §7.18.1.1 and _C11_ §7.20.1.1. When the _C_ translation environment does not provide _stdint.h_, the OpenSHMEM implemementation is not required to provide support for such exact-width integer types.

nspark:
Here's an updated version (with highlighted changes) to address that the exact-width types are optional:

> The standard RMA types include the exact-width integer types defined in _stdint.h_ by _C99_ §7.18.1.1 and _C11_ §7.20.1.1. When the _C_ translation environment does not provide **exact-width integer types with** _stdint.h_, the OpenSHMEM implemementation is not required to provide support for such ~~exact-width integer~~ types.

jdinan:
:+1:

nspark:
Here's the diff from the official reading ([6fcdf02..a7b4673](https://github.com/openshmem-org/specification/pull/13/files/6fcdf02414113492ddb1c9c822a31a4986c4092d..a7b4673049be29210f8a9cae10df95293837bacd)), excluding upstream merge to resolve conflicts in `backmatter.tex`.

nspark:
[PDF of this PR for Special Ballot (3/22)](https://github.com/openshmem-org/specification/files/838032/proposal-extended-generics-20170313.pdf)

shamisp:
Conflicts with #31 that removes floats from the list of extended AMOs.

nspark:
@shamisp Argh. Stupid inter-dependent PRs. Since I read these together, I'll do the same with the special ballot. PR #31 has / will have the final PDF.

jdinan:
@nspark Have these changes already been incorporated into #31?

jdinan:
Oh, neat!  Github was smart enough to see that #13 was merged into #31 and it automatically marked this PR as merged when I merged #31.  :octocat: :clap:

PR #14, opened 2016-08-18
Rename SMA-prefixed env-vars to use SHMEM prefix

nspark:
This PR renames SMA-prefixed environment variables to use the SHMEM prefix.

This is Redmine ticket 222. This ticket had a formal reading at the August 2016 meeting, but this PR includes subsequent changes to note the deprecated env-vars and update the deprecation and change-log annexes. Merge is pending special ballot and formal ballot.

### DIFF
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -340,7 +340,9 @@ specification.
     \FUNC{SHMEM\_PUT} & 1.2 & Yes & \FUNC{SHMEM\_PUT8} or \FUNC{SHMEM\_PUT64} \\ \hline
     \FUNC{SHMEM\_CACHE} & 1.3 & Yes & (none) \\ \hline
     \_SHMEM\_* constants & 1.3 & Yes & SHMEM\_* \\ \hline
-    \hline
+%\newtext{
+    SMA\_* environment variables & 1.4 & Yes & SHMEM\_* \\ \hline
+%}
     \end{tabular}
 \end{center}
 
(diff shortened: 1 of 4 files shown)


content/backmatter.tex
```diff
@@ -340,7 +340,9 @@ \chapter{\openshmem Specification and Deprecated API}\label{sec:dep_api}
     \FUNC{SHMEM\_PUT} & 1.2 & Yes & \FUNC{SHMEM\_PUT8} or \FUNC{SHMEM\_PUT64} \\ \hline
     \FUNC{SHMEM\_CACHE} & 1.3 & Yes & (none) \\ \hline
     \_SHMEM\_* constants & 1.3 & Yes & SHMEM\_* \\ \hline
-    \hline
+%\newtext{
+    SMA\_* environment variables & 1.4 & Yes & SHMEM\_* \\ \hline
```
jdinan: All deprecated environment variables should be listed individually, see #25.

nspark: That makes #25 effectively a gate for even reading these changes. Since the relevant changes to the deprecation annex would be non-semantic changes, they _should_ be addressable post-ballot by the section committee(s) responsible for the deprecation annex.

dkhaldi: @jdinan @nspark Since this ticket will get voted on first (before #25), I think #25 should adapt its changes to this one, after merging, and not the opposite.

jdinan: We should not delay #25 either.  The responsibility for this fix should fall on this proposal, which is proposing the deprecations.

nspark:

dkhaldi:
@nspark, I guess you chose to not break down the SMA_ variables, as James suggested, Annex F. I will just wait for your confirmation and then merge this PR as it is.

nspark:
@dkhaldi IIRC, the (my) plan is to issue an additional PR based on #25 to amend the deprecation listing and the change log grammar. I think we agreed to merge this one as-is.

dkhaldi:
Très bien.

PR #15, opened 2016-08-18
Add _Noreturn specifier to shmem_global_exit

nspark:
This PR adds the `_Noreturn` function specifier (C11) and `noreturn` attribute (C++11) to `shmem_global_exit`.

This is Redmine ticket 223. This ticket had a formal reading at the August 2016 meeting, but this PR includes subsequent changes to add C++11 support (as well as change highlighting). Merge is pending special ballot and formal ballot.

### DIFF
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -358,8 +358,8 @@ specification.
 \item The \VAR{SHMEM\_SYNC\_SIZE} constant was added.
 \\See Section \ref{subsec:library_constants}.
 %
-\item \newtext{The C11 \textbf{\_Noreturn} function specifier and the C++11
-\textbf{noreturn} attribute were added to \FUNC{shmem\_global\_exit}.
+\item \newtext{The C11 \textbf{\_Noreturn} function specifier was added to
+  \FUNC{shmem\_global\_exit}.
 \\ See Section \ref{subsec:shmem_global_exit}.}
 
 \end{itemize}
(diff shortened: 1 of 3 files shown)


jdinan:
At present, we don't have C++ bindings and we tell C++ users that they will have to get by with the C API.  Adding a C++ binding for this one function seems like a half measure.  I'm for supporting C++, but I don't like the idea of doing it through special cases (in this case a specific function and also a specific version of C++).

nspark:
That seems reasonable, since it really is _just_ for this one call.

nspark:

dkhaldi:
@nspark, we voted yes on this ticket. But there was thread-safety issues raised by Naveen in this email thread (Noreturn shmem_global_exit() and behavior changes in multithreaded Cray SHMEM).
Should I go ahead and merge this PR since it has been ratified or would you rather prefer clarify things more?

jdinan:
@dkhaldi, I think we should go ahead and merge what was ratified.  If any change is needed (e.g. backing this out), it will have to go through the full proposal process.

On the technical side, this proposal is consistent with the spec as it currently exists; if thread safety introduces a new challenge, it will need to be addressed as part of that effort.  For example, we could release another version of the spec before thread safety is integrated; in which case holding back this ratified proposal would be incorrect.

nspark:
@dkhaldi I agree with @jdinan. I thought about holding this one for a bit, but I think it would set a bad precedence because we did all vote to ratify it, so it should go in and we can deal with thread-related issues with future changes.

dkhaldi:
Sounds good.

shamisp:
@jdinan agree. 

Seems like the easiest solution would be to remove the qualifier in the future. My only hope that it will not cause any portability issues.

PR #16, opened 2016-08-19
Generic shmem_wait interfaces

nspark:
This PR adds generic interfaces for `shmem_wait` and `shmem_wait_until`.

This is Redmine ticket 216. This ticket had a formal reading at the August 2016 meeting, but this PR has a mostly-cosmetic cleanup in `shmem_wait` to condense redundant tables for C/C++ and Fortran. Merge is pending special ballot and formal ballot.

This ticket has a dependence on PR #6.

### DIFF
--- a/content/shmem_wait.tex
+++ b/content/shmem_wait.tex
@@ -5,15 +5,15 @@
 \begin{apidefinition}
 
 \begin{C11synopsis}
-void shmem_wait(volatile TYPE *ivar, TYPE cmp_value);
-void shmem_wait_until(volatile TYPE *ivar, int cmp, TYPE cmp_value);
+void shmem_wait(TYPE *ivar, TYPE cmp_value);
+void shmem_wait_until(TYPE *ivar, int cmp, TYPE cmp_value);
 \end{C11synopsis}
 where \TYPE{} is one of the point-to-point synchronization types specified by
 Table \ref{p2psynctypes}.
 
 \begin{Csynopsis}
-void shmem_<TYPENAME>_wait(volatile TYPE *ivar, TYPE cmp_value);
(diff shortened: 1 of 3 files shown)


content/shmem_wait.tex
```diff
@@ -4,18 +4,19 @@
 
 \begin{apidefinition}
 
+\begin{C11synopsis}
+void shmem_wait(volatile TYPE *ivar, TYPE cmp_value);
+void shmem_wait_until(volatile TYPE *ivar, int cmp, TYPE cmp_value);
```
nspark: +1

manjugv:
@nspark, I think squashing the commits to one commit would be good.

nspark:
@manjugv That seems reasonable. I think you can tweak the repo to [only allow squash commits](https://github.com/blog/2141-squash-your-commits).

jdinan:
IIUC, this will discard history, so we should be careful unless we plan to never make mistakes.  Certainly, I wouldn't want the owner of the PR to do this manually (i.e. rebase and push --force) because we will lose the online discussion on the PR (after the commit hashes change, github doesn't know where to hang comments).  In this particular case, I don't really see much junk that I'd want to purge from the history.

BryantLam:
I'm of the opinion that you should not squash commits unless the changes you are squashing are trivial. For the specification, a preservation of history through the commits is desirable.

manjugv:
Squashing commits doesn’t destroy history rather it can provide concise and clear history. It provides a clean way to identify commit related to the ticket and PR.

jeffhammond:
Squashing should only happen on forks and should not occur after PR. People should get their commits right, with or without squash, and then submit a PR. If squash required after PR, close original PR and resubmit after squash.

PR #17, opened 2016-09-06
define complex reductions in a better way

jeffhammond:
1) `complex` needs `#include complex.h`, where `_Complex` does not.  On the other hand, `_Complex` is only defined in C99 (this is not to suggest that `complex` is defined in C++, because it is not.
2) `std::complex<float>` and `float _Complex` (and likewise for `double`) are equivalent from a library perspective (http://stackoverflow.com/a/10540346/2189128), so those can be used for C++.
3) I added a C99 and C++ latex environment thing.

@jdinan I'm waiting for your feedback :trollface:

### DIFF
--- a/content/shmem_reductions.tex
+++ b/content/shmem_reductions.tex
@@ -64,8 +64,6 @@ CALL SHMEM_REAL16_MIN_TO_ALL(dest, source, nreduce, PE_start, logPE_stride, PE_s
 \textbf{SUM} \newline
 Performs a sum reduction across a set of processing elements (\ac{PE}s).\newline
 \begin{Csynopsis}
-void shmem_complexd_sum_to_all(double complex *dest, const double complex *source, int nreduce, int PE_start, int logPE_stride, int PE_size, double complex *pWrk, long |\mbox{*pSync);}|
-void shmem_complexf_sum_to_all(float complex *dest, const float complex *source, int nreduce, int PE_start, int logPE_stride, int PE_size, float complex *pWrk, long *pSync);
 void shmem_double_sum_to_all(double *dest, const double *source, int nreduce, int PE_start, int logPE_stride, int PE_size, double *pWrk, long *pSync);
 void shmem_float_sum_to_all(float *dest, const float *source, int nreduce, int PE_start, int logPE_stride, int PE_size, float *pWrk, long *pSync);
 void shmem_int_sum_to_all(int *dest, const int *source, int nreduce, int PE_start, int logPE_stride, int PE_size, int *pWrk, long *pSync);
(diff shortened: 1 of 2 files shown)


dkhaldi:
@jeffhammond , your PR should be generated against the new development branch osh_spec_next, not the master branch. Could you please change this?

jdinan:
We also ask for a redmine ticket to track the proposal.

The solution would address this implementor's complaining, but at the expense of shifting pain to the user.  The C binding change is not backward compatible and it also has a confusing failure mode.  If users compile the same code with C89 and C99 compilers, the former case breaks and the latter does not.

jeffhammond:
Sure, once we start following the universal best practice of developing on master ;-)

jeffhammond:
@jdinan `complex` is a macro for `_Complex`, so using the latter does nothing to the bindings other than make `complex.h` unnecessary.  There is no change in the C bindings here.  ISO C ("The macro complex expands to _Complex...") and [POSIX](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/complex.h.html) both agree on this.

Can you show me an example of how replacing `complex` with `_Complex` breaks user code or changes the binary representation of an object file (other than due to symbols from `complex.h`?

As for C89, can you explain how such code would compile in C89 in the first place?  `complex` is not a keyword in C89.

jeffhammond:
I'll create the Redmine ticket after I sort of what is likely to work here.  Github has a vastly superior interface and I don't want to pay synchronization overhead between the two sites for something that is likely to change rapidly with time.

jeffhammond:
I mean, C89 literally gives an example where `complex` is a user-defined type:
```c
         typedef int MILES, KLICKSP();
         typedef struct { double re, im; } complex;
```

From http://port70.net/~nsz/c/c89/c89-draft.html.  This appears to be a pdf2html version.  I am still looking for a PDF.

jdinan:
@jhammond, kindly RTFM https://github.com/openshmem-org/specification/blob/master/README.md :trollface:

With regard to C89, I think you have your answer.  The questions I still have are where the heck was this type coming from prior to C99 and why do we even have this horribly incomplete set of C language bindings for complex types?  Sadly, the SGI manpages seem to have vanished from the web since the acquisition by HP.  I'd be interest to know if there are any breadcrumbs there..

jeffhammond:
@jdinan I am very confused about your intent w.r.t. C89.  Are you suggesting that the current bindings can work there if the vendor uses a compiler extension, whereas moving to the C99 standard breaks codes that were using e.g. `#define complex __complex__` (https://gcc.gnu.org/onlinedocs/gcc/Complex.html)?

As for `osh_spec_next`, if the documentation tells me to punch myself in the face every time I call `shmem_init`, does that make it right?  My point is that OpenSHMEM has documented a deviant and therefore counterintuitive practice and the practice itself should change.

jdinan:
I have no idea how these functions are supposed to work prior to C99.  And yet, I am supposed to support this in the current OpenSHMEM spec.

jeffhammond:
@jdinan I know that this can be solved in Sandia OpenSHMEM, so long as the user doesn't use `CFLAGS=-std=c89 -ansi -Werror` (or something along those lines).  Feel free to assign an issue to me if you want me to contribute a fix.

jdinan:
We have a separate SOS thread on this already: https://github.com/Sandia-OpenSHMEM/SOS/pull/233.

jdinan:
@jeffhammond will this be resolved by #29?

jdinan:
@jeffhammond #29 was merged.  Do you still want to keep this PR open?

jeffhammond:
No
-- 
Jeff Hammond
jeff.science@gmail.com
http://jeffhammond.github.io/

PR #18, opened 2016-09-07
redo PR 17

jeffhammond:
This is the correct version of https://github.com/openshmem-org/specification/pull/17

### DIFF
--- a/content/shmem_reductions.tex
+++ b/content/shmem_reductions.tex
@@ -64,8 +64,6 @@ CALL SHMEM_REAL16_MIN_TO_ALL(dest, source, nreduce, PE_start, logPE_stride, PE_s
 \textbf{SUM} \newline
 Performs a sum reduction across a set of processing elements (\ac{PE}s).\newline
 \begin{Csynopsis}
-void shmem_complexd_sum_to_all(double complex *dest, const double complex *source, int nreduce, int PE_start, int logPE_stride, int PE_size, double complex *pWrk, long |\mbox{*pSync);}|
-void shmem_complexf_sum_to_all(float complex *dest, const float complex *source, int nreduce, int PE_start, int logPE_stride, int PE_size, float complex *pWrk, long *pSync);
 void shmem_short_sum_to_all(short *dest, const short *source, int nreduce, int PE_start, int logPE_stride, int PE_size, short *pWrk, long *pSync);
 void shmem_int_sum_to_all(int *dest, const int *source, int nreduce, int PE_start, int logPE_stride, int PE_size, int *pWrk, long *pSync);
 void shmem_double_sum_to_all(double *dest, const double *source, int nreduce, int PE_start, int logPE_stride, int PE_size, double *pWrk, long *pSync);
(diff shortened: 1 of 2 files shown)


jdinan:
I'd like to see a changelog entry noting that the complex routines are no longer defined for the "generic" C API.  This is technically backward compatibility breakage, although I do not expect any users to be impacted.  There should also be a changelog entry noting that C++ bindings are added for these routines.

jeffhammond:
> I'd like to see a changelog entry
> 
Sure, whenever the changes are finalized. 
> noting that the complex routines are no longer defined for the "generic" C API. This is technically backward compatibility breakage, although I do not expect any users to be impacted.
> 
Why can't they be supported under generic C API? C99 is compatible with C11. 

You want me to change to C99+ instead?
> There should also be a changelog entry noting that C++ bindings are added for these routines.
> 
Sure.

jdinan:
Changelog entries are easy to write and will help a reader understand the intent of your proposal.

The "generic" C API also encompasses earlier versions of C (since we don't specify) that don't have the _Complex type.  If someone is using a C89 compiler, they no longer get the complex functions.

jeffhammond:
By "generic" C API, you apparently do not mean the `_Generic` C API, as I originally thought.

As the generic C API could not have worked with a C89 compiler without depending on the existence of a user-defined type ` complex` whose definition is not specified anywhere in the standard, I don't see any reason to identify breakage.

If you can give me a specific example of something ISO-compliant that would be broken by this, I'll revise my position.

nspark:
By "generic C API" do you mean C11 `_Generic` or the "vanilla" C API? There isn't any support for complex types in the C11-generic API (for put/p/iput/get/g/iget) and I hadn't planned on adding them in the "expanded RMA types" PR #13.

jdinan:
These functions *must have worked* and *may still work* in a non-C99 compiler in some SHMEM implementation.  Your proposal will remove that functionality.  Whether you find is aesthetically appealing or not, we must note the breakage of backward compatibility in the changelog entry.

jdinan:
@nspark s/generic/vanilla/

jeffhammond:
> These functions must have worked and may still work in a non-C99 compiler in some SHMEM implementation.
> 
Worked in what sense? The type in the signature is defined nowhere in the spec.

Breaking undefined features seems unworthy of documentation. 
> Your proposal will remove that functionality. Whether you find is aesthetically appealing or not, we must note the breakage of backward compatibility in the changelog entry.
> 
I'll note that implementation-defined behavior is still implementation-defined.

I refuse to acknowledge "complex" has a meaning in C89 and I refuse to build standards documents on non-standard compiler extensions.

jdinan:
@jeffhammond - just because we didn't define it fully in OpenSHMEM doesn't mean that it wasn't fully defined and functional in an earlier SHMEM implementation.

jeffhammond:
And implementers are permitted to carry that torch. The standard should not.

jdinan:
I agree and I support the change.  However, it breaks backward compatibility (albeit, for a feature that was ill-defined) and this should be noted in the changelog.

jdinan:
@jeffhammond Is this superseded by #29?

jeffhammond:
@jdinan Yeah, @nspark should feel free to pull in anything he likes about this ticket.

PR #19, opened 2016-10-05
Correct active set indexing in a2a

jdinan:
Buffer indexing in all-to-all collectives was incorrect for active sets
that contain a subset of PEs.

Ref: redmine ticket 208 (http://www.openshmem.org/redmine/issues/208).


jdinan:
@spophale I merged in the wording used in redmine ticket 212.  Please review and let me know what you think.

jdinan:
The "th"s should be superscripted already.  Did I miss some?  Here's the current PDF: [main_spec.pdf](https://github.com/openshmem-org/specification/files/556434/main_spec.pdf).

jdinan:

PR #21, opened 2016-10-05
Fix bugs in the C examples

jdinan:


jdinan:
Argh, need to close and re-open to fix something...

PR #22, opened 2016-10-05
Add Makefile for examples

jdinan:
This PR only adds the makefile for automating building and running the examples.  It's not perfect, but it should be a reasonable start for us to build on.

Note: the Fortran tests are not even compiling right now, so they are omitted from the ```run``` target.

Needs two reviews to merge: @dkhaldi @shamisp or others, please help

### DIFF
--- a/example_code/Makefile
+++ b/example_code/Makefile
@@ -1,0 +1,30 @@
+CC       = oshcc
+CFLAGS   = -Wall -O3
+
+FC       = oshfort
+FFLAGS   = -Wall -O3
+
+RUNCMD   = oshrun
+RUNOPT   = -np 2
+
+C_TESTS  = $(wildcard *.c)
+C_BINS   = $(C_TESTS:.c=.cx)
+
+F_TESTS  = $(wildcard *.f90)
+F_BINS   = $(F_TESTS:.f90=.fx)
+
+.PHONY: all run clean
+
+all: $(C_BINS) $(F_BINS)
+
+%.cx: %.c
+	$(CC) $(CFLAGS) -o $@ $+
+
+%.fx: %.f90
+	$(FC) $(FFLAGS) -o $@ $+
+
+run: $(C_BINS) $(F_BINS)
+	for bin in $+; do $(RUNCMD) $(RUNOPT) ./$$bin; done
+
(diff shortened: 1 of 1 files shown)


jeffhammond:
I recommend that you remove "-Wall" since that is not a universal flag.  It is not valid for Intel Fortran (`-warn all`) and IBM XL has not direct equivalent (`-qinfo=all` is close).

jeffhammond:
I will 👍  this once you add my previous comment 😉

dkhaldi:
+1

shamisp:
@jeffhammond I don't think we have to be very concerned with C flags portability issues. 

Having ```-Wall``` is useful for the continues integration build . I think we should keep it. I'm not sure why we need ```-O3```, but I suggest to add ```-Werror```.

jdinan:
I could change this to:
```
CFLAGS ?= -Wall -Werror
FFLAGS ?= -Wall -Werror
```
This allows a usage like ```$ make FFLAGS="-warn all"```.  The eventual goal is to build these tests in CI, so I tend to agree with Pasha that I'd prefer for the defaults to support that model.

jeffhammond:
@shamisp Having the default work as often as possible is nice.  You never know when someone is going to download the examples and expect them to work out of the box.  CI systems are by definition automation and can trivially add whatever diagnostic flags are desired.

shamisp:
@jeffhammond it would be nice to have support for every possible platform, but this means that we would have to put in place some build system for example codes (and not just make.) Do we really want to deal here with autotools/cmake/etc ...

jdinan:
@shamisp I think Jeff's suggestion is that we can trivially support any platform by just removing the offending compiler options and having users or the CI script specify them on the command line.

shamisp:
The problem is that by default contributors will not do much. This will
lead to poor patches and  going back and forth to fix those.  Since
"examples" is a not real SW projects, IMHO it is fine to make some
assumptions

jdinan:
Jeff says and he's ok with the approach proposed above.  I pushed a new commit with this change.  @shamisp and @dkhaldi, are we ready to merge?

shamisp:
👍

PR #23, opened 2016-10-05
Fix bugs in the C examples

jdinan:
Bugs fixed:
- Add missing calls to `shmem_finalize()`
- Ensure tests exit with code 0 on success
- shmem_add_example.c: replace `shmem_add` with `shmem_fadd` (transcription error in preview version of the spec)

### DIFF
--- a/example_code/shmem_add_example.c
+++ b/example_code/shmem_add_example.c
@@ -18,5 +18,6 @@ int main(void)
    }
    shmem_barrier_all();
    printf("%d: old = %d, dst = %d\n", me, old, dst);
+   shmem_finalize();
    return 0;
 }
--- a/example_code/shmem_add_example.c
+++ b/example_code/shmem_add_example.c
@@ -14,7 +14,7 @@ int main(void)
    shmem_barrier_all();
 
    if (me == 1){
-      old = shmem_int_fadd(&dst, 44, 0);
+      old = shmem_fadd(&dst, 44, 0);
    }
    shmem_barrier_all();
    printf("%d: old = %d, dst = %d\n", me, old, dst);
--- a/example_code/shmem_add_example.c
+++ b/example_code/shmem_add_example.c
@@ -14,7 +14,7 @@ int main(void)
    shmem_barrier_all();
 
    if (me == 1){
-      old = shmem_add(&dst, 44, 0);
+      old = shmem_int_fadd(&dst, 44, 0);
    }
    shmem_barrier_all();
    printf("%d: old = %d, dst = %d\n", me, old, dst);
--- a/example_code/shmem_add_example.c
+++ b/example_code/shmem_add_example.c
@@ -3,21 +3,20 @@
 
 int main(void) 
 {
-   int me, old;
+   int me;
    static int dst;
 
    shmem_init();
    me = shmem_my_pe();
 
-   old = -1;
    dst = 22;
    shmem_barrier_all();
 
    if (me == 1){
-      old = shmem_fadd(&dst, 44, 0);
+      shmem_add(&dst, 44, 0);
    }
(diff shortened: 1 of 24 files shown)


dkhaldi:
@jdinan, In shmem_add_example.c:  shmem_add should be replaced with shmem_int_fadd  and not shmem_fadd (C11) since this is a C/C++ example.

jeffhammond:
@dkhaldi I think you meant that it should be replaced with `shmem_int_add`.  No need to fetch in that example.

dkhaldi:
@jeffhammond , yes actually I was looking at shmem_fadd_example.c. In both cases (shmem_add_example.c and shmem_fadd_example.c), we need to add _Type_ to the calls.

dkhaldi:
@jdinan, also, since you already improved shmem_global_exit_example.c by adding shmem_finalize, Is it also possible to remove the declaration and definition of variable ‘npes’ because it is set but not used further in the code?

dkhaldi:
My comment about adding TYPE to calls in C programs also applies to other examples such as shmem_finc_example.c (shmem_finc --> shmem_int_finc), shmem_iput_example.c (shmem_iput --> shmem_short_iput), etc
We either correct these or remove this phrase that precedes most of these examples:
"Consider the following shmem_iput example for C/C++ programs." 
Do we want the spec examples to be written in C/C++ or C11?

jdinan:
The goal of this proposal is just to make the examples correct SHMEM programs.  I do agree with all of your suggestions, but I'd like to suggest that they be separate tickets so we can address these obvious errors quickly.

For example, the ```shmem_add_example.c``` has used fetch-and-add since v1.1 even though it is supposed to be an example of add.  Both the add and fadd examples were converted in v1.3 to use the C11 generic API.  This was a conscious decision and a change requires some discussion, whereas I think we can all agree on removing the obvious errors from these programs.

Removing the unused code from shmem_global_exit.c fits well into this change.  I pushed an additional patch to remove the unused ```npes``` variable.

dkhaldi:
OK @jdinan, I will create a separate ticket to remove/modify this sentence "The following shmem_X example is for C/C++ programs:" where appropriate.

example_code/shmem_add_example.c
```diff
@@ -14,9 +14,10 @@ int main(void)
    shmem_barrier_all();
 
    if (me == 1){
-      old = shmem_add(&dst, 44, 0);
+      old = shmem_fadd(&dst, 44, 0);
```
shamisp: is this fadd or add example ? if this is fadd example the file has to be renamed

example_code/shmem_quiet_example.c
```diff
@@ -27,6 +27,7 @@ int main(void)
     shmem_int_put(&targ, &src, 1, 2);  /*put4*/
   }
   shmem_barrier_all();  /* sync sender and receiver */
```
shamisp: You missed the barrier. Please move it.

jdinan: Good catch -- fixed.

example_code/shmem_iput_example.c
```diff
@@ -17,6 +17,6 @@ int main(void)
       (int)dest[0], (int)dest[1], (int)dest[2],
       (int)dest[3], (int)dest[4] );
    }
-   shmem_barrier_all();   /* sync before exiting */
+   shmem_finalize();   /* sync before exiting */
```
dkhaldi: Very minor comment: it would good if you remove the comment here because it is not necessary and does not look consistent with other examples.

jdinan: Ok -- fixed.

jdinan:

PR #27, opened 2016-11-28
Improve shmemx.h portability

nspark:
This PR adds `SHMEM_PROVIDES_SHMEMX_H` constant and identifies `shmemx.h` as an optional header.

See Redmine [Issue 219](http://www.openshmem.org/redmine/issues/219).

### DIFF
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -373,6 +373,10 @@ Thus, the \VAR{volatile} qualifier was removed from the \VAR{ivar} arguments to
 \item The C11 \textbf{\_Noreturn} function specifier was added to
       \FUNC{shmem\_global\_exit}.
 \\ See Section \ref{subsec:shmem_global_exit}.
+%
+\item Added the \VAR{SHMEM\_PROVIDES\_SHMEMX\_H} constant and clarification
+      that the \FUNC{shmemx.h} header is optional.
+\\ See Sections \ref{subsec:bindings} and \ref{subsec:library_constants}.
 
 \end{itemize}
 
--- a/content/language_bindings_and_conformance.tex
+++ b/content/language_bindings_and_conformance.tex
@@ -17,3 +17,5 @@ variables, or identifiers with the prefix \shmemprefix (for \Clang{} and
 All \openshmem extension \ac{API}s that are not part of this specification must
 be defined in the \FUNC{shmemx.h} include file. These extensions shall use the
 \FUNC{shmemx\_} prefix for all routine, variable, and constant names.
+The existence of the \FUNC{shmemx.h} header is optional, but indicated through
+a compile-time constant (Section~\ref{subsec:library_constants}).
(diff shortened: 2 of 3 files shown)


shamisp:
I don't think we need one. You can use autotools/cmake (a.k.a project management of your choice) to detect if the file is there.

jdinan:
Adding to @shamisp, the header is the first of two problems.  Users also need to know if a particular extension exists within shmemx.h.  We could add something like "Implementors are encouraged to provide macros of the form SHMEMX_PROVIDES_[EXTENSION NAME] for each extension."

Also, a minor nit, but I think I prefer "HAVE" to "PROVIDES".

shamisp:
@jdinan - again, user can use autotools to detect availability specific shmemx_* functionality. https://www.gnu.org/software/autoconf/manual/autoconf-2.68/html_node/Generic-Headers.html

and 

https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/Libraries.html 

My point - there are well know tools that check header and function availability and we don't have to introduce special defines for this.

nspark:
> there are well know tools that check header and function availability [...]

@shamisp I agree, but this topic has become a major issue with users (who are already using autotools). We should discuss it at the upcoming developer's workshop.

swpoole:
HAVE or HAS, sound fine. Another topic to discuss. Whatever the "norm" happens to be.

nspark:
@shamisp @jdinan Regarding autoconf/CMake, [this OSHMEM issue](https://github.com/open-mpi/ompi/issues/2483) highlights the issues that motivated this feature request. I agree that the feature _shouldn't_ be necessary.

I hope my OSHMEM issue resolves the issue for that implementation. Does anyone have suggestions on how we can encourage implementations to support standalone inclusion of `shmemx.h` for tests such as these?

jdinan:
What is the underlying cause for this issue?  Does their ```shmemx.h``` have a dependence on ```shmem.h```, but not include it?  If so, this seems like a bug in OSHMEM.

nspark:
Yes, they put `OSHMEM_DECLSPEC` in front of all their function declarations in [`shmemx.h`](https://github.com/open-mpi/ompi/blob/master/oshmem/include/shmemx.h) and use types from `stdint.h` without including it.

jdinan:
A program that does following will also fail to compile, right?
```
#include <shmemx.h>
#include <shmem.h>
...
```

BryantLam:
Correct.

It's good practice to make sure your header files are standalone units, but
it isn't technically required by the spec. We should discuss whether we
need to require it versus calling it out in an "Implementer's Note" type of
style.

shamisp:
To me it sounds line a OSHMEM engineering bug. I'm not sure that for each engineering bug in an implementation we have to modify spec. We can add a note for developers that  shmemx.h is expect to function as a standalone unit or something like this. I don't think we have to add special #defines for this.

nspark:
@shamisp I agree (re: OSHMEM bug) and have filed a GitHub issue for it with them.

I'm basically happy to withdraw this proposal, but am waiting on some internal testing before I take it _off_ the agenda for the next specification meeting.

nspark:
I've withdrawn this proposal for the specification and am closing the PR.

jdinan:
Withdrawn and not read at 12/15/16 meeting.

PR #28, opened 2016-11-28
Update SHMEM_SYNC_SIZE constant listing

jdinan:
List the constant once for all language bindings, instead of listing the
same thing twice for C/C++ and Fortran.

### DIFF
--- a/content/library_constants.tex
+++ b/content/library_constants.tex
@@ -11,10 +11,9 @@ All constants that start with
 \hline 
 \hline
 %new
-\vtop{\hbox{\CorCpp:}
-\hbox{\hspace*{12mm} \const{SHMEM\_SYNC\_SIZE}} 
-\hbox{\strut \Fortran:} 
-\hbox{\hspace*{12mm} \const{SHMEM\_SYNC\_SIZE}}} 
+\vspace{3mm}
+\vtop{\hbox{\CorCppFor:}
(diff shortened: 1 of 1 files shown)


jdinan:
@dkhaldi Could you please review/handle merge of this typo/formatting fix.

dkhaldi:
@jdinan Why this does not need a reading too? Is there an exception for "trivial" changes?

nspark:
Procedures (Proposals) 2.2.2-3(b):
> After the ballot, text changes that do not change the semantics of the proposal are permitted with the unanimous consent of the relevant section committee(s).

Have we established section committees yet?

Alternatively, (Ratification Process) 2.2.3-5(c):
> Formatting and whitespace problems, spelling errors, and other typos. Such problems should be itemized and can be fxed at the meeting by Section Committees and/or the OpenSHMEM Specifcation Document Editor.

jdinan:
@nspark thanks for looking this up.  We have not established section committees and typically would not do this until we want an editorial/integration pass over particular sections of the document.  Usually this doesn't happen until the ratification process starts up, but I could also imagine it happening sooner if there is a section that needs heavy editing/integration work.

I think the right way to handle this type of change is through the latter rule that Nick identified.  To strengthen that rule, we also state that the document editor "is responsible for both maintaining the overall document and its repository" (Sec. 1.2).  I would argue that correcting this sort of formatting error/typo falls under maintenance.

dkhaldi:
+1

PR #29, opened 2016-12-13
Updated "complex" to "_Complex"

nspark:
This removes any internal dependence on `complex.h`.


jdinan:
Can we add a loophole that these functions are optional when the compiler does not support C99?  Or is this already covered by something else?

jeffhammond:
Why do we need such a loophole?  Who is making such a terrible compiler?  At some point, software developers need to move on with their lives and expect that compiler developers can get within 17 years of the present when it comes to ISO language support.

🎻  ◀️  this is the world's tiniest violin playing a sad, sad song for compilers that are stuck in 1990.

jdinan:
I would also be fine with making C99 support mandatory.  The sized integer types that we are planning to add are also not an official part of C prior to C99.

BryantLam:
I don't really care for compilers before C99 either. The reverse problem actually occurs in this case. C99 requires implementation of `complex.h`, but C11 made it optional if `__STDC_NO_COMPLEX__` is defined. Do we want to respect a C11 compiler that does not have support for complex types, or require our C11 support to include complex types?

jdinan:
I don't think that complex support is critical enough for our users to make it mandatory.  I just need permission from the spec to omit these functions when the type is not supported.

jeffhammond:
Side note: I encourage everyone to allocate 3 votes to https://visualstudio.uservoice.com/forums/121579-visual-studio-ide/suggestions/3879386-c11, because MSVC is single-handedly holding back the world w.r.t. modern C.

nspark:
@jeffhammond FTFY:
> MSVC is single-handedly holding back ~~the world~~ _Windows developers_ w.r.t. modern C.

jeffhammond:
The issue is that developers who do not want to exclude Windows users or
downstream developers have to consider MSVC in the least common denominator
analysis. PETSc is a good example.

jdinan:
@nspark Can we add a statement somewhere, similar to the C11 statement, that these functions are included only when there is compiler/language support for the ```_Complex``` type?  Otherwise we are introducing a C99 language requirement.

jdinan:
@nspark Still waiting on a response to my question above -- this change effectively adds new language requirement (e.g. complex support is optional in C11).  Can we add a small bit of text indicating that these functions are only required when the ```_Complex``` type exists?

nspark:
@jdinan This PR isn't ready for reading. I wasn't able to finish it up in time. I'll remove the label.

manjugv:
@nspark Do you want to discuss this ticket or should be removed from the agenda ?

nspark:
@manjugv A (broader) discussion would be good. The change itself is relatively minor. I'll probably put together some improved draft text before or at the meeting. The bigger point of discussion is how to clearly _and concisely_ address language conformance in the OpenSHMEM specification. (Removing Fortran will help).

If we're being pedantic, we'll have to address things like (...and I'm sure there are more):

* `_Complex` is required by C99 but optional under C11 (thanks @BryantLam)
* The `[u]intN_t` fixed-width integer types are optional under C99/C11

jdinan:
@nspark We also need to define the behavior prior to C99, unless we take on C99 as the baseline for OpenSHMEM.  I think something simpler like "when type support is available" is clearer, perhaps with a footnote to capture the currently known variations/requirements of different C language standards.

nspark:
Currently, this PR does not address support for `std::complex<>` in C++. I'm inclined to think it should, given that the prototypes are labeled for "C/C++".

jdinan:
Do we really need to add a C++ special case for an API that no one uses?  I think this would be the first instance of C++ API in the entire spec.

content/shmem_reductions.tex
```diff
@@ -210,6 +210,14 @@
     Upon return from a reduction routine, the following are true for the local
     \ac{PE}: The \dest{} array is updated.  The values in the \VAR{pSync} array are
     restored to the original values.
+
+    % Update the \Clang and \Celev macros here
+    The sum and product reduction routines include complex-typed interfaces.
+    When the \Clang{} translation environment does not support complex types%
+    \footnote{That is, under \Clang{} language standards prior to \textit{C99}
+      or under \Celev{} when \CONST{\_\_STDC\_NO\_COMPLEX\_\_} is defined to 1},
+    an \openshmem implemementation is not required to provide support for
```
nspark: s/implemementation/implementation/

nspark:
No, we don't, and you're right that it's probably a good idea not to.

For this PR, the reductions description adds the following:
> The sum and product reduction routines include complex-typed interfaces. When the C translation environment does not support complex types[1], an OpenSHMEM implementation is not required to provide support for these complex-typed interfaces.
>
> [1]: That is, under C language standards prior to C99 or under C11 when `__STDC_NO_COMPLEX__` is defined to 1.

One potential revision might be:
> The sum and product reduction routines include complex-typed interfaces **for the C API only**. When the C translation environment does not support complex types, an OpenSHMEM implementation is not required to provide support for these complex-typed interfaces.

nspark:
@jdinan Ready to merge!

content/backmatter.tex
```diff
@@ -572,6 +572,14 @@ \section{Version 1.4}
       operations.
 \\ See Section \ref{sec:amo}.
 %
+\item Replaced the \textbf{complex} macro in complex-typed reductions with the
+      \Cstd[99] (and later) type specifier \textbf{\_Complex} to remove an
```
jdinan: Should we replace ```\textbf{complex}``` with ```\CTYPE{complex}```, and similarly for ```_Complex```?

nspark: Yeah, nice catch.

jdinan:
Can I get a +1 on commit e6a567d to include this doc edit in the merge?  Thanks!

PR #30, opened 2016-12-15
Update example codes

jamesaross:
After reviewing and running all of the examples, I noticed there were several bugs with non-deterministic execution or incorrect reporting.  The examples were also inconsistent in formatting and style.  Some used barriers or extra code which were not required.  Others could be simplified and have explicit barriers removed.  Many used global variables and arrays when statically defined variables and arrays would be sufficient and simpler.  There are also codes that use `shmem_malloc()` and `shmem_free()` when a simple static variable or array declaration would work.  A couple codes were updated to use the C11 Generics.

I have incorporated James Dinan's earlier changes which passed the committee vote.

Below is a summary table.  All codes generally had minor formatting improvements and were updated with C99-style declaration and initialization.

file |bug fix|global to static w/ initialization|2 to 3 space indent|4 to 3 space indent|tab to 3 space indent|remove explicit barrier|malloc to static|simplify main()|C11 Generic|
|-|-|-|-|-|-|-|-|-|-|
| hello-openshmem.c| | |X| | | | |X| |
| shmem_add_example.c|X| | | | |X| | | |
| shmem_alltoall_example.c|X|X| |X| |X| | | |
| shmem_alltoalls_example.c|X|X| |X| |X| | | |
| shmem_barrier_example.c|X|X| | | | | | | |
| shmem_barrierall_example.c| |X| | | | | | | |
| shmem_broadcast_example.c| |X| | | |X| | | |
| shmem_collect_example.c|X|X| | | | | | | |
| shmem_cswap_example.c| | | | | | | | | |
| shmem_fadd_example.c| | | | | |X| | | |
| shmem_fence_example.c| |X|X| | | | | | |
| shmem_finalize_example.c| |X| | | | | | |X|
| shmem_finc_example.c| |X| | | | | | | |
| shmem_g_example.c| |X| | | | | | | |
| shmem_global_exit_example.c| | |X| | | | | | |
| shmem_inc_example.c|X| | | | |X| | | |
| shmem_iput_example.c| | | | | | | | | |
| shmem_lock_example.c| |X| | | | | |X|X|
| shmem_mype_example.c| | |X| | | | | | |
| shmem_npes_example.c| | |X| | | | | | |
| shmem_p_example.c| |X| | | |X|X| | |
| shmem_ptr_example.c| | | | | | | | | |
| shmem_put_example.c| | | | | | | | | |
| shmem_quiet_example.c| |X|X| | | | | | |
| shmem_swap_example.c| | | | | | |X| | |
[...]

PR #31, opened 2017-01-10
Add bitwise and, or, xor AMOs

nspark:
This PR:
* adds fetching and non-fetching bitwise and, or, xor atomic memory operations
* normalizes and renames the AMO API to use the `shmem_atomic_*` or `shmem_atomic_fetch_*` prefixes
* only affects the C API 

This PR also depends on PRs #13 and #55 .

**Special Ballot**: diff ([547f490..764633b](https://github.com/openshmem-org/specification/pull/31/files/547f49065ea9c285787375386f84baa2dc8f1602..764633b08a6c7f8884b6721b86dba893718d3411)) and [PDF](https://github.com/openshmem-org/specification/files/839217/proposal-bitwise-amos-20170313.pdf)

**To Do**:

- [x] [Deprecation annex](https://github.com/openshmem-org/specification/pull/31#discussion_r101554775)
- [x] [Minor text revision](https://github.com/openshmem-org/specification/pull/31#discussion_r101558103)
- [x] [Generic RMA cleanup](https://github.com/openshmem-org/specification/pull/31#discussion_r101574443)


jamesaross:
@nspark 
The full section for `shmem_fadd()` seems to be missing in the PR and doesn't appear in the partial PDF you emailed.

jeffhammond:
One solution to `shmem_for` is to make the inclusive aspect of OR explicit, i.e. name the function `shmem_fior`.  An argument for this is that `IOR` is the name of the [Fortran intrinsic](https://gcc.gnu.org/onlinedocs/gfortran/IOR.html#IOR).  An argument against this is that it would be inconsistent with `AND`, for which Fortran has the `IAND` intrinsic.

However, I wonder why one wouldn't want the bitwise specified explicitly with a `b` the way [MPI does](http://mpi-forum.org/docs/mpi-1.1/mpi-11-html/node78.html).  Even if we never add logical operations, the `b` is another way to avoid `shmem_for`.

I will also point out that I have found `shmem_and` and `shmem_or` used by SHMEM implementations for collectives.

I agree with your decision to avoid Fortran.  I'm don't know any bit-manipulation-happy programmers who use Fortran.

manjugv:
Nitpick : For readability, can you consider making the operations (AND, OR, XOR) caps/bold in the descriptions ? May be, you can define a macro, so we change if we need to.

Can you comment on the hardware support for bitwise operations ? Cray’s uGNI interface seems to support it. But not sure if it provides the hardware support or just a software layer. MLNX ? Intel ?

Backward compatibility constraints aside, what do you think of having one 
(or two - one for fetch type and another for non-fetch type) atomic interface for all these operations ? 
Something like : shmem_<TYPE>_atomic(TYPE *dest, ATOMIC_OP op, TYPE value, int pe)

nspark:
@jamesaross Your comment did make me realize that I forgot to include the (tiny) changes for `shmem_fadd()` and `shmem_add()`, which mostly amounted to a `s/standard/extended/` replacement. Sorry about that and any confusion.

@jeffhammond Hmm. I don't personally prefer those potential names over `shmem_for()`. The following potential names seem a bit long, but they're growing on me. They also seem stylistically consistent with [`stdatomic.h`](http://en.cppreference.com/w/c/atomic). I'd also include "new style" variants of the existing AMOs and deprecate the "old style" API.
```C
// Generic API
TYPE shmem_atomic_fetch_or(TYPE *dest, TYPE value, int pe);
void shmem_atomic_or      (TYPE *dest, TYPE value, int pe);
// Non-generic example
long shmem_long_atomic_fetch_or(long *dest, long value, int pe);
void shmem_long_atomic_or      (long *dest, long value, int pe);
```

@manjugv I don't especially mind capitalizing the operations. We also discussed hardware support a while ago [on Redmine](http://www.openshmem.org/redmine/issues/172#note-21). Both Cray and Mellanox can support these operations.

> Backward compatibility constraints aside, what do you think of having one (or two - one for fetch type and another for non-fetch type) atomic interface for all these operations ?

I thought of such an interface, too. The interfaces I proposed here are motivated in part out of consistency with the existing API and in part out of consistency with the explicit naming of the C's `stdatomic.h` API.

content/shmem_for.tex
```diff
@@ -0,0 +1,45 @@
+\apisummary{
+  Atomically perform a fetching bitwise inclusive or operation on a
+  remote data object.
+}
+
+\begin{apidefinition}
+
+\begin{C11synopsis}
+TYPE shmem_for(TYPE *dest, TYPE value, int pe);
```
jdinan: shmem_for ... 🙈 🙉 🙊

content/shmem_for.tex
```diff
@@ -0,0 +1,45 @@
...
+  \apiargument{OUT}{dest}{A pointer to the remotely accessible data object to
+    be updated.}
+  \apiargument{IN}{value}{The operand to the bitwise atomic operation.}
+  \apiargument{IN}{pe}{An integer value for the \ac{PE} on which \VAR{dest}
+    is to be updated.}
+
+\end{apiarguments}
+
+\apidescription{
+  \FUNC{shmem\_for} atomically performs a fetching bitwise inclusive or on
```
jdinan: OR

content/shmem_fxor.tex
```diff
@@ -0,0 +1,45 @@
+\apisummary{
+  Atomically perform a fetching bitwise exclusive or operation on a
```
jdinan: ... bitwise exclusive OR (XOR) operation ...

content/shmem_fxor.tex
```diff
@@ -0,0 +1,45 @@
...
+TYPE shmem_<TYPENAME>_fxor(TYPE *dest, TYPE value, int pe);
+\end{Csynopsis}
+where \TYPE{} is one of the standard \ac{AMO} types and has a corresponding
+\TYPENAME{} specified by Table \ref{stdamotypes}.
+
+\begin{apiarguments}
+
+  \apiargument{OUT}{dest}{A pointer to the remotely accessible data object to
+    be updated.}
+  \apiargument{IN}{value}{The operand to the bitwise atomic operation.}
```
jdinan: "bitwise atomic operation" seems a bit lazy.  Can we replace this with "bitwise XOR operation"

content/shmem_fxor.tex
```diff
@@ -0,0 +1,45 @@
...
+  \apiargument{IN}{pe}{An integer value for the \ac{PE} on which \VAR{dest}
+    is to be updated.}
+
+\end{apiarguments}
+
+\apidescription{
+  \FUNC{shmem\_fxor} atomically performs a fetching bitwise exclusive or on
```
jdinan: s/or/OR/

main_spec.tex
```diff
@@ -163,6 +163,24 @@ \subsubsection{\textbf{SHMEM\_FETCH}}\label{subsec:shmem_fetch}
 \subsubsection{\textbf{SHMEM\_SET}}\label{subsec:shmem_set}
 \input{content/shmem_set.tex}
```
jdinan: Could you fix the ordering of shmem_add and shmem_fadd so they appear together and with the fetching version first (CSWAP, SWAP, FINC, INC, FADD, ADD, ...)?  Also, should the bitwise ops appear before fetch/set?

content/shmem_fxor.tex
jdinan: Ok, I'll stop repeating these capitalization comments and assume you can find the others.

content/shmem_for.tex
tonycurtis: Sounds more like a loop rather than an atomic op....

nspark: I'm leaning toward overhauling the AMOs with the naming scheme I sketched in [this comment](#issuecomment-273598859). I think the explicit nature of the names is generally an improvement (despite their length) over the short-form names we currently have (which don't explicitly mention atomicity) because they clearly convey the intent of the function call.

I think this probably makes sense to issue as two PRs: one for bitwise AMOs with the "new style" names and one to rename existing AMOs (deprecate "old style" and add "new style" names).

Thoughts?

content/shmem_and.tex
```diff
@@ -0,0 +1,44 @@
...
+  remote data object.
+}
+
+\begin{apidefinition}
+
+\begin{C11synopsis}
+void shmem_and(TYPE *dest, TYPE value, int pe);
+\end{C11synopsis}
+where \TYPE{} is one of the standard \ac{AMO} types specified by
+Table~\ref{stdamotypes}.
```
nspark: The C standard says that signed integer types (&sect;6.2.5-4) have an _implementation-defined_ representation (&sect;6.2.6.2-2), so I don't think OpenSHMEM is overstepping and defining something that is undefined by the language. This would only seem to be problematic if the signed integer representations of the CPU and NIC did not match, which seems unlikely.

In general, I doubt we'll run into much trouble with regards to signed integer representation. Quoting [Wikipedia](https://en.wikipedia.org/wiki/Signed_number_representations#History):
> [...] As IC technology advanced, virtually all [processors] adopted two's complement technology. x86, m68k, Power Architecture, MIPS, SPARC, ARM, Itanium, PA-RISC, and DEC Alpha processors are all two's complement.

content/shmem_and.tex
```diff
@@ -0,0 +1,44 @@
...
+\end{Csynopsis}
+where \TYPE{} is one of the standard \ac{AMO} types and has a corresponding
+\TYPENAME{} specified by Table \ref{stdamotypes}.
+
+\begin{apiarguments}
+
+  \apiargument{OUT}{dest}{A pointer to the remotely accessible data object to
+    be updated.}
+  \apiargument{IN}{value}{The operand to the bitwise atomic operation.}
+  \apiargument{IN}{pe}{An integer value for the \ac{PE} on which \VAR{dest}
```
nspark: I _can_ change it to "PE number of the remote PE. _pe_ must be of type integer." I'm not going to say my text is much better, but I find the whole argument-description section rather irritating in the OpenSHMEM specification. It generally seems redundant and unnecessary, given a well-written description body. I guess that's a different battle to fight, or at least not one this PR will solve.

So, should this text be consistent with the rest of the AMOs, or with the RMA descriptions?

content/shmem_and.tex
jdinan: My concern is that we are adding functions (e.g. shmem_long_and) that have implementation-defined behavior.  Portable applications should never use use these functions -- so why add them in the first place?

content/shmem_and.tex
jdinan: For now, we should be consistent with the rest of the AMO section.  The chapter committees could standardize the argument description text across the full document once we get into ratification.

content/shmem_for.tex
jdinan: I prefer the new names.  Unless you think there will be a procedural advantage to splitting these up, I prefer keeping them together.  I think we will want to have both sets of changes in the spec for the next release, so it makes sense to me that they stay together.

jeffhammond:
@manjugv I dislike `shmem_atomic(TYPE *dest, ATOMIC_OP op, TYPE value, int pe)` because it entails extra branching that reduces message rate.

manjugv:
@jeffhammond  Yeah, I thought about it when proposing. My experience is that the issue rate that can be achieved for atomics is mostly lesser than the RDMA message rate on current hardware, even when atomics are supported by the hardware. I’m guessing there are more overheads in the case of atomics. When they are not supported by the hardware, it is going to be implemented by a CAS loop, so an extra branch might be OK. 

Even though intuitively it makes sense and I agree with you. I haven’t able to conclusively quantify the impact of a branch on the atomic issue rate. If you have seen something conclusive or quantified, please share.

jeffhammond:
@manjugv My guess is Blue Gene/Q will show the biggest effect here but I don't have time to measure.  You ask Paul Coffman if he is interested in testing Sandia OpenSHMEM there.  I don't know if the OFI implementation supports network hardware atomics yet.

nspark:
Here is a PDF of [this PR as of e69ef2c](https://github.com/openshmem-org/specification/files/758678/proposal-bitwise-amos-20170207.pdf).

content/backmatter.tex
```diff
@@ -373,6 +373,18 @@ \section{Version 1.5}
 \item The C11 \textbf{\_Noreturn} function specifier was added to
       \FUNC{shmem\_global\_exit}.
 \\ See Section \ref{subsec:shmem_global_exit}.
+%
+\item Expanded the type support for RMA and AMO operations, including support
+      for atomic add and fetch-and-add on floating-point types.
+\\ See Sections \ref{sec:rma} and \ref{sec:amo}.
+%
+\item Renamed AMO operations to use \FUNC{shmem\_atomic\_*} prefix and
+      deprecated old AMO routines.
```
jdinan: Did you add the deprecated interfaces to the deprecation annex?

content/shmem_atomic_compare_swap.tex
```diff
@@ -5,14 +5,16 @@
 \begin{apidefinition}
 
 \begin{C11synopsis}
-TYPE shmem_cswap(TYPE *dest, TYPE cond, TYPE value, int pe);
+TYPE shmem_atomic_compare_swap(TYPE *dest, TYPE cond, TYPE value, int pe);
```
jdinan: Minor comment -- why change ```cswap``` to ```compare_swap```?

content/shmem_atomic_add.tex
```diff
@@ -53,14 +55,17 @@
 }
 
 \apinotes{
-    The term remotely accessible is defined in Section \ref{subsec:memory_model}.
+    As of OpenSHMEM~1.4, the use of \FUNC{shmem\_add} has been deprecated
+    and should be expected to be removed in a future specification.
+    Its behavior and call signature is identical to the replacement
+    interface, \FUNC{shmem\_atomic\_add}.
```
jdinan: The deprecation statement is present for ```shmem_add``` and a couple others, but not all of deprecated atomic routines.  This seems inconsistent?  It's also a little weird that ```shmem_add``` is mentioned here, but the function itself (i.e. its prototype) is absent from the specification.

content/shmem_atomic_compare_swap.tex
jdinan: Oh, this is because of ```shmem_atomic_for```...

content/shmem_atomic_compare_swap.tex
```diff
@@ -59,7 +61,10 @@
 }
 
 \apinotes{
-    None.
+    As of OpenSHMEM~1.4, the use of \FUNC{shmem\_cswap} has been deprecated
+    and should be expected to be removed in a future specification.
```
jdinan: s/should be expected to be/will be/

content/rma_intro.tex
```diff
@@ -16,31 +16,40 @@
...
+      int8\_t            & int8       \\ \hline
+      int16\_t           & int16      \\ \hline
+      int32\_t           & int32      \\ \hline
+      int64\_t           & int64      \\ \hline
+      uint8\_t           & uint8      \\ \hline
+      uint16\_t          & uint16     \\ \hline
+      uint32\_t          & uint32     \\ \hline
+      uint64\_t          & uint64     \\ \hline
+      intptr\_t          & intptr     \\ \hline
+      uintptr\_t         & uintptr    \\ \hline
```
nspark: Note to remove `intptr_t` and `uintptr_`.

content/backmatter.tex
nspark: No, I can add that on as a post-reading change.

content/shmem_atomic_compare_swap.tex
nspark: The change from `cswap` to `compare_swap` was to be stylistically closer to the C11 [`atomic_compare_exchange`](http://en.cppreference.com/w/cpp/atomic/atomic_compare_exchange).

content/shmem_atomic_and.tex
```diff
@@ -0,0 +1,44 @@
...
+  Atomically perform a non-fetching bitwise AND operation on a
+  remote data object.
+}
+
+\begin{apidefinition}
+
+\begin{C11synopsis}
+void shmem_atomic_and(TYPE *dest, TYPE value, int pe);
+\end{C11synopsis}
+where \TYPE{} is one of the standard \ac{AMO} types specified by
```
jdinan: Replace "standard" with "bitwise"

content/shmem_atomic_and.tex
```diff
@@ -0,0 +1,44 @@
...
+Table~\ref{bitamotypes}.
+
+\begin{Csynopsis}
+void shmem_<TYPENAME>_atomic_and(TYPE *dest, TYPE value, int pe);
+\end{Csynopsis}
+where \TYPE{} is one of the standard \ac{AMO} types and has a corresponding
```
jdinan: Replace "standard" with "bitwise" (also in other bitwise ops)

content/shmem_atomic_add.tex
```diff
@@ -36,10 +38,10 @@
 \end{apiarguments}
 
 \apidescription{
-    The \FUNC{shmem\_add} routine performs an atomic add operation. It adds
+    The \FUNC{shmem\_atomic\_add} routine performs an atomic add operation. It adds
```
BryantLam: Style suggestion: `AND`, `OR`, and `XOR`, but `add` (`ADD`?)

content/shmem_atomic_fetch_xor.tex
```diff
@@ -0,0 +1,45 @@
+\apisummary{
+  Atomically perform a fetching bitwise exclusive or (XOR) operation on a
```
jdinan: Replace "or" with "OR"

content/shmem_atomic_xor.tex
```diff
@@ -0,0 +1,44 @@
+\apisummary{
+  Atomically perform a non-fetching bitwise exclusive or (XOR) operation on a
```
jdinan: Replace "or" with "OR"

content/shmem_atomic_add.tex
```diff
@@ -53,14 +55,17 @@
 }
 
 \apinotes{
-    The term remotely accessible is defined in Section \ref{subsec:memory_model}.
+    As of OpenSHMEM~1.4, the use of \FUNC{shmem\_add} has been deprecated
```
jdinan: Use the ```\openshmem{}``` macro

content/shmem_atomic_inc.tex
```diff
@@ -49,14 +51,18 @@
 }
 
 \apinotes{
-     The term remotely accessible is defined in Section \ref{subsec:memory_model}.
+    As of OpenSHMEM~1.4, the use of \FUNC{shmem\_inc} has been deprecated
+    and should be expected to be removed in a future specification.
```
jdinan: Remove "and should be expected to be removed in a future specification."

content/shmem_atomic_swap.tex
```diff
@@ -60,15 +60,18 @@
...
+    interface, \FUNC{shmem\_atomic\_swap}.
 }
 
 \begin{apiexamples}
 
 \apicexample
-    {The example below swap values between odd numbered \acp{PE} and their right
-    (modulo) neighbor and outputs the result of swap.}
-    {./example_code/shmem_swap_example.c}
+    {The example below swap values between odd numbered \acp{PE} and
```
jdinan: s/swap/swaps/
Use plural of PE (there is a latex macro for this).

content/shmem_atomic_fetch_add.tex
```diff
@@ -59,14 +61,18 @@
...
+    interface, \FUNC{shmem\_atomic\_fetch\_add}.
 }
 
 \begin{apiexamples}
 
 \apicexample
-	{The following \FUNC{shmem\_fadd} example is for \CorCpp{} programs:}
-        {./example_code/shmem_fadd_example.c}
+	{The following \FUNC{shmem\_atomic\_fetch\_add} example is for
+        \CorCpp{} programs:}
```
jdinan: The example was updated to use C11 generic selection, so the C++ part is no longer true.  Fix is to update ```\CorCpp{}``` to ```\Celev{}```.

@jamesaross will apply the same change to #30.

content/atomics_intro.tex
```diff
@@ -4,68 +4,102 @@
...
-      \hline
-      int & int\\
-      \hline
-      long & long\\
-      \hline
-      long long & longlong\\
-      \hline
+      \TYPE              & \TYPENAME  \\ \hline
+      float              & float      \\ \hline
+      double             & double     \\ \hline
```
shamisp: I'm a bit hesitant about introduction of floating point ADD/FADD AMOs.

Is this something that can be resolved by fixing teams ? If this is mostly used for a timer, I think this a wrong  "tool".

nspark: It's just one data point, but -- as you noted in conversation and I verified -- the Intel x86 ISA shows that the LOCK prefix may not be used on float and double types. I'll remove floating-point AMO support.

nspark: @shamisp Fixed in b9650b2

content/shmem_atomic_add.tex
nspark: :man_shrugging:

content/backmatter.tex
nspark: Complete as of 54ab60b

nspark:
Here is a [PDF of this PR as of a562934](https://github.com/openshmem-org/specification/files/838223/proposal-bitwise-amos-20170313.pdf).

nspark:
Here is a [PDF of this PR as of f2c43e2](https://github.com/openshmem-org/specification/files/838371/proposal-bitwise-amos-20170313.pdf)

nspark:
Here's a [PDF of this PR as of 764633b](https://github.com/openshmem-org/specification/files/839217/proposal-bitwise-amos-20170313.pdf).

jdinan:
@nspark The full text of your proposal must be available at the two-week meeting cutoff, so any changes since the cutoff (unless they are a document edit) are not eligible for a special ballot or reading at the upcoming meeting (sorry).  The reading you already did still counts; you can bring that text with special ballots forward for voting at the next meeting.

nspark:
@jdinan These can wait until the next meeting, but that's not how I read the specification procedures (which means maybe we need to clarify the procedures).

> 3. Changes to proposal text after it was made available for the formal reading (i.e., at least two weeks prior to the start date of the official OpenSHMEM Committee meeting at which it was read) are permitted in some cases:
> 
>    a. After the formal reading and before the ballot, changes are permitted if the text delta is presented at a official OpenSHMEM Committee meeting and approved via a special formal ballot of OOE organizations at that meeting: [...]
> 

The changes I'm submitting for special ballot happened between the formal reading and the ballot. As written, I understand the procedures to only require the two-week cutoff for the formal reading. I don't see anything specifying a cutoff for changes to a special ballot.

jdinan:
You're correct; the changes are eligible for a special ballot.  However, there's another requirement on scheduling the special ballot in Section 2.2.1:
```
All official ballots must be announced and scheduled at least two weeks prior (four weeks prior
in case of votes for a final OpenSHMEM Specification Document) to the start date of the official
OpenSHMEM Committee meeting at which they will be held.
```
In this case "special" is referring to the fact that the change is being expedited through the process and is being read and voted on in the same meeting.  It is still an official (i.e. not straw) ballot.

The practical argument for this is that if we didn't hold special ballots to the two week requirement, changes could be introduced without providing representatives adequate time to review the proposal within their respective organizations prior to casting the organization's vote.

nspark:
@jdinan Thanks for clarifying. I'll review my PRs and remove those from the agenda that changed after the agenda deadline.

PR #24, opened 2016-10-06
Semantics clarification for visibility ordering guarantees.

shamisp:
Redmine issue http://www.openshmem.org/redmine/issues/169

PR #32, opened 2017-01-11
Overhaul of point-to-point synchronization API

nspark:
This PR updates the point-to-point synchronization API by deprecating `shmem_wait()` in favor of `shmem_wait_until()` and adding `shmem_test()`.

**Special Ballot**: diff ([73fb599..bc75342](https://github.com/openshmem-org/specification/pull/32/files/73fb59987c8501b243080bfcdf11f26cfaaae6df..bc75342397d7788ab2a7aa5a11bb9ee98bc89a83)) and [PDF](https://github.com/openshmem-org/specification/files/838537/proposal-p2p-sync-20170313.pdf)

To Do:
- [X] Added p2p-sync intro section for common text
- [X] Add `shmem_test()`
- [x] Deprecate `shmem_wait()`
- [x] Update change log
- [x] Update deprecation annex
- [x] Clean up deprecation styling for `shmem_wait()`
- [x] Add `unsigned short`; remove `float` and `double`
- [x] Address [synchronization ordering](https://github.com/openshmem-org/specification/pull/32#pullrequestreview-23255588)
- [x] Address `stdint.h` and C99/C11 dependence
- [x] Rename `shmem_cmp` consistent with #56 

This PR is associated with [Redmine Issue 215](http://www.openshmem.org/redmine/issues/215).


BryantLam:
After playing around with this API, I'm in favor of using a bool-based return code for `shmem_test` and siblings, even if we later decide to add error-based return codes to everything *except* `shmem_test`:

```c
typedef /* assumed int-based type */ shmem_error_t;
int shmem_test(TYPE *ivar, TYPE value, shmem_error_t *err);

void user_wait(int *ivar, int value) {
  shmem_error_t err;
  while (shmem_test(ivar, value, &err) && !err) {}
  if (err) {
    // Handle error.
  }
}

// Split codepaths for error handling! Ugly!!
void test_and_do_other_things(int *ivar, int value) {
  shmem_error_t err;
  while (shmem_test(ivar, value, &err) && !err) {
    if (err) {
      // What's shmem_test's return value when an error occurs?
      // Do we make it legal to have shmem_test return 1
      // and also an error, or can we just forbid it?
      goto error_handle;
    }
    do_work(get_work());
  }
  if (err) {
    goto error_handle;
  }
  do_postprocess();
  do_work(get_remaining_work());
  return;
error_handle:
  // Error handling.
  return;
}

// ALTERNATIVELY, if return code cannot be 1 on error:
void test_and_do_other_things(int *ivar, int value) {
  shmem_error_t err;
  while (shmem_test(ivar, value, &err) /* NOT NEEDED: && !err */) {
    do_work(get_work());
  }
  if (err) {
    // Error handling.
  }
  do_postprocess();
  do_work(get_remaining_work());
  return;
}
```

So a return-based API for `shmem_test` is still attractive as long as the return code cannot be 1 on error.

nspark:
After thinking more about the test-and-wait family of functions, I'm going to revise my proposal for an API with the following signatures (generic API only listed here):

```C
// existing
void shmem_wait(TYPE *ivar, TYPE cmp_value);
void shmem_wait_until(TYPE *ivar, int cmp, TYPE cmp_value);
// new wait
int    shmem_wait_all      (TYPE *ivars, size_t nelems, TYPE cmp_value);
size_t shmem_wait_any      (TYPE *ivars, size_t nelems, TYPE cmp_value);
int    shmem_wait_until_all(TYPE *ivars, size_t nelems, int cmp, TYPE cmp_value);
size_t shmem_wait_until_any(TYPE *ivars, size_t nelems, int cmp, TYPE cmp_value);
// new test
int    shmem_test(TYPE *ivar, int cmp, TYPE cmp_value);
int    shmem_test_all(TYPE *ivars, size_t nelems, int cmp, TYPE cmp_value);
size_t shmem_test_any(TYPE *ivars, size_t nelems, int cmp, TYPE cmp_value);
```

The biggest pattern change is that the `shmem_test()` family takes an argument to specify the comparison operation (as with the `shmem_wait_until()` family). At first, I thought about adding both `shmem_test()` and `shmem_test_cmp()` (because `test_wait()` didn't make sense), but then felt like this would just cause confusion overall.

If one was to emulate `shmem_wait()` with a `shmem_test()` that doesn't take a comparison operation argument, then either:
```C
// shmem_test() is implicitly SHMEM_CMP_EQ,
// and the boolean negation seems odd
void user_wait(long *ivar, long value) {
  while (!shmem_test(ivar, value));
}

// shmem_test() is implicitly SHMEM_CMP_NE,
// which just seems likely to cause confusion
void user_wait(long *ivar, long value) {
  while (shmem_test(ivar, value));
}
```

To me, the clearest interface is just to expose the comparison operation as an argument to the entire family of test functions.

Feedback? Comments? Concerns?

nspark:
After talking it over with some users, we think the `shmem_*_{any,all}()` variants do not seem especially helpful (at least not for now). Further, we discussed how the semantic for `shmem_wait()` is actually rather confusing, and that `shmem_wait_until()` much more clearly conveys the user's intention.

Thus, I'll be revising this proposal to deprecate `shmem_wait()` and only add `shmem_test()`.


```C
// deprecate
void shmem_wait(TYPE *ivar, TYPE cmp_value);
// existing
void shmem_wait_until(TYPE *ivar, int cmp, TYPE cmp_value);
// new test
int shmem_test(TYPE *ivar, int cmp, TYPE cmp_value);
```

Poking @jdinan @jeffhammond, since Jim raised this and Jeff commented on Redmine.

nspark:
PDF of specification for [this PR as of 73fb599](https://github.com/openshmem-org/specification/files/758901/proposal-p2p-sync-20170207.pdf).

content/shmem_wait.tex
```diff
@@ -69,26 +69,16 @@
...
 \apireturnvalues{
     None.
 }
 
 \apinotes{
-    The \FUNC{shmem\_wait} and \FUNC{shmem\_wait\_until} routines can be used to
-    portably ensure that memory access operations observe remote updates in the order
-    enforced by the initiator PE using the \FUNC{shmem\_fence} and \FUNC{shmem\_quiet} routines.
+  As of OpenSHMEM~1.4, the \FUNC{shmem\_wait} routine has been deprecated
+  and should be expected to be removed in a future specification.
```
jdinan: Need to also update the deprecation annex.

content/shmem_wait.tex
```diff
@@ -5,15 +5,15 @@
...
+[[deprecated]] void shmem_wait(TYPE *ivar, TYPE cmp_value);
+void shmem_wait_until(TYPE *ivar, shmem_cmp cmp, TYPE cmp_value);
 \end{C11synopsis}
 where \TYPE{} is one of the point-to-point synchronization types specified by
 Table \ref{p2psynctypes}.
 
 \begin{Csynopsis}
-void shmem_<TYPENAME>_wait(TYPE *ivar, TYPE cmp_value);
-void shmem_<TYPENAME>_wait_until(TYPE *ivar, int cmp, TYPE cmp_value);
+[[deprecated]] void shmem_<TYPENAME>_wait(TYPE *ivar, TYPE cmp_value);
```
nspark: IIRC, this is the first routine that is deprecated outright. Others were deprecated with (more or less) a renamed replacement. I'd rather remove the `shmem_wait()` listing than add `[[deprecated]]` to all the other deprecated routines.

content/p2p_sync_intro.tex
```diff
@@ -0,0 +1,62 @@
...
+selection. Such type-generic routines are supported for the
+``point-to-point synchronization types'' identified in
+Table~\ref{p2psynctypes}.
+
+\begin{table}[h]
+  \begin{center}
+    \begin{tabular}{|l|l|}
+      \hline
+      \TYPE              & \TYPENAME  \\ \hline
+      float              & float      \\ \hline
```
spotluri: group agreement to remove float/double

content/p2p_sync_intro.tex
```diff
@@ -0,0 +1,62 @@
...
+      double             & double     \\ \hline
+      short              & short      \\ \hline
```
spotluri: missing unsigned short

example_code/shmem_test_example1.c
```diff
@@ -0,0 +1,7 @@
+#include <stdlib.h>
```
jdinan: I don't think ```stdlib.h``` is needed by this example.

content/p2p_sync_intro.tex
shamisp: I don't think we know how to support it. For this reason we should avoid introducing this

BryantLam:
Notes
- Add `unsigned short` to types table
- Remove `float` and `double`
- Use similar deprecation styling for Fortran
- Missing deprecation entry in Annex F

content/shmem_test.tex
```diff
@@ -0,0 +1,53 @@
...
+  comparison operator \VAR{cmp}.
+}
+
+\apireturnvalues{
+  \FUNC{shmem\_test} returns 1 if the comparison of the symmetric object
+  pointed to by \VAR{ivar} with the value \VAR{value} according to the
+  comparison operator \VAR{cmp} evalutes to true; otherwise, it returns 0.
+}
+
+\apinotes{
```
shamisp: It would be helpful to add a clarification about sync logic, similar to the one that we have in shmem_wait:

```
The \FUNC{shmem\_test} routine can be used to
portably ensure that memory access operations observe remote updates in the order
enforced by the initiator PE using the \FUNC{shmem\_fence} and \FUNC{shmem\_quiet} routines.
```

shamisp: shmem_fence and shmem_quiet routines description also has to be updated. 
We should add reference to the shmem_test.
Please see:
https://github.com/openshmem-org/specification/pull/24/files#diff-9f0ce1d74673e29019cee65d95111fcfR50

Adding @spotluri

nspark: That text was lifted from `shmem_wait[_until]()` and added to the [introductory text](https://github.com/openshmem-org/specification/pull/32/files/73fb59987c8501b243080bfcdf11f26cfaaae6df#diff-b039d6c2bf9af7510d362563dc21966aR4) to cover all the point-to-point synchronization routines.

shamisp: In this case shmem_wait/fence/quite files have to be updated as well

nspark:
[PDF of this PR as of bc75342](https://github.com/openshmem-org/specification/files/838537/proposal-p2p-sync-20170313.pdf)

jdinan:
Todo: #68

PR #33, opened 2017-01-19
Add Fortran MY_PE and NUM_PES routines

jdinan:
These routines were removed in error between v1.1 and v1.2.
Technically, they should be deprecated since the corresponding C
routines _my_pe() and _num_pes() were deprecated.


jdinan:
Replaced by #35

PR #34, opened 2017-01-20
Annex F: Expand SMA entry

BryantLam:
Expands the SMA entry in the deprecation table and adds a deprecation rationale in the same style as the one for _my_pe and friends.


BryantLam:
No changelog entry because this change supercedes an existing PR that listed the change.

PR #25, opened 2016-10-07
Extension #202: Improved Annex F for Deprecated API

BryantLam:
Redmine: http://openshmem.org/redmine/issues/202

The base commit for this pull request is essentially unchanged from #202 annexF_revision_proposal2.pdf (07/13/2016 01:11 PM) from the Redmine. Changes atop this base commit reflect feedback from the OpenSHMEM Workshop 2016 and misc additions.
- WORKSHOP: Simplified rationale text for **SHMEM_CACHE**
- NEW: Changed rationale text for **_SHMEM_\* constants**
- NEW: Table includes language specifiers for previously deprecated functions
- NEW: Changelog text (Where did v1.4 go?)
- Misc formatting

TODO: Table could still use some formatting wizardry. That's for another day.

PR #35, opened 2017-01-20
Annex F: Add missing Fortran routines

BryantLam:
OpenSHMEM v1.2 deprecated these Fortran routines: START_PES, MY_PE, NUM_PES. These routines were missed when the Annex F table was labeled with `C/C++` and `Fortran` attributes in #25. The routines have been added back in as unique entries along with a rationale that seems appropriate until we officially deprecate the existing Fortran bindings.

Related to #33 in addendum to or as a potentially satisfactory resolution.

### DIFF
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -377,24 +377,29 @@ specification.
 
 \section{Deprecation Rationale}\label{subsec:dep_rationale}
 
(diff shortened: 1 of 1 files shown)


BryantLam:
No changelog entry. This PR adds missing Fortran declarations that were missed when PR #25 was accepted.

PR #37, opened 2017-01-23
Init/Finalize Clarification

jdinan:
Clarify that ```shmem_init``` must be matched with a call to ```shmem_finalize```.


jdinan:
Remove the following from finalize notes:

> There is an implicit shmem_finalize at the end of main, so that having an
explicit call to shmem_finalize is optional. However, an explicit shmem_finalize may be required as an
entry point for wrappers used by profiling or other tools that need to perform their own finalization.

jdinan:
Update mention of shmem_finalize in Annex A

content/backmatter.tex
```diff
@@ -440,6 +440,10 @@ \section{Version 1.5}
 
 \begin{itemize}
 %
+\item Clarified that \FUNC{shmem\_init} must be matched with a call to
+    \FUNC{shmem\_finalize}.
+\\See Section \ref{subsec:shmem_init}.
```
jdinan: Also ref shmem_finalize

naveen-rn:
I'm bit confused with this change...Just to be sure, in this change we are not removing implicit finalize from the spec?

jdinan:
This change removes implicit finalize in cases where the library was initialized with a call to ```shmem_init```.  We don't document this adequately (probably should be in deprecated interfaces), but implicit finalize can still be used with ```start_pes```.

naveen-rn:
At present, we have information about "implicit finalize" at two places - at shmem_init and shmem_finalize explanations. With this change, we seem to remove this information at both these places.  At least, to me it looks like we have removed implicit finalize completely from the spec. 
Can you please verify?

jdinan:
Do you have an objection to the semantic that I mentioned in my previous comment, or are you requesting that we update the deprecated interfaces section to include implicit finalize as part of this proposal?  Or something else?

I don't mind updating the deprecated interfaces documentation, including the ```start_pes``` section before we ratify this proposal.

naveen-rn:
We are removing implicit finalize usage along with shmem_init in this PR. I don't want this semantics change. 

> This change removes implicit finalize in cases where the library was initialized with a call to shmem_init

At present, we require every shmem_init to be matched with shmem_finalize. But, we don't specifically say whether the finalize comes from explicit or implicit call. But this PR tends to make the explicit shmem_finalize usage a required model.

I like the intention behind this PR and this clarification is also important. So let us add new routines: shmem_finalized and shmem_initialized like MPI_initialized and MPI_finalized along with this change and completely remove implicit finalize from the specification.

naveen-rn:
FWIU, as per the current specification the following is the desired and allowed usage model.
1. Use shmem_init along with explicit shmem_finalize and there is no implicit finalize at the end
2. Deprecated, but allow using start_pes() along with implicit finalize. 

As you said, the current spec is inadequate with the following unanswered question:
1. Is using shmem_init() with implicit finalize a valid usecase? (May not be a desired model but is it valid?)
2. Is implicit finalize supposed to work only with start_pes() and a shmem_init() should *always* be used with shmem_finalize()? 

This PR tries to clarify the above question, by removing details about implicit finalize from all places w.r.t to shmem_init. To enhance this clarification, as you suggested we should update the deprecated interfaces documentation, by including the start_pes section and also explicitly telling implicit finalize is valid only with start_pes.

naveen-rn:
Assuming no one comes with the request for implicit finalize (I take back my previous request, I don't have any usage for implicit finalize now ;) ), we could proceed to adding the reference-count semantics (either first-call or nested-call) to further increase the usability for long-term.
1. As we are dealing with almost similar changes, we could either add this new semantics as part of this PR  or
2. We could open two different tickets.

jdinan:
Reference counting requires the library to support re-initialization, because of applications that may use more than one SHMEM library:
```
FOO_init(); /* e.g. I/O Library */
FOO_finalize();
BAR_init(); /* e.g. Compute Library */
BAR_finalize();
```
I suggest that we split these up -- use this ticket to clarify the current semantics as you described above and start a second ticket for reference counting.  Would you be ok with that approach?

jdinan:
@naveen-rn Could you please review the updated draft?
PDF version: [main_spec.pdf](https://github.com/openshmem-org/specification/files/910977/main_spec.pdf)

naveen-rn:
Looks fine to me.

content/execution_model.tex
```diff
@@ -6,14 +6,10 @@
...
-program finishes execution by returning from the main routine or when any PE
-calls \FUNC{shmem\_global\_exit}. When returning from main, \openshmem must
-complete all pending communication and release all the resources associated to
-the library using an implicit collective synchronization across PEs. The user
-has the option to call \FUNC{shmem\_finalize} (before returning from main) to
-complete all pending communication and release all the \openshmem library
-resources without terminating the program. Calling any \openshmem routine after
-\FUNC{shmem\_finalize} leads to undefined behavior.
+program concludes its use of the \openshmem library when all \acp{PE} call
+\FUNC{shmem\_finalize} or when any \ac{PE} calls \FUNC{shmem\_global\_exit}.
```
jdinan: From @shamisp - Do we want to preserve the bit about completing all pending communication and releasing all resources in finalize?

jdinan:
Special ballot changes: 53843d4

content/execution_model.tex
jdinan: Done.

content/execution_model.tex
```diff
@@ -6,14 +6,13 @@
 start by calling the initialization routine  \FUNC{shmem\_init}
 \footnote{\textbf{start\_pes} has been deprecated as of Specification 1.2}
 before using any of the other \openshmem library routines.  An \openshmem
-program finishes execution by returning from the main routine or when any PE
-calls \FUNC{shmem\_global\_exit}. When returning from main, \openshmem must
+program concludes its use of the \openshmem library when all \acp{PE} call
+\FUNC{shmem\_finalize} or when any \ac{PE} calls \FUNC{shmem\_global\_exit}.
```
jdinan: Split into two sentences:  "In addition, the program can be aborted when any PE calls ```shmem_global_exit.```"

content/execution_model.tex
```diff
@@ -6,14 +6,13 @@
...
+During finalization, \openshmem must
```
BryantLam: Suggest "finalization" -> "shmem_finalize"

jdinan: Fixed in c1dfa16, please take a look!

jdinan:
For next time, special ballot readings: 53843d4..c1dfa16

PR #38, opened 2017-01-23
Specify entry/exit for barrier in memory management routines

davidozog:
Clarify that the memory management routines have different requirements for calling barrier: shmem_malloc should call on exit, shmem_free on entry; and shmem_realloc on both entry and exit.

### DIFF
--- a/content/shmem_malloc.tex
+++ b/content/shmem_malloc.tex
@@ -47,7 +47,11 @@ void *shmem_align(size_t alignment, size_t size);
     are provided  so that multiple \ac{PE}s in a program can allocate symmetric,
     remotely accessible memory blocks.  These memory blocks can then be used with
     \openshmem communication routines.  Each of these routines call the
-    \FUNC{shmem\_barrier\_all} routine before returning; this ensures that all
+    \FUNC{shmem\_barrier\_all} routine before returning: \FUNC{shmem\_malloc} calls a
+    barrier on exit; \FUNC{shmem\_free} calls a barrier on entry; and
+    \FUNC{shmem\_realloc} may call barriers on both entry and exit, depending on
+    whether an existing allocation is modified and whether new memory is allocated.
+    This ensures that all
     \ac{PE}s participate in the memory allocation, and that the memory on other
     \ac{PE}s can be used as  soon as the local \ac{PE} returns.  The user is
     responsible for calling these routines with identical argument(s) on all


content/shmem_malloc.tex
```diff
@@ -47,7 +47,11 @@
     are provided  so that multiple \ac{PE}s in a program can allocate symmetric,
     remotely accessible memory blocks.  These memory blocks can then be used with
     \openshmem communication routines.  Each of these routines call the
-    \FUNC{shmem\_barrier\_all} routine before returning; this ensures that all
+    \FUNC{shmem\_barrier\_all} routine before returning: \FUNC{shmem\_malloc} calls a
```
nspark: What do you think of replacing:
> Each of these routines call the `shmem_barrier_all` routine before returning: [...]

with:

> Each of these routines include at least one call to `shmem_barrier_all`: [...]

The former, specifically the statement "before returning", could be misinterpreted as "_immediately_ before returning", even though you later clarify when the barrier(s) happen. I think this minor change would remove any potential ambiguity or confusion.

nspark:
:+1: I think this is a good clarification to make.

content/shmem_malloc.tex
davidozog: Yes, I agree - thanks.

jdinan:
An interesting (or at least, I thought amusing) side discussion I was just having with @davidozog -- How many barriers are there in the realloc call below:
```
buf = shmem_malloc(sizeof(long));
buf = shmem_realloc(buf, sizeof(long));
```
Because there's no change in the size, realloc could be a no-op.  I think the latest change clarifies this in the right way -- there is always at least one barrier.

jeffhammond:
I don't know that we should prescribe when the barrier occurs in the implementation.  Rather, we should define the semantics  that are user-visible.  This will lead to the same result for `shmem_malloc` and `shmem_free` but might make for a more sensible statement about `shmem_realloc`.

I'm thinking something along the lines of:
* "Symmetric memory will be visible to all processes when `shmem_malloc` returns"
* "`shmem_free` will not free memory while remote accesses may be in flight" (this one clearly needs improvement)

jdinan:
I tend to agree, but the trouble is that this semantic has been around for a while and users actually rely on these routines performing an implicit call to ```shmem_barrier_all```.

tonycurtis:
Agree with Jeff.  The spec. doesn't want to tell people where to insert code.  "shmem_malloc() will guarantee visibility of memory" might involve shmem_barrier_all() or it might not.  It's the guarantee of visibility that programmers will bank on even if they accidentally conflate it with an implicit call to barrier_all().

jeffhammond:
@jdinan The definitions I gave would require those implicit barriers, they just do so implicitly, rather than trying to prescribe an implementation.

swpoole:
+1

jdinan:
@jeffhammond The semantic you proposed for ```shmem_free``` is weaker than the existing barrier semantic.  For example, with your semantic, I can locally mark an allocation as "freed" without performing a barrier during the ```shmem_free``` operation.  I can later garbage collect the freed memory when I determine that all PEs have finished their accesses (e.g. at the next barrier or allocation routine).

I do _prefer_ the weaker semantic, but I'm concerned that weakening it will allow for implementations that are not compatible with existing code.  The implementation I mentioned above will definitely cause breakages in the Cray tests, which (as we recently found in SOS testing) expect a barrier in ```shmem_free```.

davidozog:
I do think some prose exists that might capture the semantics we desire, while not enforcing explicit calls to `shmem_barrier_all` (and only `shmem_barrier_all`) in the implementation - but it's not as simple to write.  Maybe something like, "shmem_free will not free memory until there is a guarantee no PE will access any region of that freed block."  I think the Cray tests in question would pass if that were accomplished - it just seems `shmem_barrier_all` at entry is the simplest and clearest way to achieve that.

jdinan:
To rule out lazy deallocation, you also need to add that ```shmem_free``` will not _return_ until the memory is freed.  That gets you the PE synchronization part of the current barrier semantic, but I can still implement this routine without quieting all traffic at every PE (e.g. I can quiet traffic to the symmetric heap, but not to the data segment).  Thus, it's still weaker than the current ```shmem_barrier_all``` requirement.

davidozog:
Does this draft get close?  Please feel free to make changes.

> ... Each of these routines require PE synchronization at entry and/or exit to ensure correct behavior. The shmem_malloc routine must not return until the new memory block is available to all PEs. The shmem_free routine must not free memory until there is a guarantee no PE will access any region of the freed block, and it must not return until the block is freed. Similarly, the shmem_realloc routine must not change memory until there is a guarantee no PE will access the old block, and it must also not return until the new block is available to all PEs...

jeffhammond:
Does `shmem_free(NULL)` have barrier semantics?

jdinan:
@jeffhammond That's another good example.  Yes, I think there is still a barrier in this case.  I looked back through past versions of the OpenSHMEM spec and they all contain this semantic.  The Cray manpage also says something similar:
> Each of these functions call the shmem_barrier_all function before returning; this ensures that all PEs participate in the memory allocation, and that the memory on other PEs can be used as soon as the local PE returns

naveen-rn:
> Does shmem_free(NULL) have barrier semantics?
I can confirm that there is a barrier even in this NULL ptr case in Cray SHMEM and also any change to weaken the shmem_free semantics would create backward incompatibility issues.

nspark:
> Does `shmem_free(NULL)` have barrier semantics?

Based on @davidozog's [revised draft text](https://github.com/openshmem-org/specification/pull/38#issuecomment-275438008), no, I do not think a barrier is _required_ for this case.

But, I do agree that the current specification includes a barrier for this case and removing the requirement would be a backwards-incompatible change. Still, I _suspect_ that such a change would not significantly affect users.

jeffhammond:
`shmem_free` is less important than `shmem_malloc`.  At some point in the past, Cray `shmalloc` didn't barrier (which is good except when unexpected) but it seems it does now.

nspark:
Is it also worth clarifying the behavior of the pointer returned by `shmem_malloc(0)`?

POSIX expands on the C Standard definition of `malloc()` [to say](http://pubs.opengroup.org/onlinepubs/9699919799/functions/malloc.html):
>If the size of the space requested is 0, the behavior is implementation-defined: either a null pointer shall be returned, or the behavior shall be as if the size were some non-zero value, except that the behavior is undefined if the returned pointer is used to access an object.

It has always bugged me that `malloc(0)` _could_ return a non-null pointer, which would pass a test for non-null-ness, but could not be dereferenced. To the extent I've thought about it, this means you also have to carry logic around to check your allocation size if it could possibly be 0.

I'm inclined to suggest that we specify that `shmem_malloc(0)` return a null pointer, which then makes me ask whether such a case should also include a barrier. This, then, led me to a few questions:

(I'm sorry, but my thoughts are snowballing...)

* In general, should `shmem_malloc()` include a barrier semantic when the return value is a null pointer? (e.g., `shmem_malloc(0)` or `shmem_malloc(SIZE_MAX)`)
* Can `shmem_malloc()` return a null pointer on a proper subset of all the PEs?
* Do we require the symmetric heap to be the same size on all PEs?

jeffhammond:
`malloc(SIZE_MAX)` may not only succeed but also return a useful pointer.  The OOM error might only happen when one faults the pages in excess of physical memory, if that.  Virtual memory isn't real, after all 😃 

Thus, I'm not sure that your implication about `shmem_malloc(SIZE_MAX)` returning `NULL` is correct.

jeffhammond:
We require the arguments to `shmem_malloc` to be identical across all PEs.  There is no need to specify how the symmetric heap is implemented, and certainly one does not have to pre-allocate it, so I'm not sure I understand your final question.

nspark:
> [...] Thus, I'm not sure that your implication about `shmem_malloc(SIZE_MAX)` returning `NULL` is correct.

Fair point. I know that Linux can over-commit memory returned by `malloc()`. It was intended as an example of an allocation that would fail, though it clearly had its flaws.

> We require the arguments to `shmem_malloc` to be identical across all PEs.  There is no need to specify how the symmetric heap is implemented, and certainly one does not have to pre-allocate it, so I'm not sure I understand your final question.

The specification provides the `SHMEM_SYMMETRIC_SIZE` environment variable to specify "the number of bytes to allocate for the symmetric heap." As a contrived example, some PEs could call `setenv()` and specify this env-var to a different value than other PEs. (The general use of `setenv()` to specify the heap size seems reasonable; for example, the heap size may be a function of a command-line argument for a problem size.)

After setting `SHMEM_SYMMETRIC_SIZE` to different values for some PEs, they could then call `shmem_init()` and should have symmetric heaps of different sizes. Later, after some number of calls to `shmem_malloc()`, the PEs with a smaller heap may fail to allocate symmetric memory despite specifying identical arguments to `shmem_malloc()`.

In this example, `shmem_malloc()` may return a non-null pointer on some PEs and a null pointer on others.

jeffhammond:
Setting `SHMEM_SYMMETRIC_SIZE` differently across PEs should be prohibited by the standard.

In any case, while we do not guarentee symmetric addresses in OpenSHMEM, we should guarentee that the validity of pointers returned from `shmem_malloc` is symmetric.

tonycurtis:
> On Jan 27, 2017, at 8:37 PM, Jeff Hammond <notifications@github.com> wrote:
> 
> Setting SHMEM_SYMMETRIC_SIZE differently across PEs should be prohibited by the standard.
> 
> In any case, while we do not guarentee symmetric addresses in OpenSHMEM, we should guarentee that the validity of pointers returned from shmem_malloc is symmetric.
> 

The user request might have to be the same size, but we accept that the implementation could allocate different size heaps?
E.g for alignment or resource reasons…

tony

jeffhammond:
If the implementation allocates more than requested, that is not user-visible since attempting to allocate more than `SHMEM_SYMMETRIC_SIZE` seems like undefined behavior.

manjugv:
If there is no need specify the barrier (besides backward compatibility reasons), shouldn’t we just mark it for removal in 2.0 ? 

If we decide to go with this ticket, IMO, we should add this as clarification rather than the part of API description.

jdinan:
There is a need for clarification, because the synchronization semantics for these routines are underspecified.  For example, shmem_free is specified to contain a barrier, but when that barrier happens is not specified.  The current text allows an implementation of shmem_free to perform the barrier _after_ freeing the object, which is not what was intended.

Routines without the barrier would be useful for instances where memory management is performance critical.  We could add new versions of these routines that are unsynchronized.  I think I would prefer this to a semantic deprecation, since we can do it right away, there is a lower likelihood of user errors, and it's easier to support.

content/shmem_malloc.tex
```diff
@@ -46,8 +46,12 @@
     The \FUNC{shmem\_malloc}, \FUNC{shmem\_free}, and \FUNC{shmem\_realloc} routines
```
nspark: Add `shmem_align()`

content/shmem_malloc.tex
```diff
@@ -46,8 +46,12 @@
...
     are provided  so that multiple \ac{PE}s in a program can allocate symmetric,
     remotely accessible memory blocks.  These memory blocks can then be used with
-    \openshmem communication routines.  Each of these routines call the
-    \FUNC{shmem\_barrier\_all} routine before returning; this ensures that all
+    \openshmem communication routines.  Each of these routines include at least one
```
nspark: Add `shmem_align()` here too

PR #40, opened 2017-02-03
Add missing SHMEM_GET64 and SHMEM_GET64_NBI routines

jdinan:
These are part of the Fortran interface, but were not listed in the
spec.


BryantLam:
Remarkable. Looks like these functions were accidentally omitted during the document restructuring from v1.0 to v1.1.

PR #41, opened 2017-02-06
Addition of Symmetric Distributed Locking Routines

jamesaross:
This is related to Redmine extension proposal [#232](http://www.openshmem.org/redmine/issues/232)

**Description**

OpenSHMEM 1.3 Section 8.9 provides Distributed Locking Routines whereby a scalar symmetric data object represents a single global lock that may be acquired/released by any PE. Thus, a scalar symmetric data object is treated semantically as a single object even though the applicable declarations associated with the lock instantiate one scalar per PE. Furthermore, in order to provide fine-grain protection of resources local to a PE, it is necessary to allocate an array of locks as a symmetric data object, and then explicitly associate an element in the lock array with a given PE. This is also semantically inconsistent since a symmetric data object is treated as a single global object by the current Distributed Locking Routines. A consequence is that, with the current specification, the total allocated size allocated for fine-grain lock will scale as O(N^2) whereas logical locks scale as O(N), where `N = shmem_n_pes()`.

The proposed Symmetric Distributed Locking Routines appear similar to the existing distributed locking routines with the addition of a second PE argument so that symmetric scalar locks may be treated consistently with ordinary scalar symmetric data objects and used fine-grain locking of a resources local to a specified PE.

**Benefits as Result of Changes**

Fine-grain inter-process locking reduces the memory footprint and potentially network traffic required for locks, leading to more efficient implementations.

The total memory used to implement fine-grain locking scales with the square of the number of processing elements, prohibiting the operation for some highly scalable architecture with limited memory.

With 1M PEs in the code below, for example, 8 Terabytes of memory in total is allocated. The proposed extension would reduce that to a total of just 8 MB.

**Typical code for existing locking routines**

```c
int npes = shmem_n_pes();
int my_pe = shmem_my_pe();
int nxt_pe = (my_pe + 1) % npes;

// x and lock are declared identically
static long x = my_pe;
static long lock = 0;

// x and lock follow different semantic as data objects
long new_value = my_pe + 100;
shmem_set_lock(&lock);
shmem_p(&x, new_value, nxt_pe);
shmem_clear_lock(&lock);
```

**Typical code for existing fine-grain locking routines**

```c
int npes = shmem_n_pes();
int my_pe = shmem_my_pe();
int nxt_pe = (my_pe + 1) % npes;

// array of locks required for fine-grain locking
static long x = my_pe;
long* locks = (long*)shmem_malloc( npes * sizeof(*locks) );
memset(locks, 0, npes * sizeof(*locks));

long new_value = my_pe + 100;
shmem_set_lock(&locks[nxt_pe]);
shmem_p(&x, new_value, nxt_pe);
shmem_clear_lock(&locks[nxt_pe]);
```

**Proposed extension solution**

```c
int npes = shmem_n_pes();
int my_pe = shmem_my_pe();
int nxt_pe = (my_pe + 1) % npes;

// x and lock are declared identically
static long x = my_pe;
static long lock = 0;

// x and lock follow identical semantic as data objects
long new_value = my_pe + 100;
shmem_set_lock_pe(&lock, nxt_pe);
shmem_p(&x, new_vlaue, nxt_pe);
shmem_clear_lock_pe(&lock, nxt_pe);
```

**SW Impact Assessment**

Three new routines must be written. They are straightforward extensions to the three existing Distributed Locking Routines whereby a second argument is added to specify the PE for the target lock.

**HW Impact Assessment**

This proposal should reduce the asymmetric network congestion for OpenSHMEM implementations that locate the actual locks on a particular PE (say PE 0), and reduce the scaling of storage for fine-grain locks from O(N^2) to O(N).

**Synopsis**

_C/C++:_
```c
void shmem_clear_lock_pe(long *lock, int pe);
void shmem_set_lock_pe(long *lock, int pe);
int shmem_test_lock_pe(long *lock, int pe);
```
**Parameters**

_Arguments:_
_IN lock_: A symmetric data object that is a scalar variable or an array of length 1. This data object must be set to 0 on all PEs prior to the first use. lock must be of type long.

_IN pe_: An integer that indicates the PE number upon which lock is to be updated.

**Return Values:**

The shmem_test_lock_pe routine returns 0 if the lock was originally cleared and this call was able to set the lock. A value of 1 is returned if the lock had been set and the call returned without waiting to set the lock.

### DIFF
--- a/content/shmem_lock_pe.tex
+++ b/content/shmem_lock_pe.tex
@@ -28,8 +28,8 @@ I = SHMEM_TEST_LOCK_PE(lock, pe)
 \apidescription{
     The \FUNC{shmem\_set\_lock\_pe} routine sets a mutual exclusion lock on a
     specific \ac{PE} after waiting for the lock to be freed by any other \ac{PE}
-    currently holding the lock.  Waiting \ac{PE}s are assured of getting the lock in
-    a first-come, first-served manner. The \FUNC{shmem\_clear\_lock\_pe} routine
+    currently holding the lock.  Waiting \ac{PE}s spin-wait on the lock resource
+    until it is released.  The \FUNC{shmem\_clear\_lock\_pe} routine
     releases a lock on a specific \ac{PE} previously set by
     \FUNC{shmem\_set\_lock\_pe} after ensuring that all local and remote stores
     initiated in the critical region are complete.  The \FUNC{shmem\_test\_lock\_pe}
--- a/content/shmem_lock_pe.tex
+++ b/content/shmem_lock_pe.tex
@@ -1,0 +1,69 @@
+\apisummary{
+    Releases, locks, and tests a mutual exclusion memory lock on a specific \ac{PE}.
+}
+\begin{apidefinition}
+
+\begin{Csynopsis}
+void shmem_clear_lock_pe(volatile long *lock, int pe);
+void shmem_set_lock_pe(volatile long *lock, int pe);
+int shmem_test_lock_pe(volatile long *lock, int pe);
+\end{Csynopsis}
+
+\begin{Fsynopsis}
+INTEGER pe, lock, SHMEM_TEST_LOCK
+CALL SHMEM_CLEAR_LOCK_PE(lock, pe)
+CALL SHMEM_SET_LOCK_PE(lock, pe)
+I = SHMEM_TEST_LOCK_PE(lock, pe)
+\end{Fsynopsis}
+
+\begin{apiarguments}
+\apiargument{IN}{lock}{A symmetric data object that is a scalar variable or an array
+    of  length \CONST{1}.  This data  object  must  be set to \CONST{0} on all
+    \ac{PE}s prior to the first use.  \VAR{lock}  must  be  of type \CONST{long}.
+    If you are using \Fortran, it must be of default kind.}
+\apiargument{IN}{pe}{An integer that indicates the \ac{PE} number upon which lock is
+    to be updated. If you are using Fortran, it must be a default integer value.}
+\end{apiarguments}
+
+\apidescription{
+    The \FUNC{shmem\_set\_lock\_pe} routine sets a mutual exclusion lock on a
+    specific \ac{PE} after waiting for the lock to be freed by any other \ac{PE}
+    currently holding the lock.  Waiting \ac{PE}s are assured of getting the lock in
+    a first-come, first-served manner. The \FUNC{shmem\_clear\_lock\_pe} routine
+    releases a lock on a specific \ac{PE} previously set by
+    \FUNC{shmem\_set\_lock\_pe} after ensuring that all local and remote stores
+    initiated in the critical region are complete.  The \FUNC{shmem\_test\_lock\_pe}
+    routine sets a mutual exclusion lock on a specific \ac{PE} only if it is
+    currently cleared.  By using this routine, a \ac{PE} can avoid blocking on a set
+    lock. If the lock is currently set, the routine returns without waiting. These
+    routines are appropriate for protecting a critical region from simultaneous
+    update by multiple \ac{PE}s at a fine-grain level.
+}
+
+\apireturnvalues{
+    The \FUNC{shmem\_test\_lock} routine returns \CONST{0} if  the lock  was
+    originally cleared and  this  call was  able  to set the lock.  A value of
+    \CONST{1} is returned if the lock had been set and the call returned without
+    waiting to set the lock.
+}
+
+\apinotes{
+    The term symmetric data object is defined in Section \ref{subsec:memory_model}.
+    The lock variable should always be initialized to zero and accessed only by the
+    \openshmem locking \ac{API}.  Changing the value of the lock variable by other
+    means without using the \openshmem \ac{API}, can lead to undefined behavior.
+    Mixing the \FUNC{shmem\_lock} and \FUNC{shmem\_lock\_pe} routines with the same
+    lock can lead to undefined behavior.
+}
+
+\begin{apiexamples}
+
+\apicexample
+    {The following example uses \FUNC{shmem\_set\_lock\_pe} and
+    \FUNC{shmem\_clear\_lock\_pe} in a \Clang{} program.}
+    {./example_code/shmem_lock_pe_example.c}
+    {}
+
+\end{apiexamples}
+
+\end{apidefinition}
--- a/example_code/shmem_lock_pe_example.c
+++ b/example_code/shmem_lock_pe_example.c
@@ -1,0 +1,21 @@
+#include <stdio.h>
+#include <shmem.h>
+
+int main(void)
+{
+   int old = -1;
+   static int dst = 22;
+   static long lock = 0;
+   shmem_init();
+   int me = shmem_my_pe();
+   if (me == 1) {
+      shmem_set_lock_pe(&lock, 0);
+      old = shmem_g(&dst, 0);
+      shmem_p(&dst, old + 44, 0);
+      shmem_clear_lock_pe(&lock, 0);
+   }
+   shmem_barrier_all();
+   printf("%d: old = %d, dst = %d\n", me, old, dst);
+   shmem_finalize();
+   return 0;
+}
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -465,6 +465,10 @@ synchronization routines.
 \item Clarified deprecation overview and added deprecation rationale in Annex F.
 \\See Section \ref{sec:dep_api}.
 %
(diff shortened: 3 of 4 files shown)


nspark:
@jamesaross says:
> The semantic of a single global lock may be supported as a special case, and these routines therefore have the potential of deprecating the existing Distributed Locking Routines, but also do not present a conflict with the existing routines.

OpenSHMEM's current lock routines have a FIFO queuing behavior. (This was a surprise to me that came up in some discussion of the lock routines.) This then made the O(_npes_) storage requirement make more sense. I agree that the existing routines aren't in conflict with your proposal, but it seems to me like the single global lock case is not trivially supported by the extensions you propose here and wouldn't directly lead to deprecation of the existing lock routines.

Practically speaking, these extensions seem like little more than a thin wrapper around `shmem_cswap`. Is that a fair characterization?

jamesaross:
@nspark said:
> ...it seems to me like the single global lock case is not trivially supported by the extensions you propose here and wouldn't directly lead to deprecation of the existing lock routines.

I've updated the documentation to remove the suggestion that they could deprecate the existing routines.  They should remain.  The FIFO requirement with the existing routines is an important distinction between these proposed routines and I've updated the documentation.

The locking mechanisms are clearly lacking from the OpenSHMEM standard and there are probably a number of ways to go about defining them.  Pthreads solved the issue by introducing mutex attributes (and associated routines) to enable the programmer to define how locking should behave.  The Pthreads locks and attributes also have opaque types, but OpenSHMEM seems to prefer to use `long` as an opaque type in the implementations of locking routines.

@nspark said:
> Practically speaking, these extensions seem like little more than a thin wrapper around shmem_cswap. Is that a fair characterization?

You could possibly say that, but I would describe `shmem_cswap` as thin wrapper around `shmem_set_lock_pe` and `shmem_clear_lock_pe` with some intermediate compare and swap operation.  The proposed routines enable arbitrary atomic operations, including `shmem_cswap`.

To your point, here's how you might implement `shmem_set_lock_pe` with `shmem_cswap`:
```
void shmem_set_lock_pe(long* lock, int pe) {
   while(shmem_cswap(lock, 0, 1, pe)); // yes, this looks pretty simple.
}
```

And here's how you might implement `shmem_long_cswap` with `shmem_set_lock_pe` and `shmem_clear_lock_pe`:
```
void shmem_long_cswap(long* dest, long cond, long value, int pe) {
   static long lock = 0;
   shmem_set_lock_pe(&lock, pe);
   long r = shmem_g(dest, pe);
   if (r == cond) shmem_p(dest, value, pe);
   shmem_clear_lock_pe(&lock, pe);
   return r;
}
```

So which should we call the thin wrapper?  They are certainly related, but I would argue `shmem_cswap` is the specialized case and the proposed extensions are the lower-level general-purpose locking routines required in order to implement arbitrary sequences of atomic operations.  As another example, here's how you might implement an inline atomic multiply-add on a remote PE:
```
static long lock = 0;
static float foo = 1.23f;
float mul = 10.0f;
float add = 5.0f;
shmem_set_lock_pe(&lock, pe);
float tmp = shmem_g(&foo, pe);
tmp = mul*tmp + add;
shmem_p(&foo, tmp, pe);
shmem_clear_lock_pe(&lock, pe);
```

content/shmem_lock_pe.tex
```diff
@@ -0,0 +1,69 @@
+\apisummary{
+    Releases, locks, and tests a mutual exclusion memory lock on a specific \ac{PE}.
+}
+\begin{apidefinition}
+
+\begin{Csynopsis}
+void shmem_clear_lock_pe(volatile long *lock, int pe);
+void shmem_set_lock_pe(volatile long *lock, int pe);
+int shmem_test_lock_pe(volatile long *lock, int pe);
```
jdinan: Ergh, no ```volatile``` please.  Ref #6.  :)

content/shmem_lock_pe.tex
```diff
@@ -0,0 +1,69 @@
...
+CALL SHMEM_SET_LOCK_PE(lock, pe)
+I = SHMEM_TEST_LOCK_PE(lock, pe)
+\end{Fsynopsis}
+
+\begin{apiarguments}
+\apiargument{IN}{lock}{A symmetric data object that is a scalar variable or an array
+    of  length \CONST{1}.  This data  object  must  be set to \CONST{0} on all
+    \ac{PE}s prior to the first use.  \VAR{lock}  must  be  of type \CONST{long}.
+    If you are using \Fortran, it must be of default kind.}
+\apiargument{IN}{pe}{An integer that indicates the \ac{PE} number upon which lock is
```
jdinan: ```lock``` -> ```\VAR{lock}```

content/shmem_lock_pe.tex
```diff
@@ -0,0 +1,69 @@
...
+void shmem_set_lock_pe(volatile long *lock, int pe);
+int shmem_test_lock_pe(volatile long *lock, int pe);
+\end{Csynopsis}
+
+\begin{Fsynopsis}
+INTEGER pe, lock, SHMEM_TEST_LOCK
+CALL SHMEM_CLEAR_LOCK_PE(lock, pe)
+CALL SHMEM_SET_LOCK_PE(lock, pe)
+I = SHMEM_TEST_LOCK_PE(lock, pe)
+\end{Fsynopsis}
```
jdinan: Prefer not to add Fortran bindings, since they are being deprecated.

content/shmem_lock_pe.tex
```diff
@@ -0,0 +1,69 @@
+\apisummary{
+    Releases, locks, and tests a mutual exclusion memory lock on a specific \ac{PE}.
```
jdinan: Why "memory" lock?

content/shmem_lock_pe.tex
```diff
@@ -0,0 +1,69 @@
...
+    of  length \CONST{1}.  This data  object  must  be set to \CONST{0} on all
+    \ac{PE}s prior to the first use.  \VAR{lock}  must  be  of type \CONST{long}.
+    If you are using \Fortran, it must be of default kind.}
+\apiargument{IN}{pe}{An integer that indicates the \ac{PE} number upon which lock is
+    to be updated. If you are using Fortran, it must be a default integer value.}
+\end{apiarguments}
+
+\apidescription{
+    The \FUNC{shmem\_set\_lock\_pe} routine sets a mutual exclusion lock on a
+    specific \ac{PE} after waiting for the lock to be freed by any other \ac{PE}
```
jdinan: freed -> cleared

jdinan:
What sort of fairness behavior do you want for this lock?  If it's implemented using as a simple spin-lock, it can be very unfair since the PE that hosts the lock will be able to issue lock requests faster than remote PEs.  This effectively prioritizes local PEs and can increase the likelihood of starvation for a producer-consumer pattern with a local consumer and remote producer.  You could also implement this API with the bakery algorithm, which is fair, but does not support a ```test``` operation.

content/shmem_lock_pe.tex
jamesaross: data object must be set to 0 on remote destination only

content/shmem_lock_pe.tex
jamesaross: cleared by calling shmem_clear_lock_pe

content/shmem_lock_pe.tex
nspark: Request: Revisit wording here to be less Fortran-y.

Suggest: "lock: A pointer to a symmetric data object of type \CTYPE{long}. This data object must be initialized to \CONST{0} on the target PE prior to first use."

(See #55 for `\CTYPE`.)

example_code/shmem_lock_pe_example.c
```diff
@@ -0,0 +1,21 @@
...
+#include <shmem.h>
+
+int main(void)
+{
+   int old = -1;
+   static int dst = 22;
+   static long lock = 0;
+   shmem_init();
+   int me = shmem_my_pe();
+   if (me == 1) {
```
jamesaross: consider removing this line and having a better example of the behavior

jdinan:
Feedback from a colleague here at Intel:

Different applications use locks in different ways, and lock performance varies with both the lock algorithm/implementation and how the lock is used.  For example, spin locks often have low overhead when uncontended, but queue locks often have low overhead when contended.  That makes it desirable to be able to choose a lock implementation that best matches the application.  In turn, it seems useful to generalize the proposal to

1. Use an opaque lock type, rather than a long.
2. Take arguments that allow the user to choose different lock algorithms.

jamesaross:
@jdinan,

I agree with your comments and my colleague and I have been discussing more Pthreads-style mutexes.  Rather than muddle with the spec now, I'm willing to remove this pull request to address locking in a more robust manner at a later time, including demonstrations on more than one platform.  The discussion at the F2F and the input here have been worthwhile. Thank you.

PR #42, opened 2017-02-07
Fixing Source Buffer Ambiguity in Collective Operations Text (RM #217)

manjugv:
The text describing Broadcast (and other collectives) operation in the current specification 
does not clarify the state of source buffer on root PE (read Page 51, 4-5), after it returns.
One would assume the source buffer is reusable.

### DIFF
--- a/content/shmem_broadcast.tex
+++ b/content/shmem_broadcast.tex
@@ -72,8 +72,8 @@ CALL SHMEM_BROADCAST64(dest, source, nelems, PE_root, PE_start, logPE_stride, PE
     needed to ensure this): The \VAR{pSync} array on all \ac{PE}s in the
     \activeset{} is not still in use from a prior call to a broadcast routine.  The
     \dest{} array on all \ac{PE}s in the \activeset{} is ready to accept the
-    broadcast data. 
-       
+    broadcast data.
+    
     Upon return from a broadcast routine, the following are true for the local
     \ac{PE}: If the current \ac{PE} is not the root \ac{PE}, the \dest{} data object
     is updated. The \source{} data object may be safely reused.  
--- a/content/shmem_broadcast.tex
+++ b/content/shmem_broadcast.tex
@@ -72,14 +72,11 @@ CALL SHMEM_BROADCAST64(dest, source, nelems, PE_root, PE_start, logPE_stride, PE
     needed to ensure this): The \VAR{pSync} array on all \ac{PE}s in the
     \activeset{} is not still in use from a prior call to a broadcast routine.  The
     \dest{} array on all \ac{PE}s in the \activeset{} is ready to accept the
-    broadcast data. \newtext{The data in the \source{} array on the root \ac{PE} is ready to be broadcasted.}
-    
-    \newtext{The \source{} and \dest{} arrays may be the same array, but they may not be overlapping arrays. If they 
-    are different, the source array on the non-root \acp{PE} and \dest{} array on the root \ac{PE} is not used by the broadcast routine.}
-    
+    broadcast data. 
+       
     Upon return from a broadcast routine, the following are true for the local
     \ac{PE}: If the current \ac{PE} is not the root \ac{PE}, the \dest{} data object
-    is updated. \newtext{If the current PE is the root PE, the \source{} data object may be safely reused.}  
+    is updated. \newtext{The \source{} data object may be safely reused.}  
     The values in the \VAR{pSync} array are restored to the original values.
 }
 
--- a/content/shmem_broadcast.tex
+++ b/content/shmem_broadcast.tex
@@ -76,7 +76,7 @@ CALL SHMEM_BROADCAST64(dest, source, nelems, PE_root, PE_start, logPE_stride, PE
        
     Upon return from a broadcast routine, the following are true for the local
     \ac{PE}: If the current \ac{PE} is not the root \ac{PE}, the \dest{} data object
-    is updated. \newtext{The \source{} data object may be safely reused.}  
+    is updated. The \source{} data object may be safely reused.  
     The values in the \VAR{pSync} array are restored to the original values.
 }
 
--- a/content/shmem_broadcast.tex
+++ b/content/shmem_broadcast.tex
@@ -72,11 +72,14 @@ CALL SHMEM_BROADCAST64(dest, source, nelems, PE_root, PE_start, logPE_stride, PE
     needed to ensure this): The \VAR{pSync} array on all \ac{PE}s in the
     \activeset{} is not still in use from a prior call to a broadcast routine.  The
     \dest{} array on all \ac{PE}s in the \activeset{} is ready to accept the
-    broadcast data.
+    broadcast data. \newtext{The data in the \source{} array on the root \ac{PE} is ready to be broadcasted.}
+    
+    \newtext{The \source{} and \dest{} arrays may be the same array, but they may not be overlapping arrays. If they 
+    are different, the source array on the non-root \acp{PE} and \dest{} array on the root \ac{PE} is not used by the broadcast routine.}
     
     Upon return from a broadcast routine, the following are true for the local
     \ac{PE}: If the current \ac{PE} is not the root \ac{PE}, the \dest{} data object
-    is updated\newtext{, and if the current PE is the root PE, the buffer pointed to by source may be safely reused.}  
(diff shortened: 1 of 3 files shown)


jeffhammond:
We should aspire to have non-NULL commit descriptions and PR descriptions, respectively. 😄

content/shmem_broadcast.tex
```diff
@@ -72,12 +72,15 @@
     needed to ensure this): The \VAR{pSync} array on all \ac{PE}s in the
     \activeset{} is not still in use from a prior call to a broadcast routine.  The
     \dest{} array on all \ac{PE}s in the \activeset{} is ready to accept the
-    broadcast data.
+    broadcast data. \newtext{The data in the \source{} array on the root \ac{PE} is ready to be broadcasted.}
+    
+    \newtext{The \source{} and \dest{} arrays may be the same array, but they may not be overlapping arrays. If they 
+    are different, the source array on the non-root \acp{PE} and \dest{} array on the root \ac{PE} is not used by the broadcast routine.}
```
naveen-rn: > if they are different, the source array on the non-root \acp{PE} and \dest{} array on the root \ac{PE} is not used by the broadcast routine.

This particular statement would cause performance impact - esp on implementations like Cray SHMEM which uses network assisted multicast operations for performing SHMEM BCAST operations. SHMEM has a unique scenario where we can't BCAST to the root destination array. Most HW BCAST operations can't handle this (let others confirm as well). Hence implementations have some sort of temporary array with the value before the HW BCAST operations and then copy back the data to the destination array just in the root PE. We can't use the pSync for this purpose, because we would need pSync to store the temporary state.    

If we need this restriction specifically, we should try to change/remove the existing statement: "The data is not copied to the dest area on the root PE."

jeffhammond: How is it application-visible if the HW overwrites the buffer at root with the same bits?

naveen-rn: The dest array in the root PE shouldn't be made visible to the application when the broadcast is in progress, because it is being used by the HW operations. But, this specific change guarantees that both the dest array in root PE and source array in the non-root PEs are application-visible.

manjugv: @naveen-rn : Other vendors don't seem to have that restriction.

manjugv: @jeffhammond: I thought that dedicated folks are you are going to read the redmine ticket :) I have updated the PR.

jeffhammond: @manjugv I avoid Redmine unless absolutely necessary 😜

content/shmem_broadcast.tex
```diff
@@ -72,12 +72,15 @@
     needed to ensure this): The \VAR{pSync} array on all \ac{PE}s in the
     \activeset{} is not still in use from a prior call to a broadcast routine.  The
     \dest{} array on all \ac{PE}s in the \activeset{} is ready to accept the
-    broadcast data.
+    broadcast data. \newtext{The data in the \source{} array on the root \ac{PE} is ready to be broadcasted.}
```
jdinan: "Before any PE calls a broadcast routine ... the data in the source array on the root PE is ready to be broadcasted."

I thought that we decided against this semantic?  IIUC, this will effectively require a barrier before every broadcast routine, which is definitely going to break backward compatibility.

manjugv: @jdinan Even without this new change, we had to check the dest buffer for readiness, before you send the data from root to non-root PEs. 

IIRC, the suggestion was to use something like this - ““The shmem_broadcast cannot start unless the root initiates .. “ The issue with this is that the dest buffer still has to checked for readiness. Further, we will essentially be restricting the implementations to not use “Gets”.

jdinan: This unit test illustrates the problem: https://github.com/openshmem-org/tests-cray/blob/master/sma1/shmem_broadcast_all.c

It's often easy for applications to resolve the data dependence involved with the destination buffer.  However, this change would introduce a race into the above test since there is no synchronization between the producer PE setting up the source buffer the consumer PEs calling the barrier.

A further point - this doesn't restrict implementations from using get operations to implement broadcast.  It restricts them from using _unsynchronized_ gets.  Unsynchronized gets would be most useful in an implementation that uses a full fan-out from the root PE, which is not a scalable algorithm.  In any hierarchical broadcast based on gets, you would still need to synchronize the RAW dependence between the interior nodes in the tree regardless of this semantic.  Thus the performance benefit is limited to nodes that are children of the root, which will vanish at large enough scale.

manjugv: > However, this change would introduce a race into the above test since there is no synchronization between the producer PE setting up the source buffer the consumer PEs calling the barrier.

Depending on the implementation, there is **already** a race in this test (line: 118), because (1) there is no explicit checking for the destination buffer to be ready and (2) same destination buffer is used for multiple iterations. The root can overwrite the destination buffers, if root is fast and non-root PEs are slower. So, I don't see, how there is a new race introduced. 


> It's often easy for applications to resolve the data dependence involved with the destination buffer.

Can you please elaborate ?

> In any hierarchical broadcast based on gets, you would still need to synchronize the RAW dependence between the interior nodes in the tree regardless of this semantic. Thus the performance benefit is limited to nodes that are children of the root, which will vanish at large enough scale.

I agree about the benefits at scale. Though this might be valid, why constraint an implementation. 

What bothers me about specifying “The shmem_broadcast cannot start unless the root initiates ..” is that it is a stronger semantic than it is needed for a broadcast, and it prescribes how to implement the broadcast. For broadcast the data on the root's source buffer should be ready, it should not matter whether root initiates or non-root PEs initiate the data transfer.

jdinan: Well, that's an interesting point about the destination buffer synchronization issue caused by the loop.  Took me a minute to see it, but I think you're right.  @naveen-rn -- might want to fix this.  :)

I do appreciate your point that the current text is ambiguous and that the semantic is stronger that what is necessary, but I submit that the de facto semantic based on code in the wild is that the source buffer should not be accessed until the root enters the call.  That is, no matter what we write in the specification, this is the semantic that I need to support as an implementer.

Here's another example from the T3E documentation (just to keep picking on Cray): http://docs.cray.com/books/2178_3.0.1/html-2178_3.0.1/z822949834dep.html

naveen-rn: I'm sorry, I'm not able to understand race here. Can you please elaborate?

manjugv: > Here's another example from the T3E documentation (just to keep picking on Cray): http://docs.cray.com/books/2178_3.0.1/html-2178_3.0.1/z822949834dep.html

@jdinan Even in this example, they already have a barrier_all, and it will be moved to a different place with this clarification. You don't require more than that.

content/shmem_broadcast.tex
```diff
@@ -72,12 +72,15 @@
...
+    
+    \newtext{The \source{} and \dest{} arrays may be the same array, but they may not be overlapping arrays. If they
```
jdinan: Suggested edit: "The \source{} and \dest{} arrays may be either identical or distinct, but they must not be partially overlapping arrays."

content/shmem_broadcast.tex
manjugv: Remove this paragraph (F2F notes)

content/shmem_broadcast.tex
```diff
@@ -72,12 +72,15 @@
     needed to ensure this): The \VAR{pSync} array on all \ac{PE}s in the
     \activeset{} is not still in use from a prior call to a broadcast routine.  The
     \dest{} array on all \ac{PE}s in the \activeset{} is ready to accept the
-    broadcast data.
+    broadcast data. \newtext{The data in the \source{} array on the root \ac{PE} is ready to be broadcasted.}
+
```
manjugv: Remove this sentence (F2F Notes)

content/shmem_broadcast.tex
```diff
@@ -72,12 +72,12 @@
     needed to ensure this): The \VAR{pSync} array on all \ac{PE}s in the
     \activeset{} is not still in use from a prior call to a broadcast routine.  The
     \dest{} array on all \ac{PE}s in the \activeset{} is ready to accept the
-    broadcast data.
-    
+    broadcast data. 
+
```
jdinan: Could you push a commit to remove this whitespace change?  This is a "document edit", so it should not impact your special ballot.

PR #45, opened 2017-02-07
Add dedication to David Knaak in OpenSHMEM 1.4

jdinan:
I would like to propose that we add a dedication to OpenSHMEM 1.4 to honor David.


jdinan:
PDF Draft: 
[main_spec.pdf](https://github.com/openshmem-org/specification/files/758538/main_spec.pdf)

PR #47, opened 2017-02-07
Add shmem_calloc function

nspark:
This PR adds the `shmem_calloc` function.

**Special Ballot**: diff ([30e8392..d31f114](https://github.com/openshmem-org/specification/pull/47/files/30e83921293fc384ae2209bd99f775a5508e572f..d31f11466f364711add00743dc8843c065f07302)) and [PDF](https://github.com/openshmem-org/specification/files/838553/proposal-calloc-20170313.pdf)

Note that `shmem_calloc` was added separately from the `shmem_{malloc,realloc,align,free}` block because the argument descriptions would conflict.

### DIFF
--- a/content/shmem_calloc.tex
+++ b/content/shmem_calloc.tex
@@ -29,8 +29,7 @@ void *shmem_calloc(size_t count, size_t size);
   all \ac{PE}s calling \FUNC{shmem\_calloc}; otherwise, the behavior is
   undefined.
 
-  The \FUNC{shmem\_calloc} routine calls \FUNC{shmem\_barrier\_all} on exit
-  prior to returning.
+  The \FUNC{shmem\_calloc} routine calls \FUNC{shmem\_barrier\_all} on exit.
 }
 
 \apireturnvalues{
--- a/content/shmem_calloc.tex
+++ b/content/shmem_calloc.tex
@@ -20,6 +20,11 @@ void *shmem_calloc(size_t count, size_t size);
   returns a pointer to the lowest byte address of the allocated symmetric
   memory. The space is initialized to all bits zero.
 
+  If the allocation succeeds, the pointer returned shall be suitably
+  aligned so that it may be assigned to a pointer to any type of object.
+  If the allocation does not succeed, or either \VAR{count} or \VAR{size} is
+  \CONST{0}, the return value is a null pointer.
+
   The values for \VAR{count} and \VAR{size} shall each be equal across
   all \ac{PE}s calling \FUNC{shmem\_calloc}; otherwise, the behavior is
   undefined.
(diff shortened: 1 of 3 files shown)


nspark:
[PDF of this PR as of 30e8392](https://github.com/openshmem-org/specification/files/759074/proposal-calloc-20170207.pdf)

jeffhammond:
Is symmetry w.r.t. C memory management the motivation here, or are we hoping that this is a performance optimization relative to `shmem_malloc(); memset()` because of the use of the zero page?  The zero page assumes symmetric allocation doesn't require page faulting or pinning, which is required by a number of RDMA networks today.

nspark:
I think that this is somewhat of a convenience request with some consideration for optimization.

If an implementation can provide better performance by mapping to the zero page and using copy-on-write, then that would be great. If faulting and pinning is required (which I'd typically expect), then it might not make a big difference.

There's also one less barrier to call `shmem_calloc()` vs `shmem_malloc(); memset(); shmem_barrier_all();`.

jeffhammond:
Ah yes, I forgot about the implicit barrier problem 😄

content/shmem_calloc.tex
```diff
@@ -0,0 +1,40 @@
...
+}
+
+\begin{apidefinition}
+
+\begin{Csynopsis}
+void *shmem_calloc(size_t nmemb, size_t size);
+\end{Csynopsis}
+
+\begin{apiarguments}
+  \apiargument{IN}{nmemb}{The number of elements to allocate.}
```
jdinan: Why ```nmemb```?  The other allocation routines call this argument ```size```.

nspark: Because `size` is already used for the per-element size. The C standard uses `nmemb` for `calloc()`; POSIX uses `nelem`.

jdinan: ... and Mac manpages call this argument ```count```.

content/shmem_calloc.tex
```diff
@@ -0,0 +1,40 @@
...
+  memory for an array of \VAR{nmemb} objects of \VAR{size} bytes each and
+  returns a pointer to the lowest byte address of the allocated symmetric
+  memory. The space is initialized to all bits zero.
+
+  The values for \VAR{nmemb} and \VAR{size} shall each be equal across
+  all \ac{PE}s calling \FUNC{shmem\_calloc}; otherwise, the behavior is
+  undefined.
+
+  The \FUNC{shmem\_calloc} routine calls \FUNC{shmem\_barrier\_all} on exit
+  prior to returning.
```
jdinan: Do we need to say both "on exit" and "prior to returning"?

nspark:
[PDF of this PR as of d31f114](https://github.com/openshmem-org/specification/files/838553/proposal-calloc-20170313.pdf)

Issue #49, opened 2017-02-09
Where did <mpp/shmem.h> go?


BryantLam:
The compatibility note in OpenSHMEM 1.0 for header files `mpp/shmem.h` and `mpp/shmem.fh` is not present in 1.1 and newer. I don't mind that it's gone, but this change is not documented in the changelog for 1.1. Are implementations expected to maintain this backwards compatibility?

BryantLam:
PR #67

PR #50, opened 2017-02-09
Clarified header files for standardized OpenSHMEM API

BryantLam:
The only locations where `shmem.h` and `shmem.fh` are spelled out are in example codes and Annex A. Added clarification to Section 5: Language Bindings.


jdinan:
What do you think about adding the ```mpp/*``` location for the header files to the deprecation annex?  There's lots of code out there that expects the headers to be in this subdirectory off the include search path.

swpoole:
Any major negative impacts we know of?

jdinan:
The ```mpp/``` copies of the header are required for backward compatibility in OpenSHMEM 1.0, but they disappear after that.  I think adding them to the deprecation chapter would be a positive change in terms of re-capturing these as a backward compatibility feature that may be removed in the future.

shamisp:
Concerns about backward compatibility ?

BryantLam:
Discussion related to #49.

I don't believe there are any concerns with backwards compatibility.  I'll work on a separate ticket later today to add the `mpp/*` headers into the deprecation table with a still-supported flag. This location can be formally removed at a later date (possibly alongside the official removal of Fortran).

swpoole:
+1

Best,
Steve (ADTSC)

jdinan:
As long as you're fixing this, could you also add a ```\header``` command to ```utils/defs.tex``` and update the formatting of the other references to ```shmem.h``` and ```shmem.fh``` in Annex A, 8.1.9, and 8.1.10 to use the new command?

jdinan:
@BryantLam I don't think that this requires a changelog entry, since there is no user-visible change.  In fact, if you drop the changelog entry, I would be inclined to handle this as a document edit.

jdinan:
@BryantLam Would you like to bring this forward for voting at the next meeting?  It looks like the only change needed is to remove the changelog entry.

BryantLam:
@jdinan Removed changelog entry. Converting this PR to a DocEdit as you suggested. Feel free to merge when reviewed/ready.

jdinan:
Looks good to me.  I'm fine with processing this as a document edit.  Can I get a +1 from another committee member?

swpoole:
+1

Best,
Steve (ADTSC)

PR #51, opened 2017-02-11
Fixed erroneous API description for shmem_info_get_name

BryantLam:
`shmem_info_get_name`'s API description is wrong. It references version numbers (and didn't do it well) instead of `SHMEM_VENDOR_STRING`. **Edit:** The API description for this function is almost entirely wrong.

Also, I hate this function. It shouldn't have made it into the spec. It seems easy for a user (or implementer!) to inadvertently cause a buffer overflow when they could instead just use `SHMEM_VENDOR_STRING`. Who is using this function?

Piggy-backed commit: Minor textual change to API description for `shmem_info_get_version` for consistency. **Edit:** I hate this function too.


jdinan:
Could you push a commit that also deletes the ```\const{}``` command from ```defs.tex``` to make sure we don't use the wrong command in the future?  This looks like it's left over from the deprecated sources.

jdinan:
I think this is more than we can handle as a document edit.  At the upcoming meeting, we should discuss forming section committees to help with these types of changes and also to help us prepare for the next spec release.

BryantLam:
Referencing #52 for `\const{}`.

I agree that this is not a document edit. The API description is changed quite significantly.

nspark:
@jdinan Is this a PR change we should read at the next meeting? (If so, I'll put it on the agenda for the next meeting.) Or should this be handled by the section committee?

jdinan:
Yes, although the changelog entries should be updated.  There's no need for a changelog entry for ```shmem_info_get_version```.  The changelog entry for shmem_info_get_name should read more like:
```
\item Clarified API description of \FUNC{shmem\_info\_get\_version}.
\\See Section \ref{subsec:shmem_info_get_version}.
```

jeffhammond:
If SHMEM_VENDOR_STRING is preprocessor symbol, it is very useful to have a
runtime call to get the same info so that one can confirm that the
implementation used to compile matches the one used to link.

On the other hand, I've never seen code that checks this consistency.
-- 
Jeff Hammond
jeff.science@gmail.com
http://jeffhammond.github.io/

jdinan:
@jeffhammond 
https://github.com/Sandia-OpenSHMEM/SOS/blob/master/test/unit/shmem_info_f.f90
https://github.com/Sandia-OpenSHMEM/SOS/blob/master/test/unit/shmem_info.c

We check it in the Fortran test, even though it's not required by the spec, to make sure the language bindings are implemented correctly.  Once the spec requires the strings to be identical, I'll add the same check to the C test.

jeffhammond:
Edit: I've never seen any code other than unit tests that does this 😆

shamisp:
@jeffhammond this is particularly useful for SHMEM implementation with "inline"

jeffhammond:
I am definitely in favor of this feature. I don't know why that didn't come
across in my initial comment.
-- 
Jeff Hammond
jeff.science@gmail.com
http://jeffhammond.github.io/

naveen-rn:
This a very useful change.  

At present, we don't have any solid requirements on what the SHMEM_VENDOR_STRING should return. We just say, it is a string representing vendor name and this statement seems to be too broad. 

1. SOS and Reference implementation returns just the vendor identity string - "Sandia SHMEM" and "UH Reference Implementation" respectively. This will be identical across different versions of the implementation.
2. In Cray SHMEM, it returns both the vendor identity and vendor version related details - Cray SHMEM 7.3. The major and minor version details will be updated on each release and will be different across different releases. 

On implementations that follow Cray-like approach, with this change we would introduce backward compatibility issues on linking against different versions of the same implementation. For completeness, along with this change can we also clarify and restrict the SHMEM_VENDOR_STRING to just indicate the vendor identity?

jdinan:
It's unfortunate that we didn't use consistent API naming between ```SHMEM_VENDOR_STRING``` and ```shmem_info_get_name```...

jeffhammond:
@jdinan It could be worse (e.g. `MPI_Reduce_scatter(_block)` :rofl:)

content/backmatter.tex
```diff
@@ -465,6 +465,12 @@ \section{Version 1.4}
 \item Clarified deprecation overview and added deprecation rationale in Annex F.
 \\See Section \ref{sec:dep_api}.
 %
+\item Applied non-functional, consistency-related change to API description for \FUNC{shmem\_info\_get\_version}.
```
BryantLam: Action: DocEdit. Remove this item.

content/backmatter.tex
```diff
@@ -465,6 +465,12 @@ \section{Version 1.4}
...
+\\See Section \ref{subsec:shmem_info_get_version}.
+%
+\item Fixed erroneous API description for \FUNC{shmem\_info\_get\_name}.
```
BryantLam: Action: Fixed --> Clarified.

jdinan: Suggest s/Fixed erroneous/Clarified/.

content/shmem_info_get_name.tex
```diff
@@ -19,17 +19,18 @@
...
+    The implementation
+    copies the vendor defined string of size at most \CONST{SHMEM\_MAX\_NAME\_LEN} to \VAR{name}.
+    In \CorCpp{}, the string is
+    terminated by a null character.  In \Fortran{}, the string of size less than
+    \CONST{SHMEM\_MAX\_NAME\_LEN} is padded with blank characters up to size
+    \CONST{SHMEM\_MAX\_NAME\_LEN}.
+    If the \VAR{name} memory buffer is provided with size less than
+    \CONST{SHMEM\_MAX\_NAME\_LEN}, behavior is undefined.
+    For a given library implementation, the vendor string
+    returned is consistent with the library constant \CONST{SHMEM\_VENDOR\_STRING}.
```
BryantLam: Action: Resolve inconsistency with definition of SHMEM_VENDOR_STRING. (Vendor-defined string as per description in `shmem_info_get_name`, or or vendor identity (defined by `SHMEM_VENDOR_STRING`.)

jdinan: Item 2 is an ABI compatibility issue, which the specification currently chooses to not address.  We don't specify the values for any of the other contants (psync sizes, comparison constants, etc.) so I don't think there's any action needed on that item.

BryantLam: This function's signature should be
```
const char *shmem_info_get_name()
```
which would solve the buffer overrun issue. If the user only wants to print this string, `SHMEM_MAX_NAME_LEN` would no longer be needed. If the user wants to manipulate this string, they're forced to copy it.

Also:
```
int shmem_info_get_major_version();
int shmem_info_get_minor_version();
```

jdinan: IIRC, the rationale for the existing get_name API was that the library may create that string during the call (e.g. based on dynamic linking with other parts of the communication stack).  I agree that what you suggested is much easier to use given the clarified definition we're currently working with (that it should be the same as the constant).

content/backmatter.tex
```diff
@@ -501,6 +501,12 @@ \section{Version 1.4}
...
 %
+\item Clarified description for \CONST{SHMEM\_VENDOR\_STRING}.
+\\See Section \ref{subsec:library_constants}.
+%
+\item Clarified API description for \FUNC{shmem\_info\_get\_name}.
+\\See Section \ref{subsec:shmem_info_get_name}.
+%
 \end{itemize}
 
 \section{Version 1.3}
```
BryantLam: Changes since reading
- New changelog entry for changes to `SHMEM_VENDOR_STRING`
- Fixed --> Clarified

content/library_constants.tex
```diff
@@ -126,10 +126,11 @@
...
-and whitespace padded.  It can also be equal in length to \const{SHMEM\_MAX\_NAME\_LEN} 
-since Fortran does not NULL terminate strings. \tabularnewline
+String representing vendor defined information of size at most
+\const{SHMEM\_MAX\_NAME\_LEN}.
+In \CorCpp{}, the string is terminated by a null character.  In \Fortran{}, the
+string of size less than \const{SHMEM\_MAX\_NAME\_LEN} is padded with blank
+characters up to size \const{SHMEM\_MAX\_NAME\_LEN}. \tabularnewline
 \hline
 
 \end{tabular}
```
BryantLam: Changes since reading
- Relaxed the semantics of `SHMEM_VENDOR_STRING`. Semantics changed from vendor name to vendor-defined string since it is more useful overall. For vendors that want to provide just a name, this change is fully backwards compatible!

content/shmem_info_get_version.tex
```diff
@@ -21,8 +21,8 @@
 \apidescription{
     This routine returns the major and minor version of the \openshmem{} standard
     in use.  For a given library implementation, the major and minor version
-    returned by these calls is consistent with the compile-time constants,
-    SHMEM\_MAJOR\_VERSION and SHMEM\_MINOR\_VERSION, defined in its \HEADER{shmem.h}.
+    returned by these calls are consistent with the library constants
+    \CONST{SHMEM\_MAJOR\_VERSION} and \CONST{SHMEM\_MINOR\_VERSION}.
 }
 
 \apireturnvalues{
```
BryantLam: Changes since reading
- Grammar: is --> are

PR #52, opened 2017-02-11
DocEdit: Added \HEADER command

BryantLam:
Added `\HEADER` command. Affects Annex A and API descriptions for `shmem_info_get_{version,name}`.

Also, there's a defs comment for \const{} vs. \CONST{}.


jdinan:
This is a document edit -- can I please get a review and +1 from another committee member so we can merge?  Thanks!

nspark:
:+1:

PR #53, opened 2017-02-11
Differentiate shmemx and SHMEMX prefixes

BryantLam:
**[Work in Progress]** The spec is inconsistent with the use of e.g., `shmemx_` for routine names and variable names and `SHMEMX_` for library constants (See almost all uses of `\shmemprefix`). This is made even more apparent with inconsistent application of `\FUNC` vs. `\CONST` for prefix uses. Do we want to address this issue?

Also, note that the text includes variable names. Are these specifically environment variables? (**Edit:** Environment variables should be using `SHMEMX_` instead of `shmemx_`.) What other variables are there if not function-local parameters?

### DIFF
--- a/content/language_bindings_and_conformance.tex
+++ b/content/language_bindings_and_conformance.tex
@@ -16,4 +16,5 @@ variables, or identifiers with the prefix \shmemprefix (for \Clang{} and
 
 All \openshmem extension \ac{API}s that are not part of this specification must
 be defined in the \FUNC{shmemx.h} include file. These extensions shall use the
-\FUNC{shmemx\_} prefix for all routine, variable, and constant names.
+\FUNC{shmemx\_} prefix for routine names and variable names and the
+\CONST{SHMEMX\_} prefix for library constants.
--- a/content/language_bindings_and_conformance.tex
+++ b/content/language_bindings_and_conformance.tex
@@ -11,8 +11,8 @@ specification.
 implementing the interfaces using macros is strongly discouraged as this could
 severely limit the use of external profiling tools and high-level compiler
 optimizations. An \openshmem program should avoid defining routine names,
-variables, or identifiers with the prefix \shmemprefix (for \Clang{} and
-\Fortran), \shmemprefixC (for \Clang) or with \openshmem \ac{API} names.
+variable names, or identifiers with the prefix \shmemprefix{} (for \Clang{} and
+\Fortran), \shmemprefixC{} (for \Clang) or with \openshmem \ac{API} names.
 
 All \openshmem extension \ac{API}s that are not part of this specification must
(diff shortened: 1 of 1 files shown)


nspark:
> This is made even more apparent with inconsistent application of `\FUNC` vs. `\CONST` for prefix uses. Do we want to address this issue?

I agree that, in general, the specification is inconsistent in its use of the {FUNC, CONST, VAR} commands. It is also inconsistent in its use of uppercase (Fortran-style) and lowercase (C-style) function names. (I'd prefer to use C-style names. Go figure.)

I'm not sure whether this PR as it currently stands is big (or small) enough to require (or skip) the full process; i.e., is it just a doc edit? If it is just a doc edit, I'd support rolling in changes for consistent use of the styling commands and capitalization. If not, I think we should fix those as a separate doc edit.

jdinan:
This is the sort of thing that chapter committees are good for.  Given that we probably want to ratify a spec this year, this seems like a decent time to create them.

content/language_bindings_and_conformance.tex
```diff
@@ -11,9 +11,10 @@
...
 optimizations. An \openshmem program should avoid defining routine names,
-variables, or identifiers with the prefix \shmemprefix (for \Clang{} and
-\Fortran), \shmemprefixC (for \Clang) or with \openshmem \ac{API} names.
+variable names, or identifiers with the prefix \shmemprefix{} (for \Clang{} and
+\Fortran), \shmemprefixC{} (for \Clang) or with \openshmem \ac{API} names.
 
 All \openshmem extension \ac{API}s that are not part of this specification must
 be defined in the \FUNC{shmemx.h} include file. These extensions shall use the
-\FUNC{shmemx\_} prefix for all routine, variable, and constant names.
+\FUNC{shmemx\_} prefix for routine names and variable names and the
```
jdinan: This would prohibit variables with uppercase names, which seems a bit much.

BryantLam: Sorry; I'm not sure I follow.

nspark: I think @jdinan's point is that, as proposed, the `SHMEMX_` prefix is reserved for constants (not variables or functions). Perhaps @BryantLam could use the term "integer constant expressions" or "preprocessor constants" to distinguish between `const int shmemx_*` constants and a constant suitable for evaluation by the preprocessor.

jdinan: I don't think it's necessary to get this specific about when the lower/uppercase names are used.  Can we change this to: "These extensions shall use the \FUNC{shmemx\_} or \FUNC{SHMEMX_} prefix for all routine, variable, and constant names."

jdinan:
@BryantLam Do you want to revive this for OpenSHMEM 1.5?

PR #54, opened 2017-02-12
Remove Deprecated Sources

jdinan:
Proposed document edit (i.e. repository maintenance) to remove the deprecated sources from the ```osh_spec_next``` branch and clean up some LaTeX code that is no longer used by the current specification.  As part of this change, I will also add a branch called ```deprecated-sources``` on commit 6af7eb7f so that we can easily find these sources again, if needed.


nspark:
:+1:

jdinan:
Could I get another +1 on this (or -1 if you don't like it)?

I'm proposing that we move the deprecated sources to a ```deprecated-sources``` branch rather than keeping them on ```osh_spec_next```.

swpoole:
+1

Best,
Steve (ADTSC)

jdinan:
I created the deprecated-sources branch on 01515de8.

PR #48, opened 2017-02-08
Deprecate the Fortran API

nspark:
This PR shall deprecate the Fortran API.

- All Fortran-specific API declarations are marked as deprecated.
- Deprecation is noted under "Programming Model Overview", "Language Conformance", deprecation listing and rationale, and the changelog.
- A few not-totally-trivial Fortran references are explicitly marked.

See also Issue #36.

PR #61, opened 2017-03-31
Add deprecation markers to cache API

jdinan:
Add deprecation begin/end markers around cache API description.

### DIFF
--- a/content/shmem_cache.tex
+++ b/content/shmem_cache.tex
@@ -1,10 +1,10 @@
-\StartDeprecateBlock \\
 \apisummary{
     Controls data cache utilities.
 }
 
 \begin{apidefinition}
 
+\begin{DeprecateBlock}
 \begin{Csynopsis}
 void shmem_clear_cache_inv(void);
 void shmem_set_cache_inv(void);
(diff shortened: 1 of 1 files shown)


jdinan:
[PDF of proposed changes](https://github.com/openshmem-org/specification/files/886573/main_spec.pdf)

jdinan:
@nspark Please review.

nspark:
@jdinan I would prefer use to use the `DeprecateBlock` environment:
```latex
\begin{DeprecateBlock}
...
\end{DeprecateBlock}
```
If this doesn't work due to LaTeX sensitivities, then I think what you have is fine.

The only other comment, which is mostly stylistic, is that for PR #48 (Fortran deprecation), I only wrapped the API listing in the synopsis section with the deprecation markings. If we prefer consistency, I'd probably favor the style of #48 because it's much easier in LaTeX form, but I see the benefit of marking the whole of the entry.

jdinan:
Sure, here's the updated PDF: [main_spec.pdf](https://github.com/openshmem-org/specification/files/886709/main_spec.pdf)

I think this should merge relatively easily with your Fortran ticket.  We'll just need to remove the extra DeprecateBlock environment.

nspark:
:+1: Looks good. Thanks, Jim!

jdinan:
We need another +1 to merge this change.  @manjugv @shamisp Can you help?

shamisp:
👍

PR #63, opened 2017-04-13
Communication Contexts Extension

jdinan:
This proposal adds support for communication contexts.

References:
* [Redmine #177](http://www.openshmem.org/redmine/issues/177)
* "Contexts: A Mechanism for High Throughput Communication in OpenSHMEM," James Dinan and Mario Flajslik.  PGAS '14.  [[doi](http://dl.acm.org/citation.cfm?doid=2676870.2676872)] [[pdf](https://pdfs.semanticscholar.org/8b8d/8d76c4cb14924172b06b152dd939293bc6a0.pdf)]


jdinan:
Current draft: [2017-04-13 -- Contexts Draft.pdf](https://github.com/openshmem-org/specification/files/920852/2017-04-13.--.Contexts.Draft.pdf)

content/shmem_ctx.tex
```diff
@@ -0,0 +1,133 @@
...
+communication and computation.  In multithreaded environments, contexts can
+additionally provide thread isolation, eliminating overheads resulting from
+thread interference within the communication subsystem.
+
+Context handles are of type \CTYPE{shmem\_ctx\_t} and are valid for
+language-level assignment and comparison.  A handle to the desired context is
+passed as an argument in the \Cstd \CTYPE{shmem\_ctx\_*} and type-generic API
+routines.  API routines that do not accept a context argument operate on the
+default context.  The default context can be used explicitly through the
+\const{SHMEM\_CTX\_DEFAULT} handle.
```
BryantLam: `\const` --> `\CONST`
We should DocEdit this command since the styling for `\const` is better but `\CONST` is already used everywhere except the Library Constants section.

content/shmem_ctx.tex
```diff
@@ -0,0 +1,133 @@
...
+
+\begin{apiarguments}
+    \apiargument{IN}{ctx}{Handle to the context that will be destroyed.}
+\end{apiarguments}
+
+\apidescription{
+    \FUNC{shmem\_ctx\_destroy} destroys a context that was created by a call to
+    \FUNC{shmem\_ctx\_create}.  It is the user's responsibility to ensure that
+    the context is no longer in use before calling this routine.  This function
+    performs an implicit quiet operation on the given context before it is freed.
```
BryantLam: What cases are there for the user to manually ensure responsibility for the non-use of a context before calling `destroy` if `destroy` has a `ctx_quiet` inside? Did you intend something along the lines of "users should not use this context after it is destroyed"?

content/shmem_ctx.tex
```diff
@@ -0,0 +1,133 @@
...
+\begin{apidefinition}
+
+\begin{Csynopsis}
+int shmem_ctx_create(long options, shmem_ctx_t *ctx);
+\end{Csynopsis}
+
+\begin{apiarguments}
+    \apiargument{IN}{options}{The set of options requested for the given context.
+        Multiple options may be requested by combining them with a bitwise
+        \textit{OR} operation.  Otherwise, $0$ can be given if no options are
```
BryantLam: - `bitwise \textit{OR}` has a style conflict with PR #31 (@nspark) -- Is `bitwise OR` sufficient?
- `$0$` has style conflict with descriptions like in `shmem_my_pe`'s use of `\CONST{0}`.

content/shmem_ctx.tex
```diff
@@ -0,0 +1,133 @@
...
+\label{subsec:shmem_ctx_create}
+\apisummary{
+    Create a communication context.
+}
+
+\begin{apidefinition}
+
+\begin{Csynopsis}
+int shmem_ctx_create(long options, shmem_ctx_t *ctx);
+\end{Csynopsis}
```
BryantLam: Typically, when I see a creation API of this style, the created object is the first argument, not the last. Is there a particular reason this prototype was chosen?

content/shmem_ctx.tex
naveen-rn: I suppose it was added to handle the scenario where multiple threads use the same context handle. Users have to make sure such context handles are not being used by other threads, when one of the threads call destroy.

content/shmem_ctx.tex
jdinan: Fixed

content/shmem_ctx.tex
jdinan: @nspark wins.  Fixed.

content/shmem_ctx.tex
jdinan: No particular reason.  For example, the output communicator is the last argument to the MPI communicator construction routines.

content/shmem_ctx.tex
jdinan: Updated to:

    It is the user's responsibility to ensure that the context is not used after it has been
    destroyed, for example when the destroyed context is used by multiple threads.

jdinan:
Updated draft: [2017-04-25 -- Contexts Draft.pdf](https://github.com/openshmem-org/specification/files/956145/2017-04-25.--.Contexts.Draft.pdf)

jdinan:
Updated draft: [2017-04-25 -- Contexts Draft.pdf](https://github.com/openshmem-org/specification/files/956247/2017-04-25.--.Contexts.Draft.pdf)

jdinan:
Note from last review: @tonycurtis requested a diagram showing logical mapping of contexts to communication operations.  We could use a "tube" diagram similar to what I have used in past presentations.  I think it could be incorporated into the memory model section without too much difficulty.  Do folks want this?  If so, I don't own a license to OmniGraffle.  How should I go about generating the figure?

shamisp:
Few questions:
1. How the concern about the receive side of the context was addressed 
2. Some API changes pushes number of arguments to six which leaves no space for adding groups. In addition once both introduced it will be very confusing.

jdinan:
(1) Receive side contexts must be automatically managed by the runtime with the current API (e.g. created during initialization and automatically mapped to contexts).  This is a consequence of removing domains from the proposal.  The WG's current plan is to move ahead with this version of the contexts API and revisit adding domains in the future once we have a better understanding of what problems we need them to solve and how.  I followed-up with Pavan on this and I think we're in sync.  We don't anticipate much disruption to the API adding domains in the future (it adds another contexts constructor); we don't need to get all of the API in the first proposal, but it's better not to get parts of it wrong.

(2) The choices are (a) delay contexts until we can sort this out or (b) try to include contexts in 1.4 and accept the risk to changing the API.  The API risk is that we potentially add more routines that accept the group argument.  This is a non-issue for the C11 interface, since we can use the same approach we are using to add contexts without increasing the number of C11 bindings.  We can poll the committee on this issue at today's meeting.

Relevant to both (1) and (2) -- I have been operating under the marching orders that contexts are a critical feature for users and we are willing to accept some risk to make sure it happens.

shamisp:
(1) I don't think domain is related to coordination with target/remote resources. it is much more challenging issue, which may lead to substantial changes in semantics.

(2) C11 can not help with the fact of introducing confusing API and exploding the list of arguments.

Relevant for (1) +(2): we should strive towards good and reasonable technical proposal

jdinan:
Managing RX-side resources may require changes to the domains API, relative to what was proposed earlier, but the domain is intended to allow users control over binding of resources to contexts.  A correctly defined domain should address the problem of user-level resource management.

Relevant to (2), I don't disagree with the goal you stated.  Because of the timing of different proposals, we have a choice between delaying contexts versus accepting some risk of API ugliness.  IIUC, contexts are important enough to our user community that the risk of API churn/ugliness is acceptable.

content/shmem_malloc.tex
```diff
@@ -54,7 +54,11 @@
     whether an existing allocation is modified and whether new memory is allocated.
     This ensures that all
     \ac{PE}s participate in the memory allocation, and that the memory on other
-    \ac{PE}s can be used as  soon as the local \ac{PE} returns.  The user is
+    \ac{PE}s can be used as  soon as the local \ac{PE} returns.
+    \newtext{The user is responsible for ensuring that no communication
```
abouteiller: There should be no particular complications for an app that communicates on a private context simultaneously, for as long as it never targets/use as source the region that gets modified by the shrealloc/shfree; But that text implies this is illegal code. Is this intentional?

jdinan: That's the intended semantic; however, I think you're right that the text is too restrictive.  I just pushed a fix.  Could you take a look at the new text?

abouteiller: 👍

abouteiller:
The update looks good to me. 👍

jdinan:
[2017-05-02 -- Contexts Draft.pdf](https://github.com/openshmem-org/specification/files/1029745/2017-05-02.--.Contexts.Draft.pdf)

example_code/shmem_ctx_pipelined_reduce.c
```diff
@@ -0,0 +1,56 @@
...
+
+    shmem_init();
+    int me = shmem_my_pe();
+    int npes = shmem_n_pes();
+
+    pbuf[0] = shmem_malloc(PLEN * npes * sizeof(int));
+    pbuf[1] = shmem_malloc(PLEN * npes * sizeof(int));
+
+    int ret_0 = shmem_ctx_create(0, &ctx[0]);
+    int ret_1 = shmem_ctx_create(0, &ctx[1]);
```
BryantLam: Observation: Private contexts would allow `sync` be replaced with `barrier` without significant performance penalty.

abouteiller: I do not think so: If the contexts are private, the barrier would have no effect, on either of them. 

Now, that has raised my attention to the fact that barrier_all has an effect on all shared contexts. 

The original intent of PRIVATE/SHARED was to pertain to thread locality (i.e. an operation on a private context does not have to care for making it consistent with possible concurrent access from other threads). I think the conflation of two different concepts here is a detriment with PRIVATE/SHARED being confounded with DEFAULT/Non default. 

I was also discussing during the call yesterday that we could have an "IN: ctx" as the last optional param to shmem_barrier[_all]; Jim had a good comeback that in many instances you may want to quiet a couple of ctx before doing the barrier, which would make the usage asymmetric, which is kinda ugly.
```
shmem_ctx_quiet(ctx1);
shmem_ctx_quiet(ctx2);
shmem_barrier_all(ctx3);
```

naveen-rn: > Observation: Private contexts would allow sync be replaced with barrier without significant performance penalty.

FWIU sync can be implemented on a separate unique resource which is completely independent of the progress on resources mapped to any SHARED/PRIVATE/DEFAULT context. If that is the case, then there will be significant performance difference.

jdinan:
As per discussion in the threading WG, I have removed implicit quiescence of shareable contexts (although the term "shareable" is still used, they are no longer automatically quiesced in barriers).  The rationale is to simplify the contexts proposal and save this feature for future expansion if it improves the model.

[2017-06-12 -- Contexts Draft.pdf](https://github.com/openshmem-org/specification/files/1069610/2017-06-12.--.Contexts.Draft.pdf)

content/shmem_barrier_all.tex
```diff
@@ -42,6 +44,12 @@
     The \FUNC{shmem\_barrier\_all} routine can be used to
     portably ensure that memory access operations observe remote updates in the order
     enforced by initiator PEs.
+
+    \newtext{The \FUNC{shmem\_barrier\_all} routine does not ensure completion
+    of operations performed on contexts with the \CONST{SHMEM\_CTX\_PRIVATE}
+    option enabled.  Calls to \FUNC{shmem\_ctx\_quiet} can be performed prior
+    to calling the barrier routine to ensure completion of operations issued on
+    private contexts.}
 }
```
abouteiller: Given that the semantic has been changed wrt shared-contexts, I feel like this small paragraph needs to be adapted as well.

jdinan: Thanks!  Should be fixed now in 2cea079.

jdinan:
Special ballot patches:
* Remove private ctx text in barrier  …			2cea079
* Add back missing newtext macro			3cb3b92
* Clarify ctx destruction and finalize semantics  …			be9deb6
* Incorporate improved ctx argument description			15464f9

jdinan:
[2017-10-09 -- Contexts Draft.pdf](https://github.com/openshmem-org/specification/files/1211516/2017-10-09.--.Contexts.Draft.pdf)

nspark:
Here's an [example demonstrating how to resolve the issue with C11 Generics](https://gist.github.com/nspark/ac07bd57e72db32d61c142ea38d5d162) for `shmem_inc()` and other interfaces that take only two arguments, however, it relies on a GNU C Preprocessor extension.

nspark:
False alarm. :wink: We _can_ portably "overload" `shmem_inc` for context and non-context calls.

Here's [an example of resolving the issue without any (non-standard) preprocessor extensions](https://gist.github.com/nspark/11b8aa58fee14397953afbabdf0e514e).

jdinan:
Special ballot: 340729c..5da7a9c

content/shmem_ctx.tex
```diff
@@ -64,6 +64,13 @@ \subsubsection{\textbf{SHMEM\_CTX\_CREATE}}
 
         \apitablerow{\CONST{SHMEM\_CTX\_PRIVATE}}{
             The given context will be used only by the thread that created it.}
+
+        \apitablerow{\CONST{SHMEM\_CTX\_NOSTORE}}{
+            Quiet and fence operations performed on the given context are not
```
manjugv: The previous two options are requirements on the applications, and the last one is on the library. To make it clear, consider change to the sentence : " When Quiet and fence operations are performed on the given context, the implementation is not required to enforce ... "

jdinan: This is also a requirement on the application; it cannot rely on quiet/fence on the given context to order store operations.  The option conveys information to the OpenSHMEM library, but it doesn't introduce a new requirement (that is, the library can ignore this option and still be a correct implementation).  Would this be clearer?  "The application does not require quiet and fence operations performed on the given context to complete or order store operations."

manjugv:
> "The application does not require quiet and fence operations performed on the given context to complete or order store operations."

It is a bit confusing. You might want to say that the applications are responsible for ordering and completion of the stores than rely on the library.

content/shmem_ctx.tex
jdinan: Ok, here's another attempt:

Quiet and fence operations performed on the given context are not required to enforce completion and ordering of memory store operations.  When ordering of store operations is needed, the application must perform a synchronization operation on a context without SHMEM_CTX_NOSTORE option enabled.

jdinan:
Special ballot: [340729c..59f9df8](https://github.com/openshmem-org/specification/compare/15464f9...59f9df8)

jdinan:
Ready for review!  Here's the PDF: [main_spec.pdf](https://github.com/openshmem-org/specification/files/1270805/main_spec.pdf)

@nspark Patch 9488b76 will require approval from your section committee.

jdinan:
All, we are cutting it close since the section committee deadline is tomorrow.  I'd like to get +2 on this before I merge it (I'm not counting my own review, since I've stared at this text so long that I'm blind to any errors).  Who is available to review?

nspark:
@jdinan Reviewing it now

nspark:
@swpoole: I think the distinction in the example is that a library-level error would leave any subsequent OpenSHMEM call, including `shmem_global_exit(2)`, as undefined behavior. In the example, it is handled as an application-level error. I can appreciate that, as an example, this distinction is subtle.

Possibly, the specification could make a stronger claim that, as specified, a call to `shmem_ctx_create` cannot leave the library in a state where the behavior of subsequent library calls are undefined.

example_code/shmem_ctx_pipelined_reduce.c
```diff
@@ -0,0 +1,56 @@
...
+    if (ret_0 || ret_1) shmem_global_exit(1);
```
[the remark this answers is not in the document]
nspark: This still does not render properly.

utils/defs.tex
```diff
@@ -167,7 +167,7 @@
 \definecolor{gray}{rgb}{0.92,0.92,0.92}
 
 \lstset{ % set defaults for languages not otherwise defined
-  breakatwhitespace=false,         % sets if automatic breaks should only happen at whitespace
+  breakatwhitespace=true,         % sets if automatic breaks should only happen at whitespace
   basicstyle=\ttfamily\footnotesize,
   breaklines=true,                 % sets automatic line breaking
   escapeinside={|}{|},          % if you want to add LaTeX within your code
```
nspark: @jdinan I wonder whether this is the source of the LaTeX rendering issues for the logical OR operator `||`.

nspark: Yes, this seems to be the issue. According to my build, the `escapeinside={|}{|},` line can be removed safely or the delimiters can be changed.

content/shmem_ctx_destroy.tex
```diff
@@ -0,0 +1,64 @@
...
+    It is invalid to pass \CONST{SHMEM\_CTX\_DEFAULT} to this routine.
+
+    Destroying a context makes it impossible for the user to complete
+    communication operations that are pending on that context.  This includes
+    nonblocking communication operations, whose local buffers are only returned
+    to the user after the operations have been completed.  An implicit quiet is
+    performed when freeing a context to avoid this ambiguity.
+
+    A context with the \CONST{SHMEM\_CTX\_PRIVATE} option enabled must be
+    destroyed by the thread that created it.
```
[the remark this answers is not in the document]
jdinan: It's specified in ```shmem_finalize```.  We included it in notes here as a cross-reference.

utils/defs.tex
jdinan: Yes, this is the source of the problem.  I thought I had fixed it by changing the escapeinside for only the ```apiexamples``` ```lstset```, but I apparently didn't commit the change.  Unless another example is using this (I don't think we currently have escapes in any examples, but I also didn't check), it would be safest to just remove it entirely.

jdinan:
Those changes all look good.  I would prefer to merge this into osh_spec_next and then have you commit the edits to your section committee branch.  Otherwise, the reading at the RCM will get tricky (since we don't read text that was already voted on).

utils/defs.tex
nspark: I ran `grep -F '|' example_code/*.*` and only turned up your one example. :+1: for removing the line entirely.

jdinan: Listings documentation indicates that escapeinside defaults to nothing (no escapes), so it should be safe to not specify an escape sequence.  :+1:

jdinan:
Ok, whose arm can I twist for a second review?

nspark:
Does @swpoole's comment count? Or, maybe @BryantLam can review?

swpoole:
I would hope it counts. Just to be sure.

+1

BryantLam:
There will be numerous conflicts in the front matter and RMA sections, but the raw textual diff looks good.

jdinan:
Thanks, all -- I'll merge and hand it over to @nspark's section committee!

Sorry @swpoole, didn't see your +1.  @nspark has your comment and will tackle it in the RMA/AMO section committee.

PR #64, opened 2017-04-14
Add sync routines

jdinan:
This proposal adds the ```shmem_sync_all``` and ```shmem_sync``` routines.  The sync routines are similar to the existing barrier routines, with the difference that they do not include an implicit quiet operation.


jdinan:
[2017-04-13 -- Sync Draft.pdf](https://github.com/openshmem-org/specification/files/921257/2017-04-13.--.Sync.Draft.pdf)

content/shmem_sync_all.tex
```diff
@@ -0,0 +1,44 @@
+\color{ForestGreen}
+\apisummary{
+    Registers the arrival of a \ac{PE} at a barrier and suspends \ac{PE}
```
manjugv: In multithreaded OpenSHMEM, you don’t have to suspend the PE rather just the thread ?

content/shmem_sync_all.tex
```diff
@@ -0,0 +1,44 @@
...
+
+\begin{apiarguments}
+
+    \apiargument{None.}{}{}
+
+\end{apiarguments}
+
+\apidescription{
+    The \FUNC{shmem\_sync\_all} routine registers the arrival of a \ac{PE} at a
+    barrier. Barriers are a fast mechanism for synchronizing all \acp{PE} at
```
manjugv: barrier -> sync routine? 

"Barriers are a fast mechanism for" This sentence seems redundant.

content/shmem_sync_all.tex
```diff
@@ -0,0 +1,44 @@
...
+    once.  This routine causes a \ac{PE} to suspend execution until all
+    \acp{PE} have called \FUNC{shmem\_sync\_all}.
+
+    In contrast with the \FUNC{shmem\_barrier\_all} routine,
+    \FUNC{shmem\_sync\_all} only ensures completion of previously issued memory
+    stores and does not ensure completion of remote memory updates issued via
+    \openshmem routines.
+}
```
manjugv: What if OpenSHMEM implements shmem_put/shmem_get as memory routines ?

content/shmem_sync_all.tex
```diff
@@ -0,0 +1,44 @@
...
+
+\apireturnvalues{
+    None.
+}
+
+\apinotes{
+    The \FUNC{shmem\_sync\_all} routine can be used to portably ensure that
+    memory access operations observe remote updates in the order enforced by
+    initiator PEs, provided that the initiator PE ensures completion of remote
```
manjugv: PEs -> \acp{PE}

content/shmem_sync_all.tex
jdinan: This is the same text that we have for barrier.  Because the proposals are happening in parallel, it's a little tricky.  If sync moves ahead at this meeting, we can handle it as part of the threading proposal before voting.  Otherwise, we could use the section committee.

I think you mean, suspend the calling thread rather than all threads in the PE.  That's the semantic that I would expect to see for barriers in the threading proposal.

content/shmem_sync_all.tex
jdinan: I think "barrier" is the accepted terminology for this type of operation, so I'm ok with leaving it as-is.  The second sentence is defining barrier, which seems non-redundant.

content/shmem_sync_all.tex
jdinan: The semantics here define the minimum completion guarantees that this function must provide.  It should be fine to generate additional completions.

content/shmem_sync_all.tex
jdinan: Fixed -- thanks for the review!

content/shmem_sync.tex
```diff
@@ -0,0 +1,83 @@
...
+    the \activeset calls an \openshmem collective routine, undefined behavior
+    results.
+
+    The values of arguments \VAR{PE\_start}, \VAR{logPE\_stride}, and
+    \VAR{PE\_size} must be equal on all \acp{PE} in the \activeset.  The same
+    work array must be passed in \VAR{pSync} to all \acp{PE} in the \activeset.
+
+    In contrast with the \FUNC{shmem\_barrier} routine, \FUNC{shmem\_sync} only
+    ensures completion of previously issued memory stores and does not ensure
+    completion of remote memory updates issued via \openshmem routines.
```
shamisp: It is kinda a bit odd that sync only imposes partial quiet semantics. What is justification for such definition ?

content/shmem_sync_all.tex
manjugv: Yes, suspend the calling thread.

content/shmem_sync.tex
jdinan: The example is probably the best explanation.  We quiet memory updates because this is an expected usage model.

content/shmem_sync_all.tex
manjugv: Without considering threads, what should be behavior of non-blocking operations during "suspend" ?

jdinan: Sync doesn't provide completion guarantees for SHMEM communication routines.  Take a look at the new pipelining example in the contexts proposal.

manjugv: I understand the difference :) Probably, the confusion is because of the usage of “suspend”.  

Suggestion : suspend -> block the calling thread

content/shmem_sync.tex
shamisp: I would suggest to generalize it a bit to memory load and stores.

content/shmem_sync_all.tex
jdinan: This sounds familiar, did we discuss this recently in the context of a different proposal that updated the barrier text?  IIRC, we preferred "suspends execution" over "blocks".

content/shmem_sync.tex
jdinan: I took this text directly from fence/quiet, which we spent several months arguing over.  :)  So, I'm hesitant to deviate too much from that.  What might be non-controversial is to update "ensures completion of ..." to "ensures completion and visibility of ..."  Does that address your concern?

content/shmem_sync_all.tex
manjugv: From this PR perspective, at least for me, it seems more appropriate to use “block the calling thread”.  When you make a call to sync, all non-blocking operations might be still progressing, is that a correct behavior, when you “suspend the execution”.

hmm .. I don’t recall the discussion. However, I will not defend the use of “suspends execution” in barrier. :)

jdinan: Oh, I see what you're talking about.  I'm not sure that swapping in "block" fixes this.  I think we would have the same potential ambiguity with the wait operations and potentially also other collectives that block (e.g. for non-root PEs in broadcast).  If we think this is an ambiguity that needs to be fixed, something at a higher level might be needed.

content/shmem_sync_all.tex
```diff
@@ -0,0 +1,44 @@
...
+\apireturnvalues{
+    None.
+}
+
+\apinotes{
+    The \FUNC{shmem\_sync\_all} routine can be used to portably ensure that
+    memory access operations observe remote updates in the order enforced by
+    initiator \acp{PE}, provided that the initiator PE ensures completion of remote
+    updates with a call to \FUNC{shmem\_quiet} prior to the call to
+    \FUNC{shmem\_sync\_all} routine.
```
jdinan: ^the

content/shmem_sync.tex
```diff
@@ -0,0 +1,83 @@
...
+
+    The same \VAR{pSync} array may be reused on consecutive calls to
+    \FUNC{shmem\_barrier} if the same active \ac{PE} set is used.
```
jdinan: shmem\_sync

content/shmem_sync.tex
```diff
@@ -0,0 +1,83 @@
...
+
+    If the \activeset does not change, \FUNC{shmem\_sync} can be called
+    repeatedly with the same \VAR{pSync} array.  No additional synchronization
+    beyond that implied by \FUNC{shmem\_sync} itself is necessary in this case.
+
+    The \FUNC{shmem\_sync\_all} routine can be used to portably ensure that
+    memory access operations observe remote updates in the order enforced by
+    initiator \acp{PE}, provided that the initiator PE ensures completion of remote
+    updates with a call to \FUNC{shmem\_quiet} prior to the call to
+    \FUNC{shmem\_sync\_all} routine.
```
jdinan: ^the

content/shmem_sync.tex
spotluri: Jim, "memory stores" can also include stores to memory pointed by address returned from shmem_ptr. This would mean sync complete stores to remote symmetric memory as well.  Is this the intent? From the examples, the intent seems to be that stores to local symmetric memory are complete.

jdinan: Yes, that's the intent.

spotluri: Sorry for repeating. I had two alternatives in my question, "complete stores to local and remote symmetric memory" or "complete stores to local symmetric memory". Is your yes for the latter?

jdinan: This function will replicate the store completion semantics of quiet, but omit the communication completion semantic.  As soon as we figure out how that should work.

jdinan:
[2017-05-02 -- Sync Draft.pdf](https://github.com/openshmem-org/specification/files/1029753/2017-05-02.--.Sync.Draft.pdf)

BryantLam:
I am okay (but also indifferent) with merging this proposal with #63. This PR has a better shot of being merged in first or at the same time as #63, so that might be one reason to leave them separate.

jdinan:
Special ballot: "Fix typos found during Aug. 2017 reading" cb16082

jdinan:
Closes #119

jdinan:
Ready to merge.  Can I get two +1s?

@manjugv Could you please review the merge of the new shmem_barrier_all text into shemem_sync_all in 6eca930?

manjugv:
@jdinan Looks good!

PR #65, opened 2017-04-14
Relaxed definition of SHMEM_MAX_NAME_LEN

BryantLam:
`SHMEM_MAX_NAME_LEN` now represents the maximum string length instead of the actual string length. The original definition has ambiguous implications with whether null/blank characters at the end of the relevant characters were counted or not.

Potentially a DocEdit, but this change technically relaxes the existing semantics even if the original semantics were sometimes not used in practice.

### DIFF
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -501,6 +501,9 @@ synchronization routines.
 \item Added the \FUNC{shmem\_calloc} function.
 \\ See Section \ref{subsec:shmem_calloc}.
 %
+\item Clarified description for \CONST{SHMEM\_MAX\_NAME\_LEN}.
+\\See Section \ref{subsec:library_constants}.
+%
 \end{itemize}
 
 \section{Version 1.3}
--- a/content/library_constants.tex
+++ b/content/library_constants.tex
@@ -120,7 +120,7 @@ Integer representing the minor version of \openshmem{} standard in use. \tabular
 \vtop{\hbox{\CorCppFor:} 
 \hbox{\hspace*{12mm} \const{SHMEM\_MAX\_NAME\_LEN}}}
 &
-Integer representing the length of vendor string. \tabularnewline
+Integer representing the maximum length of \const{SHMEM\_VENDOR\_STRING}. \tabularnewline
 \hline
 \vspace{3mm}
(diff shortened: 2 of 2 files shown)


jdinan:
I suggest we assign this to the front matter section committee.  If they are unanimous, we can merge it without going through the proposal process (it gets voted on later with all other such section edits).  If we go that route, there should be no need for a changelog entry.

Michael Raymond is the chair of the front matter section committee.  Other committee members: Swaroop, Nick, Tony Curtis, and Bryant.

H2-He-O2:
This looks fine to me.

jdinan:
Reading this again, I think this does merit a changelog entry.  Given that it's ready to read, I suggest we just follow the full proposal process.

tonycurtis:
Looks ok.

dkhaldi:
The definition of SHMEM_INFO_GET_NAME  (see 8.1.10) explains that null/blank characters at the end of the relevant characters are counted. 

In case you want to increase the scope of this ticket, I think the definition of SHMEM_VENDOR_STRING should appear before SHMEM_MAX_NAME_LEN in Section 6. Also, SHMEM_VENDOR_STRING appear only once in the spec: The spec does not explicitly say that SHMEM_INFO_GET_NAME returns 
 SHMEM_VENDOR_STRING

BryantLam:
> The definition of SHMEM_INFO_GET_NAME (see 8.1.10) explains that null/blank characters at the end of the relevant characters are counted.

Very true. That definition conflicts with the current definition for `SHMEM_MAX_NAME_LEN` (exact length). This proposed change is attempting to adopt the more relaxed definition as seen in `shmem_info_get_name`.

> In case you want to increase the scope of this ticket, I think the definition of SHMEM_VENDOR_STRING should appear before SHMEM_MAX_NAME_LEN in Section 6.

I like this idea. I'll DocEdit this change in a separate PR.

> Also, SHMEM_VENDOR_STRING appear only once in the spec: The spec does not explicitly say that SHMEM_INFO_GET_NAME returns SHMEM_VENDOR_STRING

It does now! 😀 See ratified #51.

dkhaldi:
:thumbsup:

jdinan:
Thanks for merging the upstream changes.

PR #67, opened 2017-04-14
Deprecated header directory mpp

BryantLam:
Essentially deprecated since OpenSHMEM 1.0, but OpenSHMEM 1.0 did not
*technically* list the mpp directory in Appendix A (Deprecated Items),
but rather as a Compatibility Note. This note disappeared in OpenSHMEM
1.1 and is now being restored for completeness.

Deprecation is assumed to have occured in OpenSHMEM 1.1.

### DIFF
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -334,6 +334,7 @@ specification.
     & \shortstack{\textbf{Last Version Supported}}
     & \textbf{Replaced By} \\
     \hline
+    Header Directory: \hyperref[subsec:dep_rationale:mpp]{\HEADER{mpp}} & 1.1 & Current & (none) \\ \hline
     \CorCpp: \hyperref[subsec:start_pes]{\FUNC{start\_pes}} & 1.2 & Current & \hyperref[subsec:shmem_init]{\FUNC{shmem\_init}} \\ \hline
     \Fortran: \hyperref[subsec:start_pes]{\FUNC{START\_PES}} & 1.2 & Current & \hyperref[subsec:shmem_init]{\FUNC{SHMEM\_INIT}} \\ \hline
     \CorCpp: \FUNC{\_my\_pe} & 1.2 & Current & \hyperref[subsec:shmem_my_pe]{\FUNC{shmem\_my\_pe}} \\ \hline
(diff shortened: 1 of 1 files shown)


content/backmatter.tex
```diff
@@ -334,6 +334,7 @@ \section{Overview}\label{subsec:dep_overview}
     & \shortstack{\textbf{Last Version Supported}}
     & \textbf{Replaced By} \\
     \hline
+    Header Directory: \hyperref[subsec:dep_rationale:mpp]{\HEADER{mpp}} & 1.1 & Current & (none) \\ \hline
```
jdinan: To be consistent with the other deprecations, should we list this as ```\CorCpp: \HEADER{mpp/shmem.h}``` and ```\Fortran: \HEADER{mpp/shmem.fh}```?

Also, what about the extensions headers?  Should we also list those here for completeness?

BryantLam: I went with "Header Directory" since it seemed more widely applicable and left room for interpretation of all headers (not just `shmem.h` and `shmem.fh`). It also made it easier to hyperref-link it. I don't mind either, but this seemed cleaner if less precise.

As for the extension headers, I found that to be ambiguous. The OpenSHMEM 1.0 compatibility note where the deprecation rationale was taken from states "these header files", but there are no references to extension headers in 1.0 and `shmemx.h` was not introduced until 1.1. It was assumed it was calling out only `shmem.h`. I kept the original semantics for "these header files" in the rationale, but thinking on it now, I should include `shmemx.h` into the rationale's code blocks.

BryantLam: Actually, I don't know about the extensions. It might be better to not include them in the mpp directory so as to force implementations that are using `<mpp/shmemx.h>` to side-grade to `<shmemx.h>`. The intent was backwards compatibility with SGI SHMEM, which did not have a `shmemx.h` header.

jdinan:
PDF with highlighting: [main_spec.pdf](https://github.com/openshmem-org/specification/files/964792/main_spec.pdf)

BryantLam:
Any suggested changes?

jdinan:
None that I recall.  Should be ready for voting at the next meeting.

PR #73, opened 2017-06-01
Removing 'we' from spec.

j-mike-culhane:
As of these edits, all instances of 'we', with one exception, have been removed from the document. The exception is located in the acknowledgements section on page iii and was left untouched because usage of pronouns is appropriate in this section of the document.


tonycurtis:
+1

nspark:
👍

manjugv:
:+1:

jdinan:
Thanks!

PR #82, opened 2017-06-21
Update OpenSHMEM implementations list on Annex-E

naveen-rn:
Annex:E has list of OpenSHMEM implementations. 

1. Adding Cray SHMEM as part of that OpenSHMEM implementations. 
From version 7.4.0, Cray SHMEM is OpenSHMEM-1.3 compliant .

2. Changing name of Portals-SHMEM to Sandia OpenSHMEM

### DIFF
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -301,7 +301,7 @@ interconnect, and to a very limited extent, Infiniband based clusters.
   \item SGI \openshmem
   \item University of Houston - \openshmem Reference Implementation
   \item Mellanox ScalableSHMEM
-  \item Portals-SHMEM
+  \item Sandia OpenSHMEM
   \item IBM OpenSHMEM
   \item Cray SHMEM
   \end{itemize}
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -303,6 +303,7 @@ interconnect, and to a very limited extent, Infiniband based clusters.
   \item Mellanox ScalableSHMEM
   \item Portals-SHMEM
   \item IBM OpenSHMEM
+  \item Cray SHMEM
   \end{itemize}
 \end{itemize}
 


naveen-rn:
Not sure, whether this can be considered as documentation change.
If need to pass some some particular test suite to be part of that 
OpenSHMEM implementation list, please let me know.

jdinan:
Nope, not a document edit, unfortunately.  If you're going to make a proposal for this change, could you also rename "Portals-SHMEM" to "Sandia OpenSHMEM"?

naveen-rn:
Updated Portals-SHMEM name.

jeffhammond:
Well, if we are trying to be complete, we should add:
* Paris OpenSHMEM: https://github.com/coti/POSH
* OSHMPI: https://github.com/jeffhammond/oshmpi

OSHMPI supports OpenSHMEM 1.2 as far as I know - I ran all the tests that @jdinan verified Sandia OpenSHMEM against when he declared 1.2 support, although OSHMPI does not support Fortran bindings.  While I'm not actively working on OSHMPI, it was the second implementation of 1.2 after Sandia OpenSHMEM, which should count for something :neckbeard:

jdinan:
Keeping this list current is going to be a challenge, and Jeff raises a good point -- what version of the specification must be supported for an implementation to appear in this list?  I think I prefer to update this history section so that it describes only what came before OpenSHMEM.

jeffhammond:
@jdinan We should require formal verification before implementations can be listed there 😜

jamesaross:
Is there a compliance test suite?
ARL OpenSHMEM for Epiphany: https://github.com/USArmyResearchLab/openshmem-epiphany

jeffhammond:
@jamesaross Various people are working on them but I don't know if there is an official one yet (people will surely correct me if I'm wrong).  I use the Sandia Open-SHMEM test bucket (https://github.com/Sandia-OpenSHMEM/SOS/tree/master/test), because those include debugged versions of UH/ORNL tests as well as some original ones.

swpoole:
The list of implementations @ [ http://openshmem.org/site/Links#impl ] was started a long time ago.
The goal of this is to try and be inclusive on just "implementations" of OpenSHMEM which are in the
public domain or are available for use. This list does not assume OpenSHMEM compliance. It is just a list.
There is a V&V Suite being developed. The plan all-along has been to have a V&V suite which does state compliance. There have been a large number of "test suites" from various implementors. These are being sorted and built and will be released in order to allow for any implementation of OpenSHMEM to ensure correctness. Once finalized and checked for correctness itself, we will run these test suites against proposed compliant implementations and then certify that the release is in fact VX.Y compliant. This has always been the plan. Currently LANL/Rice/SBU are working the overall V&V suite. They will most certainly be drawing from the work of others to ensure what they have is as complete as possible.

jdinan:
Indeed, I think this discussion does a nice job of capturing why we should not try to include a list like this in the specification document.  A further justification is that it is impossible in the 1.4 specification to list the implementations that are 1.4 compliant.  Unless we invent time travel.

jdinan:
Whoops, sorry for the label churn.  This has been fixed by the backmatter section committee.  @naveen-rn Can we close this PR?

jdinan:
Addressed by back matter committee.

PR #83, opened 2017-06-27
Edited usage of 'symmetric data objects' in spec.

j-mike-culhane:
While reading through the spec, I noticed that the term 'symmetric data objects' appeared in the overview for the OpenSHMEM routines on pages 1 and 2. However, the term itself is not defined until page 3 in the Memory Model section. The purpose of this file is to introduce the term 'Symmetric Data Objects' before the overview, and reword the Memory Model section so that coherence is retained and repetition is avoided.


swpoole:
I also assume terms like this will be included in the glossary?

j-mike-culhane:
That's the plan.

content/memory_model.tex
```diff
@@ -5,13 +5,13 @@
 \end{figure}      
 %
 An \openshmem program consists of data objects that are private to each \ac{PE}
-and data  objects that are remotely accessible by all \acp{PE}. Private data
+and data  objects that are remotely accessible by all \acp{PE}. Note that the latter 
+of these two data objects are symmetric data objects. Private data
```
jdinan: Instead of "Note that", how about a semicolon?

An \openshmem program consists of data objects that are private to each \ac{PE} and data  objects that are remotely accessible by all \acp{PE}; the latter of these two data objects are +referred to as+ \emph{symmetric data objects}.

content/memory_model.tex
```diff
@@ -5,13 +5,13 @@
...
 objects are stored in the local memory of each \ac{PE} and can only be accessed
 by the \ac{PE} itself; these data objects cannot be accessed by other \acp{PE}
 via \openshmem routines. Private data objects follow the memory model of
-\Cstd or \Fortran. Remotely accessible objects, however, can be accessed by
-remote \acp{PE} using \openshmem routines.  Remotely accessible data objects are
-called \emph{Symmetric Data Objects}.  Each symmetric data object has a
+\Cstd or \Fortran. Symmetric data objects, however, can be accessed by
```
jdinan: This is an abuse of "however".

content/programming_model_overview.tex
```diff
@@ -24,7 +24,7 @@
 PGAS languages, \openshmem relies on the user to use the library calls  to
 implement the correct semantics of its programming model.
 
-An overview of the \openshmem routines is described below:
+An overview of the \openshmem routines is described below. Remotely accessible data objects are referred to as \emph{Symmetric Data Objects} in this overview:
```
jdinan: Symmetric data objects should be italicized above when it is first introduced, not here.  Also, why is this term capitalized?

jdinan:
This one is not a document edit; it will need to either go through the proposal process or through the relevant chapter committee(s).

content/memory_model.tex
j-mike-culhane: Works for me. Change has been made.

content/memory_model.tex
j-mike-culhane: What if it came at the start of the sentence?

content/programming_model_overview.tex
j-mike-culhane: If these edits were to make it into the spec, this would be the first time the term 'symmetric data objects' is introduced/mentioned.

jdinan:
PR is stale.  Most of the original changes were already applied.

PR #87, opened 2017-07-20
fix apidefinition environment

boehms:
At the last f2f the API Notes none was brought up. I looked at how they can be removed. This should fix the apidescription environment and the usage of apidesctable.
Additionally the empty API Notes are removed.

### DIFF
--- a/content/atomics_intro.tex
+++ b/content/atomics_intro.tex
@@ -2,7 +2,7 @@ An \ac{AMO} is a one-sided communication mechanism that combines memory update
 operations with atomicity guarantees described in Section
 \ref{subsec:amo_guarantees}.  Similar to the \ac{RMA} routines, described in
 Section \ref{sec:rma}, the \acp{AMO} are performed only on symmetric objects.
-\openshmem{} defines the two types of \ac{AMO} routines:
+\openshmem defines the two types of \ac{AMO} routines:
 \begin{itemize}
 \item
 The \textit{fetching} routines return the original value of, and optionally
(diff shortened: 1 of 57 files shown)


manjugv:
This could be handled as document edit ? @jdinan @nspark What do you think ?

jdinan:
This seems reasonable to handle as a document edit.  I'll be back in the office next week and can do a detailed review then.  In the meantime, +2 from other committee members would be helpful.

swpoole:
+1

nspark:
@boehms You missed the removal of "Notes: None" in:
* `shmem_calloc`
* `shmem_atomic_{fetch_and, and, fetch_or, or, fetch_xor, xor}`
* `shmem_test`

Interestingly, I think these are all interfaces that I added.

boehms:
Those might have come in with the merge. I will fix them.

nspark:
:+1:

jdinan:
I'm seeing an indentation problem in the API description table, e.g. for SHMEM_ATOMIC_SWAP.  There's also a bad break that splits the table header and rows for SHMEM_ATOMIC_FETCH_INC.

[main_spec.pdf](https://github.com/openshmem-org/specification/files/1191882/main_spec.pdf)

boehms:
Making sure we don't have a page break in the table will need some reworking of the environments and the commands. It can be done, but will require some more work.

jdinan:
The internet suggests numerous ways to do this.  The simplest is to wrap it in a table float, which is unbreakable (with "here" placement).  Alternatively, there are solutions like this one: https://tex.stackexchange.com/questions/94699/absolutely-definitely-preventing-page-break

boehms:
The problem is, that is is not actually one table. Each \apitablerow is its own tabular. And to make things harder, they are in a quote environment (\apidesctable). If we where to use a float, it would break out of the quote environment.

manjugv:
@boehms @jdinan do you have an agreement here ?

jdinan:
Well ... now we have other formatting weirdness.  Is there some flexible vspace we can put in front of the table to fix this?

screen shot 2017-08-27 at 10 07 29 pm

boehms:
Are you referring to the sentence on the bottom on page 33?

I am not aware of a flexible vspace. I think the formatting is something we should take care of, when the content of the document is close to being finalized. The table might move with new content.

jdinan:
Could you take a look at \longtable (ref #123)?  That seems like it could provide a good solution.

jdinan:
Also, FWIW, if you separate the API table changes, we can merge the rest of this PR while we sort out the table formatting.

boehms:
@jdinan How about this?

nspark:
Replaced by #330 (though it doesn't address table and whitespace changes)

Issue #68, opened 2017-04-28
Update shmem_test example


jdinan:
Update the ```shmem_test``` example added by #32 to add main function and make the example complete.

manjugv:
@jdinan Should we close this issue ?

nspark:
@manjugv No, I still need to take care of this. (I gave @shamisp my word after all.)

PR #89, opened 2017-07-26
Update shmem_test() example to be runnable

nspark:
PR #32 was ratified with the expectation that the `shmem_test()` example would be updated to a runnable example and not merely be a function demonstrating the use of `shmem_test()` (see Issue #68). This PR should close out #68.

Here's looking at you, @shamisp.


jdinan:
Looks good to me.  Can I get a +1 to merge?

PR #90, opened 2017-07-26
 Query Available Symmetric Heap Memory

manjugv:
This PR introduces the routine to query the available memory on the symmetric heap.

### DIFF
--- a/content/shmem_symmetric_available.tex
+++ b/content/shmem_symmetric_available.tex
@@ -1,0 +1,33 @@
+\apisummary{
+Returns the memory available on the symmetric heap in bytes.}
+
+
+\begin{apidefinition}
+
+\begin{Csynopsis}
+void shmem_symmetric_available(size_t *available_size);
+\end{Csynopsis}
+
+\begin{apiarguments}
+\apiargument{OUT}{available\_size}{The symmetric heap memory available in bytes.}
+\end{apiarguments}
+
+
+\apidescription{
+The \FUNC{shmem\_symmetric\_available} call returns the memory available on the symmetric heap in bytes.
+The available memory is the unallocated memory on the symmetric heap, which could be allocated by using memory allocation routines such as \FUNC{shmem\_malloc} or \FUNC{shmem\_calloc}. 
+This call is a non-collective call and does not require more than one \ac{PE} to call the routine. 
+}
+
+\apireturnvalues{    
+	None
+}
+
+\apinotes{
+    None
+}
+
+\end{apidefinition}
+
+ 
+  
--- a/main_spec.tex
+++ b/main_spec.tex
@@ -103,6 +103,8 @@ symmetric data objects in the symmetric heap, in \Cstd and \Fortran.
 \subsubsection{\textbf{SHPDEALLOC}}\label{subsec:shpdealloc}
 \input{content/shpdealloc.tex}
 
+\subsubsection{\textbf{SHMEM\_SYMMETRIC\_AVAILABLE}}\label{subsec:shmem_symmetric_available}
+\input{content/shmem_symmetric_available.tex}
 
 
 


content/shmem_symmetric_available.tex
```diff
@@ -0,0 +1,33 @@
+\apisummary{
+Returns the memory available on the symmetric heap in bytes.}
+
+
+\begin{apidefinition}
+
+\begin{Csynopsis}
+void shmem_symmetric_available(size_t *available_size);
```
nspark: Does this interface use a "return"-by-pointer-argument style because of potential future error-handling return codes? It's just my preference, but I'd rather do that as little as possible in the API. (I just find it ugly.) Since we are returning a `size_t` (that is, `unsigned long` on most platforms), we could use `SIZE_MAX` as the indicator for error.

Also, `shmem_symmetric_available` sounds awkward to me. ("Symmetric" is an adjective; I want a noun that indicates what exactly is available.) What about `shmem_symmetric_heap_available`?

content/shmem_symmetric_available.tex
```diff
@@ -0,0 +1,33 @@
...
+\end{Csynopsis}
+
+\begin{apiarguments}
+\apiargument{OUT}{available\_size}{The symmetric heap memory available in bytes.}
+\end{apiarguments}
+
+
+\apidescription{
+The \FUNC{shmem\_symmetric\_available} call returns the memory available on the symmetric heap in bytes.
+The available memory is the unallocated memory on the symmetric heap, which could be allocated by using memory allocation routines such as \FUNC{shmem\_malloc} or \FUNC{shmem\_calloc}.
```
nspark: I'm not an expert in memory allocators and maybe this is something for the API notes, but I think there's a difference in the total amount of unallocated memory and the (largest) amount of memory that could be allocated by a single call to `shmem_malloc`.

Allocations and deallocations may cause some fragmentation of the symmetric heap and the size returned by this interface may be larger in cases than the largest allocation that `shmem_malloc` could actually provide.

content/shmem_symmetric_available.tex
MattBBaker: If we have to change the name I think ```shmem_heap_available``` would be better. The ```shmem_symmetric_heap_available``` seems too long.

content/shmem_symmetric_available.tex
bcernohous: I'm not sure a simple 'memory available in bytes' is helpful.  For example the existing dmapp symmetric heap stats api/structure:

The dmapp_sheap_stats_t structure contains symmetric heap usage information and is returned by dmapp_sheap_stats.

typedef struct {
uint64_t size;           
uint64_t usage;    
uint64_t hwm;  
uint64_t biggest_used;  
uint64_t biggest_free;  
uint64_t used_blks;  
uint64_t free_blks; 
uint64_t allocs;  
uint64_t frees;  
uint64_t fails; 
uint64_t reserved[8]; 
  } dmapp_sheap_stats_t;

5.2.3.2 Members
size |   | Total heap size in bytes.
usage |   | Current usage in bytes.
hwm |   | High water mark in bytes
biggest_used |   | Size of largest allocator block in use
biggest_free |   | Size of largest free allocator block.
used_blks |   | Number of allocator blocks in use.
free_blks |   | Number of free allocator blocks.
allocs |   | Running total number of alloc calls.
frees |   | Running total number of free calls.
fails |   | Running total number of failed alloc calls.
reserved[8] |   | Reserved for future use.

content/shmem_symmetric_available.tex
manjugv: I'm definitely OK with an alternate name. :)

jamesaross:
The way this is written, I couldn't tell you what the code below would print or even if it would allocate ptr3.  If the Unix `brk`/`sbrk` routines were used in place of `shmem_malloc` and `shmem_free`, the behavior would be more predictable, I believe.

```c
size_t available_size;
shmem_symmetric_available(&available_size);
printf("available_size = %d\n", available_size);

void* ptr1 = shmem_malloc(1024);

shmem_symmetric_available(&available_size);
printf("available_size = %d\n", available_size);

void* ptr2 = shmem_malloc(1024);

shmem_symmetric_available(&available_size);
printf("available_size = %d\n", available_size);

shmem_free(ptr1);

shmem_symmetric_available(&available_size);
printf("available_size = %d\n", available_size);

void* ptr3 = shmem_malloc(available_size);

shmem_symmetric_available(&available_size);
printf("available_size = %d\n", available_size);
```

jeffhammond:
Querying max available segment makes sense. Sum of defenders does not. It's
a heap not a stack and thus can be fragmented. Can have 1G as 1M 1K
segments free and allocation of 2K will fail.
-- 
Jeff Hammond
jeff.science@gmail.com
http://jeffhammond.github.io/

swpoole:
My question is along the same lines as @jeffhammond . If I do not have homogeneous nodes, how does it fail when I check the amount, or assume the same amount is across all nodes. I like the idea, though. Just wondering if we have looked at all of the potential failure modes.

jeffhammond:
Is heterogeneity in the size of the heap supported? I'm not sure how it
could be used, given the requirement of symmetric allocation.

On the other hand, if heap manager is nondeterministic, the max available
fragment size may not be the same across all nodes. In this case, the query
must be collective.
-- 
Jeff Hammond
jeff.science@gmail.com
http://jeffhammond.github.io/

swpoole:
I will have to check on some of our development clusters. I just wanted to raise a potential issue.
Thanks...

jdinan:
I'm concerned that the proposed standardizes certain assumptions about how OpenSHMEM symmetric objects are implemented.  A pre-allocated symmetric heap is certainly the most common and efficient implementation that we all use.  However, nothing in the API requires this approach.  It's possible to allocate symmetric objects on-demand and grow/shrink the symmetric space as the application needs it.  In that case, an implementation might ignore the ```SHMEM_SYMMETRIC_SIZE``` environment variable (and wouldn't that be nice for users who struggle with this knob, e.g. because they have to predict the memory needs for a given data set or because their application has phases which alternate between using memory for local and symmetric objects).

Imagining an implementation that does not use a statically mapped symmetric heap (e.g. it may also over-commit memory), how should I answer this query?  Which Linux API will provide an answer that's meaningful to the user?

jeffhammond:
What is the output of the following program when the argument is e.g. `1024` and `4096`?  In particular, what is the result of the third call to `shmem_symmetric_available()` and is that information useful to SHMEM users?  How does the behavior of this program depend on the implementation of the heap manager?
```c
#include <stdio.h>
#include <stdlib.h>

#include <shmem.h>

int main(int argc, char* argv[])
{
    int chunk = (argc>1) ? atoi(chunk) : 8;
    setenv("SHMEM_SYMMETRIC_SIZE","1073741824",1);
    size_t nchunk = ((size_t)1073741824)/((size_t)chunk);

    shmem_init();

    void** addrs  = malloc(nchunk*sizeof(void*));
    printf("available=%zu\n", shmem_symmetric_available() );

    for (size_t i=0; i<nchunk; i++) {
        printf("allocating chunk %zu\n", i);
        addrs[i] = shmem_malloc(chunk);
    }
    printf("available=%zu\n", shmem_symmetric_available() );

    for (size_t i=0; i<nchunk; i+=2) {
        shmem_free(addrs[i]);
    }
    printf("available=%zu\n", shmem_symmetric_available() );

    for (size_t i=1; i<nchunk; i+=2) {
        shmem_free(addrs[i]);
    }
    printf("available=%zu\n", shmem_symmetric_available() );

    free(addrs);

    shmem_finalize();

    return 0;
}
```

manjugv:
I think the proposal is misunderstood (may be there are not enough details here). 

With the current specification, there is no easy way to know if my next shmem_malloc (collective call) will succeed/fail. The main motivation of this proposal is to provide a query call (local) which will provide such information.

I guess, the confusion arises from the fact that we have not clearly defined what SHMEM_SYMMETRIC_SIZE and available memory mean to the User and implementation. 

AFAIU, when we set some value in “SHMEM_SYMMETRIC_SIZE”, the library should honor this allocation. Otherwise, the library initialization should fail; since the library cannot allocate the required resources. After successful initialization,  the memory can be allocated either by a single call or multiple calls, and library should honor this. This interpretation led me to propose the available memory rather than largest chunk available.

The other interpretation that shmem_malloc can fail if the library cannot allocate the memory chunk, even though the total memory available might be valid. I’m not sure. May be this is what most implementations do.

content/shmem_symmetric_available.tex
manjugv: @bcernohous How do you keep this information updated ? Do you have some kind of distributed memory allocator ?

manjugv: @bcernohous I disagree. There are many scenarios where we want to allocate all of available symmetric memory heap memory. One scenario where we encountered this - using the symmetric heap as a cache for I/O workloads. Here, we wanted to use  all of available 
heap as a cache. If you want to do something like this, with the current specification, you have do the bookkeeping in the application to figure out the available memory. 

I’m not disputing that more details are useful. If there is clear motivation, we can add it.

manjugv:
> Querying max available segment makes sense. Sum of segments does not. It's a heap, not a stack, and thus can be fragmented. One can have 1GiB free in the form of 2^20 1KiB free segments and allocation of 2KiB will fail.

If we want to return the largest segment available, the value of this call would be much less because the query call has to co-ordinate with all PEs to figure the largest segment available.

manjugv:
> I'm concerned that the proposal standardizes certain assumptions about how OpenSHMEM symmetric objects are implemented. A pre-allocated symmetric heap is certainly the most common and efficient implementation that we all use. However, nothing in the API requires this approach. It's possible to allocate symmetric objects on-demand and grow/shrink the symmetric space as the application needs it. In that case, an implementation might ignore the SHMEM_SYMMETRIC_SIZE environment variable (and wouldn't that be nice for users who struggle with this knob, e.g. because they have to predict the memory needs for a given data set or because their application has phases which alternate between using memory for local and symmetric objects).
> 
> Imagining an implementation that does not use a statically mapped symmetric heap (e.g. it may also over-commit memory), how should I answer this query? Which Linux API will provide an answer that's meaningful to the user?

@jdinan I’m not sure, if it is does more than what `shmem_malloc` already does. 

Say we return “available_size” from a call to `shmem_symmetric_available()`. If the query call is followed by shmem_malloc, the shmem_malloc will always succeed on “available_size” .  This information might be already available in the library, and this call enables the User to query it. That is motivation behind this proposal.

jdinan:
> Say we return “available_size” from a call to shmem_symmetric_available(). If the query call is followed by shmem_malloc, the shmem_malloc will always succeed on “available_size” . This information might be already available in the library, and this call enables the User to query it. That is motivation behind this proposal.

@manjugv The problem is that the runtime might not actually know whether a call to ```shmem_malloc``` will succeed until it attempts to allocate memory.  OpenSHMEM runtimes that set aside physical memory during initialization can answer this query reliably (with fragmentation caveats).  Any other method of managing symmetric objects will likely be unable to answer such a query.

To illustrate the latter case, imagine that I have an implementation that allocates a buffer from the OS at every call to ```shmem_malloc``` (addresses need not be symmetric across PEs).  Because allocations are global, every PE knows that this is the _N_ th call to ```shmem_malloc```.  Every PE then adds the buffer's address range and _N_ to a lookup structure that will be used to look up _N_ for a given address (e.g. when performing a put operation).  On a Portals network, I can then append the buffer to a matching list* with tag _N_.  The only information my peers need to access this buffer is _N_ and the offset from the start of the buffer.  Thus, on such a system, a static symmetric heap is not required to for an efficient SHMEM implementation.

_*_ Note: More optimal implementation methods are possible using the nonmatching Portals interface; the behavior is similar.  I tried to choose the method that is simplest for this example.

jeffhammond:
Prove to me that you can implement `shmem_symmetric_available` better than this, given that this function must be collective and should never lie on the critical path.
```c
int notshmem_symmetric_available(size_t bytes)
{
  void * ptr = shmem_malloc(bytes);
  if (ptr==NULL) return 0;
  else shmem_free(ptr);
  return 1;
}
```

manjugv:
> To illustrate the latter case, imagine that I have an implementation that allocates a buffer from the OS at every call to shmem_malloc (addresses need not be symmetric across PEs). Because allocations are global, every PE knows that this is the N th call to shmem_malloc. Every PE then adds the buffer's address range and N to a lookup structure that will be used to look up N for a given address (e.g. when performing a put operation). On a Portals network, I can then append the buffer to a matching list* with tag N. The only information my peers need to access this buffer is N and the offset from the start of the buffer. Thus, on such a system, a static symmetric heap is not required to for an efficient SHMEM implementation.

@jdinan I do agree that for such an implementation, it might require a collective. I would like to understand from Cray, HPE, and OpenMPI’s OpenSHMEM, if they can provide this information without a collective. Do you do on-demand allocation of symmetric heap in SOS ?

manjugv:
> Prove to me that you can implement shmem_symmetric_available better than this, given that this function must be collective and should never lie on the critical path.

Imagine an implementation where the physical memory (required to honor `SHMEM_SYMMETRIC_SIZE`) is set aside during initialization. In all likelihood, the library is maintaining the information (total available, largest chunks ..), why do I need a collective in that case. This is not an imaginary implementation, but a quite realistic one.

For other implementations where a collective might be required, it is quite likely that such information is maintained by the library. In such cases, the library can just return the information from the library instead of User invoking a `shmem_malloc`. 

I'm I missing something.

manjugv:
I guess the PR is not yet ready for reading. I will do "unofficial reading"/discussion at the F2F rather than do an official reading.

jdinan:
@naveen-rn Any thoughts on how this would interact with Cray's growable symmetric heap mode?

jeffhammond:
@manjugv What if my heap allocator is not strictly deterministic and the fragmentation pattern of the heap is not symmetric?  Can you audit the common heap allocator implementations and provide a summary of their behavior w.r.t. determinism?

jeffhammond:
@manjugv I want to reiterate that `available_size` is utterly meaningless because of heap fragmentation.  If you assume the symmetric heap is allocated during library initialization, the only thing that a heap manager can provide that is directly correlated with successful allocation is the size of the largest free segment.

manjugv:
OpenSHMEM 2017 F2F Notes : Closing this PR as there was no interest ...

PR #96, opened 2017-08-02
Punctuation edits for pages 30 to 60.

j-mike-culhane:
As stated in the title, this file consists of punctuation edits from pages 30 to 60.

### DIFF
--- a/content/atomics_intro.tex
+++ b/content/atomics_intro.tex
@@ -1,5 +1,5 @@
 An \ac{AMO} is a one-sided communication mechanism that combines memory update
-operations with atomicity guarantees described in Section
+operations with atomicity guarantees, described in Section
 \ref{subsec:amo_guarantees}.  Similar to the \ac{RMA} routines, described in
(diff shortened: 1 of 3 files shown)


content/shmem_barrier.tex
```diff
@@ -38,7 +38,7 @@
 \apidescription{
     \FUNC{shmem\_barrier} is a collective synchronization routine over an
     \activeset.  Control returns from \FUNC{shmem\_barrier} after all \acp{PE} in
-    the \activeset (specified by \VAR{PE\_start}, \VAR{logPE\_stride}, and
+    the \activeset(specified by \VAR{PE\_start}, \VAR{logPE\_stride}, and
```
jdinan: Is this fixing a space issue?  If so, the correct fix is to the ```\activeset``` macro, rather than removal of the trailing space.  We can't apply this same fix in instances where the macro is followed by another word.

jdinan: Looking at the implementation of the ```\activeset``` macro, it already contains an ```\xspace```, so this change should not be necessary (i.e. it should have no effect).

PR #105, opened 2017-08-17
Consistency edits for the term: active set.

j-mike-culhane:
The purpose of this file is to change the formatting of the term ‘active set,’ **and one instance of the term ‘collective routines,’** so that they are consistent with the way other important terms appear in the spec. The following is an explanation justifying the edit:

When an important term is introduced in the spec it is written in italics and every first letter is capitalized. Symmetric data objects, for example, reads as _Symmetric Data Objects_ when first introduced. Every subsequent time an important term appears, all letters are written in the lower case and the term is not italicized. The term ‘active set,’ when it first appears, reads like so: _Active set_. Collective routines, when it first appears, also follows this formatting. Almost every subsequent time ‘active set’ appears in the spec the term is written this way. The a is capitalized and the term itself is written in italics. No other important terms follow this pattern.

### DIFF
--- a/content/shmem_alltoall.tex
+++ b/content/shmem_alltoall.tex
@@ -1,6 +1,6 @@
 \apisummary{
     shmem\_alltoall is a collective routine where each \ac{PE} exchanges a fixed amount of data with all other \acp{PE} in the
-    \activeset.
+    active set.
 }
 
 \begin{apidefinition}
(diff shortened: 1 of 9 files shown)


content/shmem_alltoall.tex
```diff
@@ -21,20 +21,20 @@
 
 \apiargument{OUT}{dest}{A symmetric data object large enough to receive
     the combined total of \VAR{nelems} elements from each \ac{PE} in the
-    \activeset.}
+    active set.}
 \apiargument{IN}{source}{A symmetric data object that contains \VAR{nelems}
-    elements of data for each \ac{PE} in the \activeset{}, ordered according to
+    elements of data for each \ac{PE} in the active set{}, ordered according to
```
jdinan: You have extra ```{}``` left over after removing the ```\activeset{}``` macro.

utils/defs.tex
```diff
@@ -63,7 +63,7 @@
 \newcommand{\reduction}{\textit{Reduction Operations}}
 \newcommand{\alltoall}{\FUNC{SHMEM\_ALLTOALL}}
 \newcommand{\alltoalls}{\FUNC{SHMEM\_ALLTOALLS}}
-\newcommand{\activeset}{\textit{Active~set}\xspace} % why here and not others?
```
jdinan: Do we still need this macro?

content/shmem_alltoall.tex
j-mike-culhane: @jdinan Brackets have been removed in most recent commit.

manjugv:
@j-mike-culhane can you please add a card with collectives section committee ? We would have several PRs that might conflict.

jdinan:
@manjugv Do you want to pull these changes into the collectives section PR or merge it now as a DocEdit?  Both options are workable; let me know which you prefer.

manjugv:
Please go ahead and merge, if it is ready. I will pull those changes into the collectives and add a card. Thanks.

jdinan:
@j-mike-culhane Please review/approve 4367f20.

All, this should be ready to merge.  Can I get a +1 from @j-mike-culhane and from another committee member?

j-mike-culhane:
👍

PR #128, opened 2017-08-31
Typo fix in sync section ("no PE my"->"no PE shall").

min-si:


min-si:
Close it as already merged into section/sync PR.

PR #129, opened 2017-08-31
Use consistent param name for wait/test routines.

min-si:


min-si:
Close it as already merged into section/sync PR.

min-si:
Close it as already merged into section/sync PR.

PR #139, opened 2017-09-01
Simplify and update the makefile

jdinan:
The existing makefile doesn't run pdflatex enough times:

    LaTeX Warning: Label(s) may have changed. Rerun to get
    cross-references right.

Update the makefile to fix the above (run standard 3 times), improve
build dependencies, and remove bit rotted DVI/PS build.


jdinan:
Ready for review.

BryantLam:
Have you considered using `latexmk`? I typically use that to "run pdflatex enough times."

jdinan:
Latexmk is great.  I wasn't sure if others who build the spec would have it installed, but also we pass ```-halt-on-error``` directly to pdflatex in Travis CI testing and I'm not sure whether we can get the same error behavior from ```latexmk```.

jdinan:
@manjugv ORNL touched the Makefile last.  What do you think of these changes?

manjugv:
@jdinan 👍

PR #141, opened 2017-09-02
Glossary for the spec.

j-mike-culhane:
This PR contains a rough draft for a potential glossary. This PR is meant to serve as a point of reference regarding how a glossary would appear in the spec. Please view and comment on whichever PR you think works better.

Note: I would like to stress that the 'Index for the spec' and 'Glossary for the spec' PRs are meant to serve as points of reference and, as they are currently, would require further editing before being added to the spec.

### DIFF
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -877,6 +877,120 @@ collectives.
 
 
 
+\chapter{Glossary}\label{sec:Glossary}
+
+$\;$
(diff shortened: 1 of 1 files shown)


BryantLam:
Closed. Superceded by https://github.com/BryantLam/openshmem-specification/pull/22 (@nspark).

PR #173, opened 2017-10-19
Fix typo in function name

jdinan:
Fix a couple "shmemx" function name typos.  These are left over from when I ran this example on the SOS prototype implementation of contexts, which used the "shmemx_" prefix instead of "shmem_" for the new routines.


jdinan:
I'd like to merge this as a document edit for inclusion in 1.4.  Can I get two +1's to merge?

PR #200, opened 2018-02-05
Add support for multipliers in SHMEM_SYMMETRIC_SIZE env var

### LINKED ISSUE #199: SHMEM_SYMMETRIC_SIZE Argument
## Issue

SHMEM_SYMMETRIC_SIZE does not adequately specify the format for its argument.

## Proposed Solution

We should support floating point numbers with the standard size suffixes of K, M, G, and T corresponding to powers of two bytes, e.g. SHMEM_SYMMETRIC_SIZE=1.5G requests 1.5 * 2^30 bytes.

An unqualified integer, e.g. SHMEM_SYMMETRIC_SIZE=1024, corresponds to a size in bytes and may not contain a decimal point.

## Notes from Redmine Discussion [Ticket 206](http://www.openshmem.org/redmine/issues/206)

Cray and SGI SHMEM both support a floating point number with K, M, G, and T suffix.

Sandia OpenSHMEM support an integer number with K, M, G, and T suffix.

naveen-rn:
Closes #199

### DIFF
--- a/content/environment_variables.tex
+++ b/content/environment_variables.tex
@@ -47,9 +47,11 @@ deprecation rationale for more details.
       equivalent to the integer value 3250586. Only one multiplier is
       recognized, so \enquote{20kk} will not produce the same desired result as
       \enquote{20m}. Usage of string \enquote{.5m} will yield the same result as
-      the string \enquote{0.5m}. Any unrecognized suffix is an error, generates
-      an error report, and either returns from shmem_init_thread or otherwise
-      causes program termination.
+      the string \enquote{0.5m}.
+
+      An invalid value for {SHMEM\_SYMMETRIC\_SIZE} is an error, which the
+      OpenSHMEM library shall report by either returning a nonzero value from
+      \FUNC{shmem\_init\_thread} or causing program termination.
     \tabularnewline\hline
 %%
 \EnvVarDecl{SHMEM\_DEBUG}
--- a/content/environment_variables.tex
+++ b/content/environment_variables.tex
@@ -33,7 +33,7 @@ deprecation rationale for more details.
     & Non-negative integer
     & Controls the size of the symmetric heap per PE. The value
       set in this environment variable is interpreted as number
-      of bytes, unless the number is followed by a char that
+      of bytes, unless the number is followed by a character that
       acts as a multiplier, where:
       \begin{itemize}
         \item g or G multiplies by 2**30 (gigabytes)
--- a/content/environment_variables.tex
+++ b/content/environment_variables.tex
@@ -34,8 +34,8 @@ deprecation rationale for more details.
     character suffix
     & The value set in this environment variable is interpreted as an
     implementation-defined size (in bytes) at least as large as the integer
-    ceiling of the product of the numeric prefix and the scaling factor per PE.
-    The allowed character suffix for the scaling factor are as follows:
+    ceiling of the product of the numeric prefix and the scaling factor per
+    \ac{PE}. The allowed character suffix for the scaling factor are as follows:
       \begin{itemize}
         \item k or K multiplies by \(2^{10}\)  (kibibytes)
         \item m or M multiplies by \(2^{20}\)  (mebibytes)
@@ -50,7 +50,7 @@ deprecation rationale for more details.
       the string \enquote{0.5m}.
 
       An invalid value for {SHMEM\_SYMMETRIC\_SIZE} is an error, which the
-      OpenSHMEM library shall report by either returning a nonzero value from
+      \openshmem library shall report by either returning a nonzero value from
       \FUNC{shmem\_init\_thread} or causing program termination.
     \tabularnewline\hline
 %%
--- a/content/environment_variables.tex
+++ b/content/environment_variables.tex
@@ -32,10 +32,10 @@ deprecation rationale for more details.
 \EnvVarDecl{SHMEM\_SYMMETRIC\_SIZE}
     & Non-negative integer or floating point value with an optional character
     suffix
-    & The value set in this environment variable is interpreted as an
-    implementation-defined size (in bytes) at least as large as the integer
-    ceiling of the product of the numeric prefix and the scaling factor per
-    \ac{PE}. The allowed character suffixes for the scaling factor are as
+    & Specifies the size (in bytes) of the symmetric heap memory per \ac{PE}.
+    The resulting size is implementation-defined and must be least as large as
+    the integer ceiling of the product of the numeric prefix and the scaling
+    factor. The allowed character suffixes for the scaling factor are as
     follows:
       \begin{itemize}
         \item k or K multiplies by \(2^{10}\)  (kibibytes)
--- a/content/environment_variables.tex
+++ b/content/environment_variables.tex
@@ -30,12 +30,13 @@ deprecation rationale for more details.
     \tabularnewline\hline
 %%
 \EnvVarDecl{SHMEM\_SYMMETRIC\_SIZE}
-    & Non-negative decimal value (integer or floating point) with an optional
-    character suffix
+    & Non-negative integer or floating point value with an optional character
+    suffix
     & The value set in this environment variable is interpreted as an
     implementation-defined size (in bytes) at least as large as the integer
     ceiling of the product of the numeric prefix and the scaling factor per
-    \ac{PE}. The allowed character suffix for the scaling factor are as follows:
+    \ac{PE}. The allowed character suffixes for the scaling factor are as
+    follows:
       \begin{itemize}
         \item k or K multiplies by \(2^{10}\)  (kibibytes)
         \item m or M multiplies by \(2^{20}\)  (mebibytes)
@@ -43,11 +44,11 @@ deprecation rationale for more details.
         \item t or T multiplies by \(2^{40}\)  (tebibytes)
       \end{itemize}
       For example, string \enquote{20m} is equivalent to the integer value
-      20971520, or 20 megabytes. Similarly the string \enquote{3.1m} is
+      20971520, or 20 mebibytes. Similarly the string \enquote{3.1m} is
       equivalent to the integer value 3250586. Only one multiplier is
-      recognized, so \enquote{20kk} will not produce the same desired result as
-      \enquote{20m}. Usage of string \enquote{.5m} will yield the same result as
-      the string \enquote{0.5m}.
+      recognized, so \enquote{20kk} will not produce the same result as
+      \enquote{20m}. Usage of string \enquote{.5m} will yield the same result
+      as the string \enquote{0.5m}.
 
       An invalid value for {SHMEM\_SYMMETRIC\_SIZE} is an error, which the
       \openshmem library shall report by either returning a nonzero value from
--- a/content/environment_variables.tex
+++ b/content/environment_variables.tex
@@ -14,7 +14,7 @@ deprecation rationale for more details.
 
 \medskip{}
 
-\begin{longtable}{|p{0.275\textwidth}|p{0.130\textwidth}|p{0.50\textwidth}|}
+\begin{longtable}{|p{0.260\textwidth}|p{0.145\textwidth}|p{0.50\textwidth}|}
 \hline
 \textbf{Variable} & \textbf{Value} & \textbf{Description}
 \tabularnewline\hline
@@ -46,8 +46,9 @@ deprecation rationale for more details.
       20971520, or 20 megabytes. Similarly the string \enquote{3.1m} is
       equivalent to the integer value 3250586. Only one multiplier is
       recognized, so \enquote{20kk} will not produce the same desired result as
-      \enquote{20m}. Any unrecognized suffix is treated as an error and results
-      in program termination with an error report.
+      \enquote{20m}. Usage of string \enquote{.5m} will yield the same result as
+      the string \enquote{0.5m}. Any unrecognized suffix is treated as an error
+      and results in program termination with an error report.
     \tabularnewline\hline
 %%
 \EnvVarDecl{SHMEM\_DEBUG}
--- a/content/environment_variables.tex
+++ b/content/environment_variables.tex
@@ -43,8 +43,8 @@ deprecation rationale for more details.
         \item t or T multiplies by \(2^{40}\)  (terabytes)
       \end{itemize}
       For example, the string 20m is equivalent to the integer value 20971520,
-      or 20 megabytes. Similarly the string 3.1g is equivalent to the integer
-      value 3250585. Only one multiplier is recognized, so 20kk
+      or 20 megabytes. Similarly the string 3.1m is equivalent to the integer
+      value 3250586. Only one multiplier is recognized, so 20kk
       will not produce the same desired result as 20m. Any unrecognized
       suffix is treated as an error and results in program termination with an
       error report.
--- a/content/environment_variables.tex
+++ b/content/environment_variables.tex
@@ -42,12 +42,12 @@ deprecation rationale for more details.
         \item g or G multiplies by \(2^{30}\)  (gigabytes)
         \item t or T multiplies by \(2^{40}\)  (terabytes)
       \end{itemize}
-      For example, the string 20m is equivalent to the integer value 20971520,
-      or 20 megabytes. Similarly the string 3.1m is equivalent to the integer
-      value 3250586. Only one multiplier is recognized, so 20kk
-      will not produce the same desired result as 20m. Any unrecognized
-      suffix is treated as an error and results in program termination with an
-      error report.
+      For example, string \enquote{20m} is equivalent to the integer value
+      20971520, or 20 megabytes. Similarly the string \enquote{3.1m} is
+      equivalent to the integer value 3250586. Only one multiplier is
+      recognized, so \enquote{20kk} will not produce the same desired result as
+      \enquote{20m}. Any unrecognized suffix is treated as an error and results
+      in program termination with an error report.
     \tabularnewline\hline
 %%
 \EnvVarDecl{SHMEM\_DEBUG}
--- a/content/environment_variables.tex
+++ b/content/environment_variables.tex
@@ -47,8 +47,9 @@ deprecation rationale for more details.
       equivalent to the integer value 3250586. Only one multiplier is
       recognized, so \enquote{20kk} will not produce the same desired result as
       \enquote{20m}. Usage of string \enquote{.5m} will yield the same result as
-      the string \enquote{0.5m}. Any unrecognized suffix is treated as an error
-      and results in program termination with an error report.
+      the string \enquote{0.5m}. Any unrecognized suffix is an error, generates
+      an error report, and either returns from shmem_init_thread or otherwise
+      causes program termination.
     \tabularnewline\hline
 %%
 \EnvVarDecl{SHMEM\_DEBUG}
--- a/content/environment_variables.tex
+++ b/content/environment_variables.tex
@@ -30,7 +30,7 @@ deprecation rationale for more details.
     \tabularnewline\hline
 %%
 \EnvVarDecl{SHMEM\_SYMMETRIC\_SIZE}
-    & Non-negative integer
+    & Non-negative integer with optional character suffix
     & Controls the size of the symmetric heap per \ac{PE}. The value
       set in this environment variable is interpreted as number
       of bytes, unless the number is followed by a single
@@ -39,11 +39,14 @@ deprecation rationale for more details.
         \item k or K multiplies by \(2^{10}\)  (kilobytes)
         \item m or M multiplies by \(2^{20}\)  (megabytes)
         \item g or G multiplies by \(2^{30}\)  (gigabytes)
+        \item t or T multiplies by \(2^{40}\)  (tebibytes)
       \end{itemize}
       For example, the string 20m is equivalent to the integer value 20971520,
       or 20 megabytes. Only one multiplier is recognized, so 20kk
       will not produce the same value as 20m, nor will invalid
-      strings such as 20MB produce the desired result.
+      strings such as 20MB produce the desired result. Any unrecognized
+      suffix is treated as an error and results in program termination with an
+      error report.
     \tabularnewline\hline
 %%
 \EnvVarDecl{SHMEM\_DEBUG}
--- a/content/environment_variables.tex
+++ b/content/environment_variables.tex
@@ -14,7 +14,7 @@ deprecation rationale for more details.
 
 \medskip{}
 
-\begin{longtable}{|p{0.275\textwidth}|p{0.125\textwidth}|p{0.5\textwidth}|}
+\begin{longtable}{|p{0.275\textwidth}|p{0.130\textwidth}|p{0.50\textwidth}|}
 \hline
 \textbf{Variable} & \textbf{Value} & \textbf{Description}
 \tabularnewline\hline
@@ -30,8 +30,8 @@ deprecation rationale for more details.
     \tabularnewline\hline
 %%
 \EnvVarDecl{SHMEM\_SYMMETRIC\_SIZE}
-    & Non-negative decimal with character suffix or a whole number with optional
-    character suffix
+    & Positive decimal with a character suffix or a positive integer with
+    optional character suffix
     & Controls the size of the symmetric heap per \ac{PE}. The value
       set in this environment variable is interpreted as number
       of bytes, unless the number is followed by a single
@@ -44,8 +44,8 @@ deprecation rationale for more details.
       \end{itemize}
       For example, the string 20m is equivalent to the integer value 20971520,
       or 20 megabytes. Similarly the string 3.1g is equivalent to the integer
-      value 3250585. Only one multiplier is recognized and a decimal value is
-      always required to be followed by a character suffix, so 20kk
+      value 3250585. Only one multiplier is recognized and a positive decimal
+      value is always required to be followed by a character suffix, so 20kk
       will not produce the same value as 20m, nor will invalid
       strings such as 20MB or 20.1 produce the desired result. Any unrecognized
       suffix is treated as an error and results in program termination with an
--- a/content/environment_variables.tex
+++ b/content/environment_variables.tex
@@ -37,10 +37,10 @@ deprecation rationale for more details.
     ceiling of the product of the numeric prefix and the scaling factor. The
     allowed character suffix for the scaling factor are as follows:
       \begin{itemize}
-        \item k or K multiplies by \(2^{10}\)  (kilobytes)
-        \item m or M multiplies by \(2^{20}\)  (megabytes)
-        \item g or G multiplies by \(2^{30}\)  (gigabytes)
-        \item t or T multiplies by \(2^{40}\)  (terabytes)
+        \item k or K multiplies by \(2^{10}\)  (kibibytes)
+        \item m or M multiplies by \(2^{20}\)  (mebibytes)
+        \item g or G multiplies by \(2^{30}\)  (gibibytes)
+        \item t or T multiplies by \(2^{40}\)  (tebibytes)
       \end{itemize}
       For example, string \enquote{20m} is equivalent to the integer value
       20971520, or 20 megabytes. Similarly the string \enquote{3.1m} is
--- a/content/environment_variables.tex
+++ b/content/environment_variables.tex
@@ -14,7 +14,7 @@ deprecation rationale for more details.
 
 \medskip{}
 
-\begin{tabular}{|l|l|l|}
+\begin{longtable}{|p{0.275\textwidth}|p{0.125\textwidth}|p{0.5\textwidth}|}
 \hline
 \textbf{Variable} & \textbf{Value} & \textbf{Description}
 \tabularnewline\hline
@@ -31,16 +31,18 @@ deprecation rationale for more details.
 %%
 \EnvVarDecl{SHMEM\_SYMMETRIC\_SIZE}
     & Non-negative integer
-    & Controls the size of the symmetric heap per PE. The value 
-      set in this environment variable is interpreted as number 
-      of bytes, unless the number is followed by a char that 
+    & Controls the size of the symmetric heap per PE. The value
+      set in this environment variable is interpreted as number
+      of bytes, unless the number is followed by a char that
       acts as a multiplier, where:
-        g or G multiplies by 2**30 (gigabytes)
-        k or K multiplies by 2**10 (kilobytes)
-        m or M multiplies by 2**20 (megabytes)
-      For example, the string 20m returns the integer value 20*2**20, 
-      or 20 megabytes. Only one multiplier is recognized, so 20kk 
-      will not produce the same value as 20m, nor will invalid 
+      \begin{itemize}
+        \item g or G multiplies by 2**30 (gigabytes)
+        \item k or K multiplies by 2**10 (kilobytes)
+        \item m or M multiplies by 2**20 (megabytes)
+      \end{itemize}
+      For example, the string 20m returns the integer value 20*2**20,
+      or 20 megabytes. Only one multiplier is recognized, so 20kk
+      will not produce the same value as 20m, nor will invalid
       strings such as 20MB produce the desired result.
     \tabularnewline\hline
 %%
@@ -48,6 +50,6 @@ deprecation rationale for more details.
     & Any
     & Enable debugging messages
     \tabularnewline\hline
-\end{tabular}
+\end{longtable}
 
 \medskip{}
--- a/content/environment_variables.tex
+++ b/content/environment_variables.tex
@@ -39,7 +39,7 @@ deprecation rationale for more details.
         \item k or K multiplies by \(2^{10}\)  (kilobytes)
         \item m or M multiplies by \(2^{20}\)  (megabytes)
         \item g or G multiplies by \(2^{30}\)  (gigabytes)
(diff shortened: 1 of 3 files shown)


content/environment_variables.tex
```diff
@@ -31,13 +31,25 @@
 %%
 \EnvVarDecl{SHMEM\_SYMMETRIC\_SIZE}
     & Non-negative integer
-    & Number of bytes to allocate for symmetric heap
+    & Controls the size of the symmetric heap per PE. The value
+      set in this environment variable is interpreted as number
+      of bytes, unless the number is followed by a char that
```
jdinan: s/char/character/ please

BryantLam: s/char/character suffix/

BryantLam:
Is it intended for unrecognized suffixes (e.g., T, kk) to be undefined behavior? I would like something to be defined, such as an error message and program termination.

BryantLam:
Nitpick, but G,K,M alphabetical ordering is an odd choice for powers-of-two. Why not K,M,G[,T,P]?

content/environment_variables.tex
```diff
@@ -31,13 +31,25 @@
 %%
 \EnvVarDecl{SHMEM\_SYMMETRIC\_SIZE}
     & Non-negative integer
-    & Number of bytes to allocate for symmetric heap
+    & Controls the size of the symmetric heap per PE. The value
+      set in this environment variable is interpreted as number
+      of bytes, unless the number is followed by a character that
```
nspark: Suggest: "a **single** character"

content/environment_variables.tex
```diff
@@ -31,13 +31,25 @@
...
+      acts as a multiplier, where:
+      \begin{itemize}
+        \item k or K multiplies by 2**10 (kilobytes)
+        \item m or M multiplies by 2**20 (megabytes)
+        \item g or G multiplies by 2**30 (gigabytes)
+      \end{itemize}
+      For example, the string 20m returns the integer value 20*2**20,
```
nspark: Suggest: "the string 20m **is equivalent to** the integer value **20971520**"

content/environment_variables.tex
```diff
@@ -31,13 +31,25 @@
...
+      or 20 megabytes. Only one multiplier is recognized, so 20kk
```
nspark: Instead of "Only one multiplier is recognized, ...", I suggest:
> If more than one multiplier character or any other non-numeric character is provided in this environment variable, the behavior is undefined.

content/environment_variables.tex
```diff
@@ -31,13 +31,25 @@
 %%
 \EnvVarDecl{SHMEM\_SYMMETRIC\_SIZE}
     & Non-negative integer
-    & Number of bytes to allocate for symmetric heap
+    & Controls the size of the symmetric heap per PE. The value
```
naveen-rn: PE changes

manjugv: PE -> \ac{PE}

jdinan: Replace "PE" with `\ac{PE}`.

content/environment_variables.tex
```diff
@@ -31,13 +31,25 @@
...
+      set in this environment variable is interpreted as number
+      of bytes, unless the number is followed by a character that
+      acts as a multiplier, where:
+      \begin{itemize}
+        \item k or K multiplies by 2**10 (kilobytes)
+        \item m or M multiplies by 2**20 (megabytes)
+        \item g or G multiplies by 2**30 (gigabytes)
```
BryantLam: What's the LaTeX rendered output for the double-star notation? Should this be guarded by math mode?

nspark:
Do we want to consider decimal heap sizes here (e.g., "4.5g")?

tonycurtis:
> On Mar 19, 2018, at 1:19 PM, Bryant C. Lam <notifications@github.com> wrote:
> 
> @BryantLam commented on this pull request.
> 
> In content/environment_variables.tex <https://github.com/openshmem-org/specification/pull/200#discussion_r175516544>:
> 
> > @@ -31,13 +31,25 @@
>  %%
>  \EnvVarDecl{SHMEM\_SYMMETRIC\_SIZE}
>      & Non-negative integer
> -    & Number of bytes to allocate for symmetric heap
> +    & Controls the size of the symmetric heap per PE. The value
> +      set in this environment variable is interpreted as number
> +      of bytes, unless the number is followed by a character that
> +      acts as a multiplier, where:
> +      \begin{itemize}
> +        \item k or K multiplies by 2**10 (kilobytes)
> +        \item m or M multiplies by 2**20 (megabytes)
> +        \item g or G multiplies by 2**30 (gigabytes)
> What's the LaTeX rendered output for the double-star notation? Should this be guarded by math mode?
> 
That would be \(2^{10}\) etc.

tony

naveen-rn:
> Do we want to consider decimal heap sizes here (e.g., "4.5g")? My expectation would be that the heap size would round up to the nearest (implementation-defined) page size.

We specifically say the value should be a "non-negative integer" - so this should be a implementation-defined allocation. I can be the nearest round up or even some default value?

jdinan:
@nspark Perhaps move forward with this PR and reopen #570?  IIRC, we had a broad discussion about decimal sizes and decided to stick with integers.

content/environment_variables.tex
```diff
@@ -31,13 +31,25 @@
...
     & Non-negative integer
-    & Number of bytes to allocate for symmetric heap
+    & Controls the size of the symmetric heap per \ac{PE}. The value
+      set in this environment variable is interpreted as number
+      of bytes, unless the number is followed by a single
+      character that acts as a multiplier, where:
+      \begin{itemize}
+        \item k or K multiplies by \(2^{10}\)  (kilobytes)
+        \item m or M multiplies by \(2^{20}\)  (megabytes)
+        \item g or G multiplies by \(2^{30}\)  (gigabytes)
```
BryantLam: Add `t or T` for terabytes (tebibytes?).

content/environment_variables.tex
```diff
@@ -31,13 +31,25 @@
 %%
 \EnvVarDecl{SHMEM\_SYMMETRIC\_SIZE}
     & Non-negative integer
```
BryantLam: "Non-negative integer with optional character suffix"

content/environment_variables.tex
```diff
@@ -31,13 +31,25 @@
...
+      character that acts as a multiplier, where:
+      \begin{itemize}
+        \item k or K multiplies by \(2^{10}\)  (kilobytes)
+        \item m or M multiplies by \(2^{20}\)  (megabytes)
+        \item g or G multiplies by \(2^{30}\)  (gigabytes)
+      \end{itemize}
+      For example, the string 20m is equivalent to the integer value 20971520,
+      or 20 megabytes. Only one multiplier is recognized, so 20kk
+      will not produce the same value as 20m, nor will invalid
+      strings such as 20MB produce the desired result.
```
BryantLam: Needs an error condition. I *hate* the idea that this is implementation-defined. I want the program to error out instead of silently using the wrong symmetric heap size (or worse, the implementation picks one for you and ignores your obviously wrong size request).

"Only one multiplier is recognized, so 20kk and 20MB are not equivalent to 20m and will <generate some error>."
with e.g., <generate some error> = "terminate the program with an error report for unrecognized suffix"

BryantLam:
Alas, clearly I meant to hit "request for changes".

content/environment_variables.tex
naveen-rn: @BryantLam How about:
Any unrecognized suffix is treated as an error and results in program termination with an error report.

content/environment_variables.tex
```diff
@@ -30,14 +30,29 @@
...
+    & Non-negative integer with optional character suffix
+    & Controls the size of the symmetric heap per \ac{PE}. The value
+      set in this environment variable is interpreted as number
+      of bytes, unless the number is followed by a single
+      character that acts as a multiplier, where:
+      \begin{itemize}
+        \item k or K multiplies by \(2^{10}\)  (kilobytes)
+        \item m or M multiplies by \(2^{20}\)  (megabytes)
+        \item g or G multiplies by \(2^{30}\)  (gigabytes)
+        \item t or T multiplies by \(2^{40}\)  (tebibytes)
```
nspark: Use base-2 or base-10 size names consistently. "Tebibytes" is used here, while "gigabytes" et al. are used previously.

nspark:
@naveen-rn I don't want to derail this one before the upcoming meeting, but I had [previously suggested](https://github.com/openshmem-org/specification/issues/199#issuecomment-357354498) a "decimal + suffix" form:

> We [users] were thinking along similar lines that that a decimal + suffix value should be at least that many bytes, rounded up to the page size.

Since we're updating (expanding) the semantics here, I think we should do the "whole thing" and not leave a piece left over that we'll probably want to fix in the next release. I think we should reconsider supporting a floating-point base + scaling as acceptable values for `SHMEM_SYMMETRIC_SIZE`.

I think it is appropriate to leave the final heap size as implementation-defined, provided it is at least as large as the size in bytes that would result from the natural operations and conversions to/from `double` and `size_t`. This could look something like:

```c
char *endptr = NULL;
char *envvar = getenv("SHMEM_SYMMETRIC_SIZE");
if (!envvar) return /* impl. default size */;
double value = strtod(envvar, &endptr));
size_t scaling = 1;
if (endptr)
  switch (*endptr) {
  case 't' : scaling <<= 10; /* fallthrough */
  case 'g' : scaling <<= 10; /* fallthrough */
  case 'm' : scaling <<= 10; /* fallthrough */
  case 'k' : scaling <<= 10; break;
  case '\0': break;
  default: abort();
  }
size_t heap_size = value * scaling;
return heap_size;
```

Since implementations may want to round up to the nearest CPU or NIC page size, that should be permissible as implementation-defined behavior.

naveen-rn:
@jdinan if possible, can you refresh us with the reason why we decided to stay with non-negative integers with char suffix?

@nspark after reading back the text again and trying to add float specific change - it looks like we are still trying to make an assertion that the users can expect a specified size from the env variable for SHEAP allocation. I think we should say explicitly that this env var is just a **hint** for the library and the libraries are not expected to provide the asked size.

BryantLam:
@naveen-rn, my definition of a hint is that the implementation would be free to ignore your value and give you any size. I don't think that's what you're proposing, because we would expect the implementation to at least give you the requested size (but realistically a bit more for rounding up). I don't think saying "hint" is required if this is the case.

jdinan:
@naveen-rn IIRC the rationale was to standardize existing behavior.  We surveyed implementations and many already support what is proposed.

nspark:
@naveen-rn Yes, I'm fine excluding fractional bytes, though I suggest saying that `SHMEM_SYMMETRIC_SIZE` is a "**positive** decimal with a character suffix or a **positive integer** with optional character suffix". I don't see a point in allowing `SHMEM_SYMMETRIC_SIZE` to be equal to zero.

BryantLam:
Is there a reason why the decimal portion is not also optional?

My preference: "**positive** ~~non-negative~~ decimal with ~~a~~ **optional** character suffix or a **positive integer** ~~whole number~~ with optional character suffix".

Scripts could calculate and pass floating-point values to `SHMEM_SYMMETRIC_SIZE`. I don't expect that as the common case (`bc` and Bash `(( arithmetic ))` both emit integer results by default), but I don't see why we should exclude it.

naveen-rn:
> Is there a reason why the decimal portion is not also optional?

If you look at the current description, it is generic about the actual SHEAP size available for the users. There is no such actual SHEAP size available throughout the runtime of the application. Consider the example in https://github.com/openshmem-org/specification/issues/199#issuecomment-357483535

The decimal to integer conversion without multiplier needs some kind of rounding off by the implementation and we have to speak about what the resultant value will be after the round off in description. This would imply that implementations have to provide that actual value for allocation, but they can actually allow this only on the first initialization. 

I'm not sure how to word this subtle difference in the description. That's the reason I decided to skip supporting positive decimal without multiplier.

BryantLam:
@naveen-rn -- Sorry if I'm misunderstanding. The example https://github.com/openshmem-org/specification/issues/199#issuecomment-357483535 pertains to integer values that would equally affect decimal values for `SHMEM_SYMMETRIC_HEAP`. I don't see why a decimal-without-suffix value for SHEAP would be any more ambiguous than integer-without-suffix values, especially if "no such actual SHEAP size available throughout the runtime of the application" applies for both.

> This would imply that implementations have to provide that actual value for allocation, but they can actually allow this only on the first initialization.

I agree with you. I suspect implementations would simply `ceil(float value)` and use that as the integer value as the starting point for SHEAP (again, it'll probably be higher in the end due to page rounding). It allows the first-allocation to still work as intended and isn't any more ambiguous with or without a suffix on the decimal. Maybe the ambiguity that needs to be resolved is to say that rounding-up will occur?

I could be totally wrong though. I'm not going to challenge this if you or others don't think this is an unambiguous change to make. It's simply odd to me that this argument was being applied for decimal-without-suffix while not recognizing that it equally affected integer-without-suffix.

BryantLam:
Re-reading your comment because I'm a bad reader.

> Maybe the ambiguity that needs to be resolved is to say that rounding-up will occur?

I think that's the subtle difference that you're alluding to that is missing. We would have to specify something in the description about rounding the decimal value.

Though 1.23456k doesn't yield a nice integer value, either, so why is this one okay?

naveen-rn:
> Though 1.23456k doesn't yield a nice integer value, either, so why is this one okay?
Its a good catch... I'm not sure whether there will be any actual use for such precision but still it is a valid input.

As you said, we either need to explain about the rounding off or restrict the decimal digits.

content/environment_variables.tex
```diff
@@ -30,14 +30,32 @@
...
+      character that acts as a multiplier, where:
+      \begin{itemize}
+        \item k or K multiplies by \(2^{10}\)  (kilobytes)
+        \item m or M multiplies by \(2^{20}\)  (megabytes)
+        \item g or G multiplies by \(2^{30}\)  (gigabytes)
+        \item t or T multiplies by \(2^{40}\)  (terabytes)
+      \end{itemize}
+      For example, the string 20m is equivalent to the integer value 20971520,
+      or 20 megabytes. Similarly the string 3.1g is equivalent to the integer
+      value 3250585. Only one multiplier is recognized and a positive decimal
```
jdinan: I think the above number is incorrect -- 3.1g = 3.1 * 2^30 = 3328599654.4

Should we specify what to do with the 4/10 of a byte?

content/environment_variables.tex
```diff
@@ -30,14 +30,32 @@
...
+      value is always required to be followed by a character suffix, so 20kk
+      will not produce the same value as 20m, nor will invalid
+      strings such as 20MB or 20.1 produce the desired result. Any unrecognized
```
jdinan: Why is 20.1 invalid if I can generate the same thing with a multiplier (0.01962890625K)?

jdinan:
Does there need to be a leading digit?  Are both ".5G" and "0.5G" are allowed?

jdinan:
@naveen-rn It's not as simple as limiting the number of decimal digits, e.g. 1.1K = 1126.4.

nspark:
How about something like...?
> `SHMEM_SYMMETRIC_SIZE`: A positive decimal value (integer or floating point) with an optional character suffix that specifies a scaling factor. The value set in this environment variable is interpreted as an implementation-defined size (in bytes) at least as large as the integer ceiling of the product of the numeric prefix and the scaling factor, which is specified by the optional character suffix as in the following table: [...]

nspark:
If we want to be specific about the acceptable forms of the floating-point numeric prefix, we could reference the _fractional-constant_ grammar of C11/C17 &sect;6.4.4.2-1:

> *fractional-constant:
> digit-sequence <sub>opt</sub> . digit-sequence
> digit-sequence .*

Or, we could reference the textual definition of _significant part_ &sect;6.4.4.2-2:

> A floating constant has a _significand part_ that may be followed by an _exponent part_ and a suffix that specifies its type. The components of the significand part may include a digit sequence  representing  the  whole-number part,  followed  by  a  period  (`.`),  followed  by  a digit sequence representing the fraction part.  [...] Either  the  whole-number  part  or  the  fraction  part  has  to  be  present;  for  decimal  floating constants, either the period or the exponent part has to be present.

Also, maybe "floating point" isn't quite the right word either, since we're not accepting exponents.

naveen-rn:
@jdinan Thanks for the comments - I will fix it. 
But, if i'm correct - we are still contemplating on the use of the term floating point and are also trying to cover the case with leading decimal digit as 0 or empty. Are we still planning for regex?

content/environment_variables.tex
```diff
@@ -30,14 +30,30 @@
...
+    implementation-defined size (in bytes) at least as large as the integer
+    ceiling of the product of the numeric prefix and the scaling factor. The
+    allowed character suffix for the scaling factor are as follows:
+      \begin{itemize}
+        \item k or K multiplies by \(2^{10}\)  (kilobytes)
+        \item m or M multiplies by \(2^{20}\)  (megabytes)
+        \item g or G multiplies by \(2^{30}\)  (gigabytes)
+        \item t or T multiplies by \(2^{40}\)  (terabytes)
+      \end{itemize}
+      For example, the string 20m is equivalent to the integer value 20971520,
```
jdinan: Please add quotations around the strings.  For example, ``20m".

naveen-rn: from the review comments during Sep spec meeting: quotes for the values and add example for 0.5 vs .5 for the usage

content/environment_variables.tex
```diff
@@ -30,14 +30,31 @@
...
+        \item g or G multiplies by \(2^{30}\)  (gigabytes)
+        \item t or T multiplies by \(2^{40}\)  (terabytes)
+      \end{itemize}
+      For example, string \enquote{20m} is equivalent to the integer value
+      20971520, or 20 megabytes. Similarly the string \enquote{3.1m} is
+      equivalent to the integer value 3250586. Only one multiplier is
+      recognized, so \enquote{20kk} will not produce the same desired result as
+      \enquote{20m}. Usage of string \enquote{.5m} will yield the same result as
+      the string \enquote{0.5m}. Any unrecognized suffix is treated as an error
+      and results in program termination with an error report.
```
BryantLam: @nspark made a comment that `shmem_init_thread` could return (an unspecified) non-zero status instead of aborting. Assuming the error report is left in, this request is reasonable. Perhaps something like:
> Any unrecognized suffix is an error, generates an error report, and either returns from `shmem_init_thread` or otherwise causes program termination.

The second clause is left vague, but it only affects `shmem_init` today. Writing that in explicitly is also an option:

> Any unrecognized suffix is an error, generates an error report, and either returns from `shmem_init_thread` or causes program termination from `shmem_init`.

The **Return Values** from `shmem_init_thread` already says it returns non-zero if not success.

There's the separate concern of whether this error should be consistent and always result in program termination. (This would be my preference, but others can chime in.)

jdinan: There have been (and may still be) SHMEM implementations that map the physical address space during loading and, as such, will detect this error before any user code runs.  The first text suggestion may better support such implementations by allowing them to abort before any OpenSHMEM library calls are performed (including `shmem_init_thread`).

naveen-rn: I too would prefer the first text. Will make the change and do a special ballot. Let me know, if this works.

naveen-rn:
Special Ballot for March-2019 specification discussion:
Changes: fc74200 
1. Adding support for return from shmem_init_thread incase of incorrect usage of SHMEM_SYMMETRIC_SIZE env variable

jdinan:
I'm concerned about the following special ballot change:

> "Any unrecognized suffix is an error, generates an error report, and either returns from shmem_init_thread or otherwise causes program termination."

What value is returned by `shmem_init_thread`?  Are `shmem_init` and `start_pes` required to terminate?  A second nitpick is that it might be better to say "unrecognized string" since a value like "#*@$ Jim Ruins Everything" is invalid in a way that is not exactly an "unrecognized suffix."

naveen-rn:
image

If I can go back in time - #199

nspark:
[This one](https://www.youtube.com/watch?v=BsKbwR7WXN4)'s for @naveen-rn

nspark:
@jdinan I think the _intent_ is that a malformed value for `SHMEM_SYMMETRIC_SIZE` would cause `shmem_init_thread` to return a nonzero value indicating that SHMEM was not properly initialized. If the symmetric heap was mapped during loading (vs. during the API initialization call), or if `shmem_init` or `start_pes` were called, termination would be expected.

What about something like...?

> An invalid value for `SHMEM_SYMMETRIC_SIZE` is an error, which may be reported by returning a nonzero value from `shmem_init_thread` or causing program termination.

jdinan:
Another question -- "Positive decimal value (integer or floating point) with an optional character suffix" -- is `SHMEM_SYMMETRIC_SIZE=0` still allowed?  Used to be non-negative, which allowed 0.

jdinan:
@nspark Is `shmem_init_thread` required to return a nonzero value, or can it also abort?  "May" reads to me like either is allowed.

naveen-rn:
> is SHMEM_SYMMETRIC_SIZE=0 still allowed?
AFAIU, 0 is neither positive or negative - so not allowed. Between, why should be allow?

nspark:
> Is `shmem_init_thread` required to return a nonzero value, or can it also abort? "May" reads to me like either is allowed.

Even if it said "...shall be reported...", I suppose the "or" makes it ambiguous what could happen when the library is initialized by `shmem_init_thread`. While my preference is to require nonzero return from `shmem_init_thread`, I'm inclined to leave it ambiguous for implementations that might manage the symmetric heap before executing user code, in addition to the `shmem_init` and `start_pes` initialization paths.

So, with the slight tweak to be more direct, I'll suggest...

> An invalid value for `SHMEM_SYMMETRIC_SIZE` is an error, which the OpenSHMEM library shall report by either returning a nonzero value from `shmem_init_thread` or causing program termination.

I'm not sure how to suggest the preference that any library that _can_ return a nonzero value from `shmem_init_thread` should do so vs. terminating the program.

jdinan:
@naveen-rn Symmetric size of 0 makes sense for apps that don't use the symmetric heap.  It's allowed in OpenSHMEM 1.4 with the current "non-negative" text.

jdinan:
Ref: https://github.com/openshmem-org/specification/issues/199#issuecomment-473444661

@jeffhammond raised a concern about loss of precision when converting from decimal (base 10) provided via `SHMEM_SYMMETRIC_SIZE` and binary used for arithmetic used by implementations (e.g. 0.2 is repeating in binary 0.0011 0011 ...).  When you scale this by 2^20 or 2^30, it's possible to get a number that is not exactly what the user asked for.

naveen-rn:
> Ref: #199 (comment)

@jdinan @jeffhammond I see the issue from the comment. I can change kilo/mega/giga/tera to kibi/mebi/gibi/tebi. 

That said, the issue I understand is that there is mismatch in the name and representation. It has nothing to do with the precision guarantees. Are we guaranteeing precision as a requirement in this proposal? FWIU, ```SHMEM_SYMMETRIC_HEAP``` is a hint rather than a hard requirement from an implementation. It is never guaranteed to the user that they will always get the requested amount of memory. There are multiple factors which can result in such similar loss, implementation optimization, data structure and others.

Let me know, what you think.

minsii:
@naveen-rn I am following Jim and Nick's discussion about the text description for invalid value of `SHMEM_SYMMETRIC_SIZE `. Could you please let me know the final decision ? Will you change the text and specify a program termination behavior when the program calls `shmem_init` and `start_pes` with an invalid value ?

naveen-rn:
The following are the consolidated suggested changes from recent review comments.

1. Handling 0 as value
In the current value requirement (_Positive decimal value (integer or floating point) with an optional character suffix_) we drop supporting 0 as input. I'm fine with this change and would prefer to keep the behavior undefined. 
Reason: Irrespective of this hint from the user - I'm not sure how effectively can an implementation make use of this information. If people disagree, then please verify whether we could change the value text into something as follows: _Decimal value(0, positive integer, or floating point) with an optional character suffix._ To me the new change looks a bit ugly, if needed please modify.

2. Mismatch in the name and definition of the suffix
I understand there is an issue, will change kilo/mega/giga/tera to kibi/mebi/gibi/tebi.

3. Error handling for invalid suffix/string/value
The most recent suggested text change from @nspark is: _An invalid value for SHMEM_SYMMETRIC_SIZE is an error, which the OpenSHMEM library shall report by either returning a nonzero value from shmem_init_thread or causing program termination._ With respect to ```shmem_init``` and ```start_pes```, I don't see a huge difference with in the suggested change with the current text, we are still leaving the  usage ambiguous. The major improvement is that we are clarifying that in fact the return from ```shmem_init_thread``` is a non-zero value. As we are introducing ```shmem_init_thread``` in the explanation before the actual defn - this change is very useful. 
@jdinan and @minsii - let me know if this change looks fine.

minsii:
Thanks for the summary, @naveen-rn . 
> Handling 0 as value
In the current value requirement (Positive decimal value (integer or floating point) with an optional character suffix) we drop supporting 0 as input. I'm fine with this change and would prefer to keep the behavior undefined.
Reason: Irrespective of this hint from the user - I'm not sure how effectively can an implementation make use of this information. If people disagree, then please verify whether we could change the value text into something as follows: Zero or Positive Decimal value (integer or floating point) with an optional character suffix. To me the new change looks a bit ugly, if needed please modify.

`non-negative` reads to me as `zero or positive`. I think it is better to keep the same rule of valid values as current spec, because it allows the runtime to optimize performance (e.g., in OSHMPI, we have to always flush symm-heap window and symm-data window at `shmem_quiet`. If user says the symm-heap is zero, then we can skip the first flush) & memory consumption (entire symm-heap is pre-allocated in OSHMPI) when possible. If `non-negative` is confusing (we use this term in MPI, but we can make it clearer here), I think "Zero or Positive Decimal value (integer or floating point) " is good too.


> Error handling for invalid suffix/string/value
The most recent suggested text change from @nspark is: An invalid value for SHMEM_SYMMETRIC_SIZE is an error, which the OpenSHMEM library shall report by either returning a nonzero value from shmem_init_thread or causing program termination. With respect to shmem_init and start_pes, I don't see a huge difference with in the suggested change with the current text, we are still leaving the usage ambiguous. The major improvement is that we are clarifying that in fact the return from shmem_init_thread is a non-zero value.

I think it is OK to leave the behavior ambiguous. I.e., in the program using `shmem_init_thread`, the runtime can either continue with a nonzero return value, or terminate.

manjugv:
March 25th Spec Meeting: Voting postponed

jamesaross:
A small, but important detail...

The language about the symmetric heap size being **per PE** seems to have disappeared from the definition.

naveen-rn:
@jamesaross 
I think the **per PE** is a good addition. But, it seems to be a new text and not something missing from previous text. The previous text was **Number of bytes to allocate for symmetric heap**.

naveen-rn:
@nspark @jdinan @jamesaross @minsii 
With the following commits - we have made the following fixes for SHMEM_SYMMETRIC_HEAP env var:
b4d5e4, ac9980, fabc73, beb079, 163f42

1. Changed decimal to binary names for suffix
2. Added _non-negative decimal_ instead of _positive decimal_ to include in zero in the value
3. Included _per PE_ in the description for more clarity
4. Updated error handling option

Let me know, for any other fixes.

content/environment_variables.tex
```diff
@@ -30,14 +30,34 @@
     \tabularnewline\hline
 %%
 \EnvVarDecl{SHMEM\_SYMMETRIC\_SIZE}
-    & Non-negative integer
-    & Number of bytes to allocate for symmetric heap
+    & Non-negative decimal value (integer or floating point) with an optional
+    character suffix
```
jdinan: Does "decimal" mean base-10 or that it has a decimal point?  The former seems unnecessary and the latter contradicts the parenthetical "integer" comment.  Can we shorten this to "Non-negative integer or floating point value with an optional character suffix"?

content/environment_variables.tex
```diff
@@ -30,14 +30,34 @@
...
+    & The value set in this environment variable is interpreted as an
+    implementation-defined size (in bytes) at least as large as the integer
+    ceiling of the product of the numeric prefix and the scaling factor per
+    \ac{PE}. The allowed character suffix for the scaling factor are as follows:
```
jdinan: suffix -> suffixes

content/environment_variables.tex
```diff
@@ -30,14 +30,34 @@
...
+      \begin{itemize}
+        \item k or K multiplies by \(2^{10}\)  (kibibytes)
+        \item m or M multiplies by \(2^{20}\)  (mebibytes)
+        \item g or G multiplies by \(2^{30}\)  (gibibytes)
+        \item t or T multiplies by \(2^{40}\)  (tebibytes)
+      \end{itemize}
+      For example, string \enquote{20m} is equivalent to the integer value
+      20971520, or 20 megabytes. Similarly the string \enquote{3.1m} is
```
jdinan: megabytes -> mebibytes

content/environment_variables.tex
```diff
@@ -30,14 +30,34 @@
...
+      equivalent to the integer value 3250586. Only one multiplier is
+      recognized, so \enquote{20kk} will not produce the same desired result as
+      \enquote{20m}. Usage of string \enquote{.5m} will yield the same result as
```
jdinan: Remove "desired".

Another comment -- this seems to indicate that "20kk" is a valid input.  Do we want to say something like, "Any characters following the multiplier are ignored, so 20kk ..."?  Or say that the multiplier must be the last character in the string?  The current text doesn't indicate whether "20k stupid comments from Jim" is a valid string.  According to the paragraph below, if this is an invalid string, I need to check for it and raise an error.

content/environment_variables.tex
naveen-rn: fixed

content/environment_variables.tex
naveen-rn: fixed

content/environment_variables.tex
naveen-rn: fixed

content/environment_variables.tex
naveen-rn: Based on this text specifically the only one multiplier is recognized statement, 20kk or 20K<something something> is still a valid input. It won't error out, it just uses the valid identifiable input characters in the string. 
What do we want? Do we need to consider this as an invalid string?

I prefer the following statement:
Only one multiplier is recognized and any characters following the multiplier are ignored, so ....

To me the multiplier must be the last character statement seems strong and unnecessary.

jdinan: I don't really have a strong preference.  Since we've had this extra characters thing all along and people seem ok with it, your proposed change is probably the easiest option.

naveen-rn: fixed

content/environment_variables.tex
```diff
@@ -30,14 +30,36 @@
...
 %%
 \EnvVarDecl{SHMEM\_SYMMETRIC\_SIZE}
-    & Non-negative integer
-    & Number of bytes to allocate for symmetric heap
+    & Non-negative integer or floating point value with an optional character
+    suffix
+    & The value set in this environment variable is interpreted as an
+    implementation-defined size (in bytes) at least as large as the integer
+    ceiling of the product of the numeric prefix and the scaling factor per
+    \ac{PE}. The allowed character suffixes for the scaling factor are as
```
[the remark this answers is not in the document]
naveen-rn: fixed

content/environment_variables.tex
naveen-rn: marking as resolved - as the text has gone multiple updates and this one is no longer relevant.

content/environment_variables.tex
naveen-rn: fixed added *equivalent* - marking it as resolved.

content/environment_variables.tex
naveen-rn: changed as  
>Only one multiplier is recognized and any characters following the multiplier are ignored

content/environment_variables.tex
naveen-rn: Fixed - marking as resolved

naveen-rn: closing - changes are outdated

content/environment_variables.tex
naveen-rn: fixed with ^^ - marking as resolved

content/environment_variables.tex
naveen-rn: fixed - recommended changes are outdated - updated with new texts - marking as resolved

content/environment_variables.tex
naveen-rn: fixed - recommended changes are outdated - updated with new texts - marking as resolved

content/environment_variables.tex
naveen-rn: recommended changes are outdated - updated with new texts - marking as resolved. please open new review if needed.

content/environment_variables.tex
naveen-rn: recommended changes are outdated - updated with new texts - marking as resolved. please open new review if needed.

naveen-rn:
Planned for special ballot and voting in May 2019 F2F meeting:

With the following commits - we have made the following fixes for SHMEM_SYMMETRIC_HEAP env var:
b4d5e4, ac9980, fabc73, beb079, 163f42

1. Changed decimal to binary names for suffix
2. Added non-negative decimal instead of positive decimal to include in zero in the value
3. Included per PE in the description for more clarity
4. Updated error handling option

PR #204, opened 2018-02-06
Update atomic memory model text

### LINKED ISSUE #172: Memory Model Clarifications
Reported by a reader of the 1.4 draft:

In 1.4 draft, Chapter 3 is “memory model”.  Section 3.1 is “atomicity guarantees”.  It says atomicity is guaranteed under simultaneous by atomics of the same type, but says nothing about atomicity or ordering otherwise.
 
That seems like an obvious place to mention concurrent non-atomic accesses.  I will claim that the negative (“an implementation is free to …”) _is_ still defining, but maybe not everybody agrees.
 
Another place is Annex C, which seems to me like a reasonable place to note the race-y behavior under multiple simultaneous writes.
 

Reading the atomics section suggests to me another use case which is maybe undefined:
 
put(dst, src, n, dstpe)
barrier()
x  = atomic_finc(&dst[ran() % n], ran() % num_pes())
barrier()
 
One observation here is that the spec (3.1) says accesses by atomics are exclusive, but not exclusive by other routines.
 
I’ll claim a reasonable programmer will expect that the barrier() ensures only the when any atomic_finc() routine runs, that only atomic_finc() routines are accessing dst, and so guarantees all atomic_finc() accesses are atomic with respect to _all_ memory accesses – e.g., cannot overlap in time with the earlier put().
 
However, there was an earlier non-atomic access (the put()), and I do not see anything in the memory model section which suggests barrier() can be used to “scope” the atomic and non-atomic calls.
 
The documentation for, say, shmem_barrier_all() says it “ensures completion of all previously issued memory stores … AMOs and RMA routines …” but again I don’t see a direct line saying this makes the put() and atomic_finc() usage above safe-to-use.
 
This may be an excessively picky reading, but it _is_ a specification :^)

jdinan:
Clarify the various cases that violate atomicity.  There should be no semantic changes.

Closes #172

### DIFF
--- a/content/memory_model.tex
+++ b/content/memory_model.tex
@@ -55,19 +55,58 @@ memory section of each \ac{PE}.
 \subsection{Atomicity Guarantees}\label{subsec:amo_guarantees}
 
 \openshmem contains a number of routines that perform atomic operations on
-symmetric data objects (Section \ref{sec:amo}).  These routines guarantee that
-concurrent accesses to the same location by \openshmem atomic operations using
-the same datatype will be exclusive.  However, they do not guarantee
-exclusivity when concurrent accesses to the same location are performed using
-\openshmem atomic operations using different datatypes, when non-atomic
-\openshmem operations are used to access the given location, or when
-non-\openshmem operations are performed on the given location.  The result in
-such scenarios is undefined.
+symmetric data objects, which are defined in Section \ref{sec:amo}.  In
+addition, \openshmem defines routines that are atomic compatible, e.g. the
+point-to-point synchronization routines defined in
+Section~\ref{subsec:p2p_intro}.  The atomic and atomic compatible routines
+guarantee that concurrent accesses by any of these routines to the same
+location and using the same datatype (specified in Tables~\ref{stdamotypes} and
+\ref{extamotypes}) will be exclusive.  Two operations are concurrent when the
+first operation is not completed at the target PE (e.g. by a call to
+\FUNC{shmem\_quiet}) prior to the start of the second operation.
 
-For example: during the execution of an atomic remote integer increment
+\openshmem atomic operations do not guarantee exclusivity in the following
+scenarios, all of which result in undefined behavior.
+\begin{enumerate}
+    \item When concurrent accesses to the same location are performed using
+        \openshmem atomic or atomic compatible operations using different
+        datatypes.
+    \item When atomic and non-atomic \openshmem operations are used to access
+        the same location concurrently.
+    \item When \openshmem atomic operations and non-\openshmem operations (e.g.
+        load and store operations) are used to access the same location
+        concurrently.
+\end{enumerate}
+For example, during the execution of an atomic remote integer increment, i.e. \FUNC{shmem\_atomic\_inc},
 operation on a symmetric variable \VAR{X}, no other \openshmem atomic operation
 may access \VAR{X}.  After the increment, \VAR{X} will have increased its value
 by \CONST{1} on the destination \ac{PE}, at which point other atomic operations
 may then modify that \VAR{X}.  However, access to the symmetric object \VAR{X}
 with non-atomic operations, such as one-sided \OPR{put} or \OPR{get} operations,
 will invalidate the atomicity guarantees.
+
+\cexample
+    {The following \CorCpp example illustrates scenario 1.  In this example,
+    different datatypes are used to access the same location concurrently,
+    resulting in undefined behavior.  The undefined behavior can be resolved by
+    using the same datatype in all concurrent operations.  For example, the
+    32-bit value can be left-shifted and a 64-bit atomic OR operation can be
+    used.}
+    {./example_code/amo_scenario_1.c}
+
+\cexample
+    {The following \CorCpp example illustrates scenario 2.  In this example,
+    atomic increment operations are concurrent with a non-atomic reduction
+    operation, resulting in undefined behavior.  The undefined behavior can be
+    resolved by inserting a barrier operation before the reduction.  The
+    barrier ensures that all local and remote AMOs have completed before the
+    reduction operation accesses $x$.}
+    {./example_code/amo_scenario_2.c}
+
+\cexample
+    {The following \CorCpp example illustrates scenario 3.  In this example, an
+    \openshmem atomic increment operation is concurrent with a local increment
+    operation, resulting in undefined behavior.  The undefined behavior can be
+    resolved by replacing the local increment operation with an \openshmem
+    atomic increment.}
+    {./example_code/amo_scenario_3.c}
--- a/content/memory_model.tex
+++ b/content/memory_model.tex
@@ -58,10 +58,12 @@ memory section of each \ac{PE}.
 symmetric data objects, which are defined in Section \ref{sec:amo}.  In
 addition, \openshmem defines routines that are atomic compatible, e.g. the
 point-to-point synchronization routines defined in
-Section~\ref{subsec:p2p_intro}.  The atomic and atomic compatible routines
+Section~\ref{subsec:p2p_intro}.  The atomic routines
 guarantee that concurrent accesses by any of these routines to the same
 location and using the same datatype (specified in Tables~\ref{stdamotypes} and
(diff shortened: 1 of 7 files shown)


content/memory_model.tex
```diff
@@ -54,11 +54,15 @@
...
-exclusivity in combination with other routines, either inside \openshmem or
-outside.
+\openshmem contains a number of routines that perform atomic operations on
+symmetric data objects (Section \ref{sec:amo}).  These routines guarantee that
+concurrent accesses to the same location by \openshmem atomic operations using
+the same datatype will be exclusive.  However, they do not guarantee
+exclusivity when concurrent accesses to the same location are performed using
+\openshmem atomic operations using different datatypes, when non-atomic
+\openshmem operations are used to access the given location, or when
+non-\openshmem operations are performed on the given location.  The result in
```
jdinan: Add: "(e.g. load and store operations)"

content/memory_model.tex
```diff
@@ -54,11 +54,15 @@
...
-atomically (Section \ref{sec:amo}).  These routines guarantee that accesses by
-\openshmem's atomic operations with the same datatype will be exclusive, but do not guarantee
-exclusivity in combination with other routines, either inside \openshmem or
-outside.
+\openshmem contains a number of routines that perform atomic operations on
+symmetric data objects (Section \ref{sec:amo}).  These routines guarantee that
+concurrent accesses to the same location by \openshmem atomic operations using
+the same datatype will be exclusive.  However, they do not guarantee
+exclusivity when concurrent accesses to the same location are performed using
+\openshmem atomic operations using different datatypes, when non-atomic
```
jdinan: Add wait/test with same datatype

content/memory_model.tex
```diff
@@ -54,11 +54,15 @@
...
 \subsection{Atomicity Guarantees}\label{subsec:amo_guarantees}
 
-\openshmem contains a number of routines that operate on symmetric data
-atomically (Section \ref{sec:amo}).  These routines guarantee that accesses by
-\openshmem's atomic operations with the same datatype will be exclusive, but do not guarantee
-exclusivity in combination with other routines, either inside \openshmem or
-outside.
+\openshmem contains a number of routines that perform atomic operations on
+symmetric data objects (Section \ref{sec:amo}).  These routines guarantee that
+concurrent accesses to the same location by \openshmem atomic operations using
```
jdinan: Add: "and atomic-compatible operations"?

content/memory_model.tex
```diff
@@ -54,11 +54,15 @@
...
+the same datatype will be exclusive.  However, they do not guarantee
```
jdinan: Instead of "datatype" reference the table in the AMO section and use consistent terminology.

content/memory_model.tex
spotluri: @jdinan, you mean to add wait/test with same datatype to the allowed/defined list, correct?

jdinan:
Requesting committee feedback on the updated draft: [osh-6ad7676.pdf](https://github.com/openshmem-org/specification/files/1952955/osh-6ad7676.pdf).
Flagged reviewers: @shamisp, @spotluri, @manjugv, @npark

jdinan:
Feedback from May 21 reading:
 * Follow up with @manjugv about shmem_barrier / interprocess synchronization in concurrency definition.
 * Examples: add comment above each incorrect line of code to clarify that it causes undefined behavior
 * Example 3: Look into using C11 atomic operation (is the type compatible with SHMEM operations?)

jdinan:
C11 specifies that "The size, representation, and alignment of an atomic type need not be the same as those of the corresponding unqualified type."  OpenSHMEM operations on such types are therefore not well-defined, so C11 atomics won't work in example 3.

jdinan:
Special ballot changes: 252a6a1 and 04a876d

content/memory_model.tex
```diff
@@ -54,16 +54,64 @@
...
-exclusivity in combination with other routines, either inside \openshmem or
-outside.
+\openshmem contains a number of routines that perform atomic operations on
+symmetric data objects, which are defined in Section \ref{sec:amo}.  In
+addition, \openshmem defines routines that are atomic compatible, e.g. the
+point-to-point synchronization routines defined in
+Section~\ref{subsec:p2p_intro}.  The atomic and atomic compatible routines
+guarantee that concurrent accesses by any of these routines to the same
+location and using the same datatype (specified in Tables~\ref{stdamotypes} and
+\ref{extamotypes}) will be exclusive.  In contrast to atomic operations, atomic
```
anshumang: `short` and `unsigned short` is supported by `wait_until` (not supported in amotypes); `float` and `double` are the other non-overlapping types.

manjugv:
July 23rd Meeting Notes: Ballot postponed

BryantLam:
### Atomic completion of `shmem_wait_until`
@shamisp expressed concern that `shmem_wait_until` is not efficiently implementable if it is to both guarantee atomic and non-atomic completion. He claimed `shmem_wait_until` should only work on non-atomic operations or only on atomic operations, but not both.

OpenSHMEM 1.4 already specifies that atomic and non-atomic completion can be achieved with `shmem_wait_until`:
> shmem_wait and shmem_wait_until wait for ivar to be changed by a write or an atomic operation issued by a PE. ...
>
> Note to implementors -- Implementations must ensure that shmem_wait and shmem_wait_until do not return before the update of the memory indicated by ivar is fully complete. Partial updates to the memory must not cause shmem_wait or shmem_wait_until to return.

This led into a discussion about whether the semantics for wait-until are effectively broken.

### Definition of concurrent
https://github.com/openshmem-org/specification/commit/04a876dcd16148e65aa105188831927724e86de4 specifies generic statements on concurrency that could apply to elsewhere in the document. One suggestion was to move this definition to a higher level/another section and simply reference it here.

content/memory_model.tex
jdinan: This text is intended to specify semantics for atomics, so those types shouldn't really come into play.  But, I see your point that as it's written, it also specifies semantics for overlapping wait operations without atomics.  How about splitting these cases up:
> The atomic routines guarantee that concurrent accesses by any of these routines to the same location and using the same datatype (specified in Tables~\ref{stdamotypes} and \ref{extamotypes}) will be exclusive.  Similarly, concurrent accesses by atomic and atomic compatible routines to the same location and using the same datatype will also be exclusive.

jdinan:
@shamisp @BryantLam Let's discuss this at the next RMA WG meeting.  The purpose of this ticket is only to clarify existing semantics (broken or not).  Restricting the behavior of `shmem_wait` is being pursued in separate proposal led by @anshumang and @spotluri.

I'm fine with moving the concurrency definition to a terms section.  Can we handle this separately as part of #228?

content/memory_model.tex
anshumang: Follow-up question : will there be a usecase for `short` and `unsigned short` if the amo APIs end up being the only way to signal `wait_until`?

jdinan: IIRC, InfiniBand doesn't have hardware support for 8- and 16-bit atomics.  That's one argument in favor supporting scalar put with the wait operations..

jdinan:
@manjugv Which definition are you referring to?  The definition of concurrency?

shamisp:
@jdinan I see it as an integral part of memory model.

jdinan:
@shamisp Can you spot me a couple more words?  I'm not sure how to interpret your feedback, and github comments are wonderfully context-free.

jdinan:
Feedback from Aug. 2 WG meeting: remove concurrency definition and atomic compatible; handle these as part of the memory model proposal.

jdinan:
BTW, the "left shift" comment in example 1 strikes me as questionable advice given that the original code has different behaviors depending on the endianness of the system it's running on.  Do we want to change that bit to something like: "For example, the 32-bit value can be left-shifted and a 64-bit atomic OR operation can be used."?

jdinan:
Special ballot changes: 252a6a1, 04a876d, 12e0687, and b3b5cfb.

shamisp:
@jdinan looks good to me.

PR #213, opened 2018-05-07
Clarify semantics of lock API

### LINKED ISSUE #174: Clear lock completes what?
This is need of clarifiction: "The ```shmem_clear_lock``` routine releases a lock previously set by ```shmem_set_lock``` after ensuring that all local and remote stores initiated in the critical region are complete."

Does this include RMA/AMO operations?

This also needs the standard threading disclaimer: that only the calling thread's action (and those of other threads that are ordered beforehand) can be ordered by this call.

### LINKED ISSUE #193: Lock API and Contexts
## Issue

OpenSHMEM 1.4 specifies:

> The shmem_clear_lock routine releases a lock previously set by shmem_set_lock after ensuring that all local and remote stores initiated in the critical region are complete.

The "remote stores" bit needs clarification.  In particular, does this requires that all contexts touched during the critical section must be quieted?  What if the context is private or serialized?

## Proposed Solution

> The shmem_clear_lock routine releases a lock previously set by shmem_set_lock after ensuring that all local stores and _operations initiated on the default context_ in the critical region are complete.

### LINKED ISSUE #205: Locks and Threads
## Summary

In OpenSHMEM 1.4, the lock API was defined to provide mutual exclusion between PEs and the behavior of recursive lock operations is undefined.  With the introduction of threading, we need to clarify the behavior when a lock that a given PE already holds or has already requested is requested again (e.g. by a second thread in that PE).  In terms of user impact, we must decide whether the lock API can provide:

1. Only mutual exclusion between PEs or 
1. Mutual exclusion between threads within PEs

The following example illustrates multiple threads using the lock API simultaneously:
```
int main(void) {
  static long lock = 0;
  shmem_init_thread(SHMEM_THREAD_MULTIPLE, ...);
  ...
#pragma omp for
  for (... ; ... ; ...) {
      shmem_lock_set(&lock);
      shmem_get(&lock);
      // local update
      shmem_put(...);
      shmem_lock_clear(&lock);
  }
  ...
}
```
## Proposed Changes

Requesting committee feedback to determine whether we should pursue option 1 or 2 above.

## Impact on Implementations

Existing implementations may use a non-reentrant lock algorithm and would need to be changed to support threading.

Implementations that provide a distributed queueing lock using the space provided in the symmetric lock object (e.g. https://github.com/Sandia-OpenSHMEM/SOS/blob/master/src/shmem_lock.h implements the MCS locking algorithm) may be unable to directly support multiple threads within a PE requesting a given lock at the same time.  This is because multiple threads at a PE cannot share the "who gets the lock after me" field that is stored in the lock object at that PE.  That is, the space requirements for such locking algorithms are proportional to the number of lock requesters and the space available (`sizeof(long)*npes`) is fixed.  In such cases, a two-level locking scheme could be used (local mutex gating access to the SHMEM mutex).

Implementations that poll over the network (e.g. spinlocks, bakery locks, etc.) have constant space requirements and can be used to support mutual exclusion between threads within PEs.

## Additional Materials

* [03-06-2018 -- Threads WG.pdf](https://github.com/openshmem-org/specification/files/1786915/03-06-2018.--.Threads.WG.pdf)

jdinan:
* Reorganize text to describe lock API before unlock API
 * Define unlock as performing quiet; this clarifies the interaction with contexts and nonblocking API routines
 * Define OpenSHMEM locks to be non-reentrant; this clarifies the  behavior when used by multithreaded PEs

Closes #205
Closes #193 
Closes #174

### DIFF
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -558,6 +558,10 @@ The following list describes the specific changes in \openshmem[1.5]:
 \item This item is a template for changelist entries and should be deleted
     before this document is published.
     \\See Annex~\ref{sec:changelog}.
+\item Clarified that the \openshmem lock API provides a non-reentrant mutex and
+    that \FUNC{shmem\_clear\_lock} performs a quiet operation on the default
+    context.
+    \\See Section~\ref{subsec:shmem_lock}
 \end{itemize}
 
 \section{Version 1.4}
--- a/content/shmem_lock.tex
+++ b/content/shmem_lock.tex
@@ -54,9 +54,15 @@ I = @\FuncDecl{SHMEM\_TEST\_LOCK}@(lock)
 
 \apinotes{
     The term symmetric data object is defined in Section \ref{subsec:memory_model}.
-    The lock variable should always be initialized to zero and accessed only by the \openshmem locking
-    \ac{API}.  Changing the value of the lock variable by other means without using
-    the \openshmem \ac{API}, can lead to undefined behavior.
+
+    The lock variable must be initialized to zero before any PE performs an
+    \openshmem lock operation on the given variable.  Accessing an in-use lock
+    variable using any method other than the \openshmem lock API, e.g. using
+    local load/store, RMA, or AMO operations, results in undefined behavior.
+
+    Calls to \FUNC{shmem\_ctx\_quiet} can be performed prior to calling the
+    \FUNC{shmem\_clear\_lock} routine to ensure completion of operations issued
+    on additional contexts.
 }
 
 \begin{apiexamples}
--- a/content/shmem_lock.tex
+++ b/content/shmem_lock.tex
@@ -18,28 +18,36 @@ I = @\FuncDecl{SHMEM\_TEST\_LOCK}@(lock)
 
 \begin{apiarguments}
 \apiargument{IN}{lock}{A symmetric data object that is a scalar variable or an array
-    of  length \CONST{1}.  This data  object  must  be set to \CONST{0} on all
-    \acp{PE} prior to the first use.  \VAR{lock}  must  be  of type \CONST{long}.
+    of length \CONST{1}.  This data object must be set to \CONST{0} on all
+    \acp{PE} prior to the first use.  \VAR{lock} must be of type \CONST{long}.
     When using \Fortran, it must be of default kind.}
 \end{apiarguments}
 
 \apidescription{
-    The \FUNC{shmem\_set\_lock} routine sets a mutual exclusion lock after  waiting
-    for  the lock  to be freed by any other \ac{PE} currently holding the lock.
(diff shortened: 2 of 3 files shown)


jdinan:
Feedback from May 21 reading:
 * Add sentence that users are responsible for quieting user-created contexts
 * Remove shmem_quiet() in the lock example
 * Follow up with @naveen-rn -- Does clear_lock in Cray SHMEM require a quiet?
 * Follow up with @BryantLam -- Do users expect quiet in clear_lock?
 * Anshuman Goswami -- Do we also want set/test lock to perform quiet?

BryantLam:
@jdinan -- Users expect quiet in clear_lock, so we should preserve the current behavior.

BryantLam:
Instead of simply removing the quiet from the shmem_lock example, I suggest replacing it with a succinct comment indicating that quiet-like behavior will be performed by clear_lock.

jdinan:
Special ballot changes: 45894b8

anshumang:
@jdinan What about quiet being a requirement only for set but not for test?

manjugv:
July 23rd Meeting Notes: Special ballot did not pass.

BryantLam:
OpenSHMEM 1.4: "The shmem_clear_lock routine releases a lock previously set by shmem_set_lock *after ensuring* that all local and remote stores initiated in the critical region are complete."

This proposal: 
The \FUNC{shmem\_clear\_lock} routine releases a lock previously set by
\FUNC{shmem\_set\_lock} or \FUNC{shmem\_test\_lock} *after performing a
quiet operation on the default context to ensure* that all symmetric memory
accesses that occurred during the critical region are complete.

This proposal specifies `shmem_quiet` in `shmem_clear_lock`. @naveen-rn noted that the Cray implementation doesn't have a quiet in clear-lock (but the Cray manpage does specify local- and remote- store completion, but not `shmem_quiet`). He's going to look into it.

nspark:
@naveen-rn I think this PR primarily is clarifying the changes that arose when we added contexts. As of 1.4, the behavioral description for `shmem_clear_lock` says:

> The `shmem_clear_lock` routine releases a lock previously set by `shmem_set_lock` after ensuring
that all local and remote stores initiated in the critical region are complete.

This, I think, is roughly equivalent to a call to `shmem_quiet`. The implicit completion of local and remote stores, which was left ambiguous after we added context, is clarified to be `shmem_quiet`, which operates on the default context.

jdinan:
We will repeat the special ballot vote (add changelog entry - 45894b8) at the upcoming meeting.

naveen-rn:
> @naveen-rn I think this PR primarily is clarifying the changes that arose when we added contexts.

I got a bit confused during the reading - this change looks fine.

jdinan:
Special ballot changes: 45894b8 (changelog entry)

Issue #207, opened 2018-02-28
Waitsome / Testsome


davidozog:
The `shmem_wait_until` and `shmem_test` routines only operate on a single variable.  If users want to wait/test on multiple values, they could avoid the cost of multiple fences if the OpenSHMEM API had "waitsome" and "testsome" routines (analagous to [MPI_Testsome](https://www.mpich.org/static/docs/v3.1/www3/MPI_Testsome.html)/[MPI_Waitsome](https://www.mpich.org/static/docs/v3.1/www3/MPI_Waitsome.html)).

There is also the notion of waiting for "All" or "Any" condition(s), which involves blocking until satisfying _all_ conditions or a _single_ condition, respectively.  One question is whether OpenSHMEM would benefit from these routines, or whether "Some" is sufficient for implementing "All" or "Any" efficiently.

#### Proposed interface for "wait some" and "test some":
```c
void shmem_wait_until_some (TYPE *ivar, size_t nelems, size_t *count, size_t *indices, TYPE cmp, TYPE cmp_value);
int shmem_test_until_some (TYPE *ivar, size_t nelems, size_t *count, size_t *indices, TYPE cmp, TYPE cmp_value);
```

This topic is scheduled for discussion at the RMA WG on ~~March 1st~~ March 15, 2018.  
[Here are the slides](
https://docs.google.com/presentation/d/1vp1XWFE0E2JMUDA5lEdXngEhazETnhJRPwpwDmcqYTM/edit?usp=sharing). 
[ PDF Version here](https://github.com/openshmem-org/specification/files/1768792/RMA.WG.-.Test.2FWaitsome.pdf).

nspark:
Reviewing @davidozog's slides, I recall my main concern when I initially considered such a {wait,&nbsp;test} x {some,&nbsp;all,&nbsp;any} API.

In MPI -- and, of course, since I'm not MPI expert, please correct any errors -- these routines accept an array of opaque `MPI_Request` objects, each of which captures the state of some associated communication operation. These requests can be in one of three states: null, active, or inactive. For the discussion of this API, the main difference between MPI and SHMEM is that an `MPI_Request` object will only change state through local action of the MPI rank; e.g., from null to active via `MPI_Isend` and active to null via `MPI_Wait`.

[...]

PR #214, opened 2018-05-17
Add shmem_wait_until_some and shmem_test_some

davidozog:
This is a work-in-progress proposal to add `shmem_wait_until_some` and `shmem_test_some` routines to the specification.  This topic is is up for discussion at the May 21, 2018 specification meeting, where I hope to do an informal reading to gather feedback, but please don't hesitate to provide feedback on this before then.

Edit: This proposal now covers the complete `shmem_wait/test_until_any/some/all` family of functions.


davidozog:
This PR addresses issue #207

anshumang:
@davidozog is there a case for `cmp` and `cmp_value` to be arrays?

jamesaross:
The proposal for `shmem_test_some` appears to be a convenience function, replacing only a couple lines of code.  It seems to be to be only one instance of many different permutations of functionality built on `shmem_test`.  The example code was taken from [example_code/shmem_test_example1.c](https://github.com/openshmem-org/specification/blob/master/example_code/shmem_test_example1.c)  and changes the behavior of the `user_wait_any` routine.  In that example, the `user_wait_any` routine returns the index number of the first value tested true.

Is this code higher performance than the code below it?
```c
_Bool *status = calloc(npes, sizeof(_Bool));
size_t count = shmem_test_some(ivars, npes, status, SHMEM_CMP_NE, 0);
```
```c
size count = 0;
for (int pe = 0; pe < npes; pe++) count += shmem_test(ivars[pe], SHMEM_CMP_NE, 0);
```
Additionally, `_Bool` is C99 and there is no guarantee for C++ compatibility.  It's probably not a problem but it might be. 

The `shmem_wait_until_some` proposal is also a convenience function and just one instance of many different permutations built on `shmem_wait` or `shmem_wait_until`.

The shmem_wait_until_some_all2all_sum.c example code doesn't seem to be a very good use case of the routine because it loops through the flags data once for the `shmem_wait_until_some` and then again for the completed array.  Then the process repeats many times until there are no values left. Example code should strive to show good practice and this doesn't look right to me.

It's not clear to me that these methods would have any performance benefits (please help me understand why if they do).  Additional memory must be allocated/freed.  Instead of a function return value, an array of status values are written.  Then, as in the example, there's an additional loop over the completed array.  They are both one specialization of many number of permutations of functionality a user may want without saving much application code.

nspark:
> The proposal for `shmem_test_some` appears to be a convenience function, replacing only a couple lines of code.

IIRC, part of the motivation here came from a use-case informed by the weak memory model crowd. While stronger memory models may not show a difference in the code examples you shared, a call to `shmem_test` can require a memory fence that would be issued _N_ times in the manual loop vs. one time in `shmem_test_some`.

Perhaps @spotluri, @anshumang, or @shamisp can add more?

nspark:
> Additionally, `_Bool` is C99 and there is no guarantee for C++ compatibility. It's probably not a problem but it might be.

I agree that as we start adding C99+-specific features, we're losing compatibility with C++. While I don't mind requiring `stdbool.h`, IIRC, we were trying to avoid that. (Maybe it was just `complex.h` due to its macro definition of `I`.) If we require `stdbool.h`, we can use `bool` and preserve the appearance of compatibility with C++.

davidozog:
In SOS, for example, the `shmem_test_some` variant avoids the cost of several memory fences (`npes` of them in this case).  If it were `shmem_wait_until_some`, SOS may avoid additional overheads as well (like synching the NIC's atomics cache).  We've convinced ourselves in the past that these overheads are necessary...

I agree that it could become a slight problem with C++.  I didn't realize until now that C++ requires `<stdbool.h>` to define `_Bool`, but even worse is that apparently C++'s `bool` is not guaranteed to be compatible with C's `_Bool`, though it might in practice.  (Yes, this came from mimicking our previous solution to the `_Complex` issue).  I'd be fine with a change to `bool`, requiring `stdbool.h`, unless anybody opposes?

I hear what you're saying about the all-to-all example, but I personally prefer this more interesting task-processing use-case at the expense of slightly more complexity (as in the style of the shmem_ctx examples).  However, if there's overall agreement to simplify the example (perhaps similar to my original user_wait_all example  [here](https://github.com/davidozog/openshmem-specification/blob/wip/waitsome_testsome/example_code/shmem_wait_until_some_example.c)), I'd be happy to do that as well.

content/shmem_test_some.tex
```diff
@@ -0,0 +1,77 @@
...
+    section.} 
+  \apiargument{IN}{nelems}{The number of elements in the \VAR{ivars} array to be
+    compared with \VAR{cmp\_value}.}
+  \apiargument{INOUT}{status}{A mask array of length \VAR{nelems}, where each
+    element corresponds to the respective element in \VAR{ivars} and represents
+    the status of each implied wait condition.  On input, elements of \VAR{status}
+    set to 0 will be tested, and elements set to 1 will be ignored.  On output,
+    each element of \VAR{status} equal to 1 corresponds to a satisfied
+    condition in the respective element of \VAR{ivars} (or an ignored
+    condition, if initialized to 1 on input).}
```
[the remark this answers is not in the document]
davidozog: Yup, this is exactly the problem.  IIRC, some in the WG greatly preferred that `status` have a boolean type rather than some stranger alternative which supports the 3 states.  I do believe that committing to 0/1 `status` elements leads to two options: 1) adjust the `status` array each iteration or 2) keep track of a separate "completed" array.  I agree that neither is "the correct way", but I settled on #2 because I like the convenience of being able to reuse the `status` array, and I kinda think it's a little easier to follow.  But, yes, it's a bit of a wash...  I also agree with @jamesaross that the code example looks like bad practice at a glance.  Do you think Option #1 could improve that?

I'm thinking of re-writing the proposal with Option 1 to see how it looks.  Does that sound good?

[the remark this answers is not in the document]
nspark: I apologize for not taking a closer look at the proposal earlier in the process. I think that I expected a prototype (and behavioral function) more like:

```c
size_t shmem_test_some(TYPE *ivars, size_t nelems, size_t *satisfied,
                       int cmp, TYPE cmp_value)
{
  size_t count = 0;
  for (size_t i = 0; i < nelems; i++) {
    switch (cmp) {
    case SHMEM_CMP_EQ:
      if (ivars[i] == cmp_value)
        satisfied[count++] = i;
      break;
    /* case ... */
    }
  }
  return count;
}
```

This way, the call to `shmem_test_some` requires _O_(nelems) accesses to check the array of values, but the user of the call only needs _O_(satisfied) accesses.

Is this "better" than the current proposal (under either Option 1 or Option 2)? In some cases, I think it is because there are fewer memory references in user code. What this alternate interface lacks is the "mask" feature of the current proposal. To resolve this, a mask array could be provided as an extra argument to the function with a null-check fast path when the user wants to unconditionally check the entire array.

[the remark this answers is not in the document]
davidozog: Yeah, this is actually quite similar to what was presented in my first set of WG slides [here](https://docs.google.com/presentation/d/1vp1XWFE0E2JMUDA5lEdXngEhazETnhJRPwpwDmcqYTM/edit?usp=sharing), but somewhere along the line the `indices`/`satisfied` array morphed into the masked version, so that we could feed it back to the routine to better support task-processing-like patterns.  But yes, this is probably better, especially with the optional mask array idea, which I like.

jdinan:
> RE: stdbool

`stdbool.h` defines `true` and `false` macros, which could cause issues for some apps (this was the `I` issue with `stdcomplex.h`).
> Re: Memory fences in test

The OpenSHMEM spec requires the memory load fence to ensure that a subsequent load operation sees any values previously updated.  This is needed on x86, not just weaker memory models.  Bear in mind that with OpenSHMEM 1.4, the test/wait operation can be satisfied by a store operation performed by another thread in the same PE or by a store operation performed by another PE using the `shmem_ptr` interface.

davidozog:
I've updated this ticket to include the `indices` array, with semantics similar to @nspark's example above.  I've also made the `status` mask array optional.  This indeed simplifies the example quite a bit - no more need for those `completed` and `processed` arrays.

@nspark, @jamesaross, and @jdinan:  Since there is still a little time for last minute changes today, perhaps you can take a look?

One unresolved issue is with the type of `status`.  I tentatively changed the type from `bool` to `int *` to avoid the `true/false` macro problem with a dependence on `stdbool.h`.  Is this ok?  Perhaps `char *` or `unsigned char *` is better for limiting memory consumption?  Perhaps we should just use `bool` and deal with the consequence?

jamesaross:
At the moment, the new file `example_code/shmem_wait_until_some_example.c` doesn't seem to be used and doesn't follow the new proposal.

Respectfully, the example code `example_code/shmem_wait_until_some_all2all_sum.c` doesn't seem to make sense.  Has it executed on any platform?  It looks like it could either be caught in an infinite loop or fail the total sum check depending on the order in which RMAs occur.  I'm not sure what the code is trying to accomplish, but it doesn't seem like a pattern I would recommend for example code.

jamesaross:
Although the put-with-signal routines haven't yet been ratified, I believe a combination between the signalling routines and the wait_until_some routines would make a good pairing by using the signal as the wait flag.

nspark:
- I think we need to specify that `indices` must be an array of at least `nelems` elements -- OR -- provide an argument specifying the maximum number of indices that may be returned. (I prefer the former.)
- I don't really understand the clause:

  > After returning, the `indices` array contains the indices of some (but not necessarily all) of the elements in `ivars` that satisfied the test condition during the call to `shmem_test_some`.  The return value of `shmem_test_some` is equal to the total number of these satisfied elements.  If the return value is _N_, then the first _N_ elements of the `indices` array contain those unique indices that satisfied the test condition.

  In particular, under what condition does it not return the all of the indicies of the elements that satisfied the condition? I expect that a call to `shmem_test_some` starts at some unspecified index and tests all `nelems` elements of `ivars`.
- I was surprised that the `status` mask array has `INOUT` semantics. I guess I was expecting it to be a pure input (e.g., `const bool *`) and not _also_ provide output information on which wait variables satisfied the condition, which is redundant anyway. (To be clear: I hadn't envisioned that use-model, but if there is value to others, I won't oppose it.)
- I am not personally bothered by requiring `<stdbool.h>`. In modern C, I think macros for `true` and `false` are _far_ less common than the macro `I` so graciously provided by `<complex.h>`. I also am fine with `int*` or `char*` as the type of the `status` mask array. I can comfortably accept the argument for staying away from `_Bool` and `bool` due to the lack of C++ ABI compatibility.
- In the `shmem_wait_until_some_all2all_sum.c` example, this line is not valid C:
  https://github.com/openshmem-org/specification/blob/ad9f07f56a38c0078da54a2974e790375b01b051/example_code/shmem_wait_until_some_all2all_sum.c#L32-L33
  as the `while` loop accepts an expression, not a declaration. Also, beware of using an assignment as an expression; GCC and Clang both warn to wrap the assignment expression in parentheses to use as a truth value. I think you want:

  ```c
  int ncompleted;
  while ((ncompleted = shmem_wait_until_some(flags, npes, indices,
                                             status, SHMEM_CMP_NE, 0))) { ... }
  ```

jamesaross:
The proposal isn't clear about the value in `indices`.  Does it zero out previous values that passed?  Are the array indices in order?  The proposal needs more clarification on what happens with this array.

Will specializations of routines such as `shmem_wait_until_some` and `shmem_test_some` lead to rationalizing a proliferation of other minor specializations?  Should we now consider adding `shmem_wait_until_all` or `shmem_test_all` in order to reduce implicit memory fences?

nspark:
@jamesaross said:

> The proposal isn't clear about the value in `indices`. Does it zero out previous values that passed? Are the array indices in order? The proposal needs more clarification on what happens with this array.

Zero what values out? As currently written, it says that:

>If the return value is _N_, then the first _N_ elements of the `indices` array contain those unique indices that satisfied the test condition.

The first _N_ elements of the `indices` array contain an unspecified ordering of the indices of the `ivars` elements that satisfied the test condition. The values of `indices` at or beyond the _N_-th element are left unspecified, as they should be. (Why would we zero them out? That would be unnecessary memory accesses.)

The last time we discussed this on the RMA WG call, I think we agreed to leave the order of the indices that are used to check `ivars` unspecified. Per my previous comment, I think we need to be clear whether we're checking all `nelems` elements, but we decided not to specify an order for checking the elements.

davidozog:
One last thing: I would think exactly how many elements the implementation checks with the "some" semantic is kinda up to the implementation.  It that's a deal-breaker, then I think it's fine to specify that all `nelems` are checked (unless setting `status` of course).  Is that what you're looking for @nspark?

nspark:
> One instance `shmem_test_some` may not return all of the satisfied indices is if an condition was not satisfied until immediately (or shortly) after having been checked by the implementation. [...]

I see what you mean: after `shmem_test_some` checks `ivars[i]`, it may be updated by another PE. I think we should improve the wording, though. It sounds somewhat like the function is saying, "I'll check all _N_ items, and I might find _M_ satisfied states, but I can report _P_ <= _M_," which just seemed bizarre.

How about something like...?

> The `shmem_test_some` routine tests the numeric comparison of each **of the `nelems` elements in the array of** symmetric object**s** pointed to by `ivars` with the value `cmp_value` according to the comparison operator `cmp` at the calling PE. **The order in which these elements are tested is unspecified.**
>
> ~~After returning~~ **Upon return**, the `indices` array contains the indices of **all** ~~of some (but not necessarily all)~~ of the elements in `ivars` that, **when tested,** satisfied the test condition of the call to `shmem_test_some`. The return value of `shmem_test_some` is equal to the total number of these satisfied elements.  If the return value is _N_, then the first _N_ elements of the `indices` array contain those unique indices that satisfied the test condition. **The array pointed to by `indices` must be at least `nelems` long.**

-------

> On that note, perhaps it does make sense to add that length argument to `indices`... If you pass 1, then you get "test_any" behavior, if you pass `nelems`, you get "test_all" behavior...  But if you want "some" then I'm not so sure what you would pass...

As you note, given this hypothetical `numindices`, it seems well-defined when `numindices` equals 1 or `nelems`. Anything in-between seems odd, in particular because we don't provide a way to control where the next round of testing starts. If we want to support testing a 100-long array and retrieving batches of five satisfied indices at a time, then I think we also need a mechanism to specify where to start on the next call. (I'd prefer not to, since this feels like bloat, especially since if a user wants to check one element, we already have `shmem_test` and the user controls which element to test.)

I would expect that the test is a relatively fast operation, but if one wanted to run it in batches over a larger array, then I think this is a case for using the `status` array. The `status` array complicates things a bit&mdash;at least in terms of what to write in the behavioral description&mdash;because now `indices` only really needs to be as long as the number of indices that are requested to be tested.

> One last thing: I would think exactly how many elements the implementation checks with the "some" semantic is kinda up to the implementation.  It that's a deal-breaker, then I think it's fine to specify that all `nelems` are checked (unless setting `status` of course).  Is that what you're looking for @nspark?

Yeah, I don't think we should allow the `some` semantic. Should this function then be called `shmem_test_all`?

davidozog:
Yeah, I like those changes.  And I presume we need similar changes to the `wait_until_some` text?   I'm not sure how I feel about "... are waited upon..." replacing "are tested" throughout, but I can't think of any better phrasing...

I don't think `shmem_test_all` is necessary.  If it were like MPI TestAll, the return would return zero unless *all* tests are satisfied.  I'd say we're still in "some" territory, right?

jdinan:
I like the idea of wait-some with an argument to specify the maximum "some"; this broadens the set of usage models, including supporting wait-any.  For example, a user might have a sense for how much they can profitably pipeline and could use the "some" value as a pipelining knob.

One design point in between test-some and test-all (or perhaps an alternative "some" definition) is a test-this-many routine that tests until it has either tested all ivars or filled the output array (of "this-many" size).

jamesaross:
The routines permit a null pointer for `status`, preventing unnecessary memory allocation and access.  What about accepting a null pointer for `indices`? The implication might be that it would wait until all or test all.

davidozog:
FYI, it seems like I'm making more last minute changes than I expected on this ticket.  I will see what I can iron out tonight, but at this point it seems like an informal reading might be a little more appropriate at the F2F.

nspark:
Do we need to add `restrict` type qualifiers for `ivars`, `indices`, and `status`? While it's not uncommon for developers to throw `-fno-strict-aliasing`, even with the strict aliasing semantics of C, a compiler may consider `ivars` to alias `indices` when `TYPE` is `size_t` or to alias `status` when `TYPE` is `int`.

content/shmem_test_some.tex
```diff
@@ -0,0 +1,85 @@
...
+
+    Upon return, the \VAR{indices} array contains the indices of the elements
+    in the test set that satisfied the test condition during the call to
+    \FUNC{shmem\_test\_some}.  The return value of \FUNC{shmem\_test\_some} is
+    equal to the total number of these satisfied elements.  If the return value
+    is $N$, then the first $N$ elements of the \VAR{indices} array contain
+    those unique indices that satisfied the test condition.  The array pointed
+    to by indices must be at least \VAR{nelems} long.
+
+    The optional \VAR{status} array passed to \FUNC{shmem\_test\_some} has the
```
agrippa: Even though the rules around status are explained in shmem_wait_until_some, my personal preference would be to have them repeated in full here so that this API definition is self-contained.

davidozog: @grippa Yeah, I understand.  It's funny - I've also received the exact opposite critique, which is how this shorter version came to be :wink:.  Perhaps we can defer this decision to the doc edit process, since copy/pasting the description of `status` wouldn't change any semantics?

davidozog:
With the addition of the wait and test "**Any**" routines, a small issue has come up where I would appreciate any feedback you might have.  If we commit to returning the value of the index that satisfies the condition, then `0` is a valid return value.  If we also commit to returning `0` when no elements satisfy the condition, then (in my opinion) this results in a slightly weird usage model.  Here's a simple example of a linear barrier with overlap using `shmem_test_any`:
```c
...
int mype = shmem_my_pe();
int npes = shmem_n_pes();
 
int *flags = shmem_calloc(npes, sizeof(int));
int *status = calloc(npes, sizeof(int));
int *processed = calloc(npes, sizeof(int));
 
for (int i = 0; i < npes; i++)
    shmem_p(&flags[mype], 1, i);
 
int ncompleted = 0, completed_idx;
 
while (ncompleted < npes) {
    completed_idx = shmem_test_any(flags, npes, status, SHMEM_CMP_EQ, 1);
    if (!processed[completed_idx] && status[completed_idx]) {
        ncompleted++;
        processed[completed_idx] = 1;
    } else {
        /* Overlap some computation here */
    }
}
```
Note that the check on `status[completed_idx]` is only necessary for the zeroth index... but checking it for all indices might be a little more concise than adding an `else if` branch to handle the zeroth index.

Any thoughts here?  I hesitate to change the return value semantic and add an output argument to the routine or something like that...  Maybe the current design isn't so bad?

Side note:  I've also added a fairness guarantee to the "Any" and "Some" flavors that assures code like this will actually work.  For example: "If an entry _i_ in `ivars` within the test set satisfies the condition, then a series of calls to `shmem_test_any` must eventually return _i_."

davidozog:
@nspark Good thought, that would certainly work - I hadn't considered that.  

Do you prefer this to the other options?  As I see it, we have:
1) return 0
2) return SIZE_MAX
3) add a `size_t * index` output argument and return whatever

I'm partial to 2 or 3, but I might have to think about it a bit and consider what return value would be most useful for 3.  Maybe 1 on success, 0 on failure?

Here are the loop bodies for that example:

Option 1:
```c
index = shmem_test_any(flags, npes, status, SHMEM_CMP_EQ, 1);
if (!processed[index] && status[index]) {
    ncompleted++;
    processed[index] = 1;
} else { /* Overlap */ }
```
Option 2:
```c
index = shmem_test_any(flags, npes, status, SHMEM_CMP_EQ, 1);
if (index != SIZE_MAX && !processed[index]) {
...
```
 Option 3:
```c
completed = shmem_test_any(flags, npes, &index, status, SHMEM_CMP_EQ, 1);
if (completed && !processed[index]) {
...
```

nspark:
I definitely prefer Options 2 and 3 over 1. I think I slightly prefer Option 2 over Option 3. (As it was my suggestion, I'm hardly unbiased.) My feelings on Option 3 are mainly that six arguments just feels like a lot. That's a weak argument, though.

davidozog:
Some comments from the RMA WG about the `wait_any`/`test_any` return value:

* Proposal needs to specify that `nelems < SIZE_MAX`.
* Could return a value greater than or equal to `nelems` instead of `SIZE_MAX`.
* A 4th option is to use a type like `ssize_t` and reserve -1 for the none-completed case.

We'll continue discussion at an RMA WG meeting next week (because Oct. 22nd is deadline for changes for the next committee meeting).

davidozog:
While testing my [prototype here](https://github.com/davidozog/sandia-shmem/tree/wip/wait_test_all_any_some), I found that [C++ doesn't recognize `restrict`](https://stackoverflow.com/questions/22724980/why-is-the-restrict-keyword-not-part-of-c).

@jdinan, @nspark, et al. - any thoughts on how to proceed?  Perhaps we could have a separate C++ API without `restrict`?

davidozog:
Here's that straw poll I mentioned in the RMA WG:
https://github.com/davidozog/openshmem-specification/issues/1

davidozog:
Here is a diff for the upcoming special ballot:
https://github.com/openshmem-org/specification/compare/691aff7..724bd6f

As of now, the [straw poll above](https://github.com/davidozog/openshmem-specification/issues/1) shows most votes for `int` as the `status` array type with 9 out of 33 total votes (a pretty good turnout!).  I'll have to cutoff the poll tomorrow night so I have time to make the change if needed.  For those following this ticket, please cast your vote and/or let me know if you're comfortable with `int` for the special ballot.

manjugv:
January 2019 Meeting: Special ballot passed; Voting on the proposal postponed

minsii:
@davidozog @jdinan : As we discussed at today's spec meeting, I would like to see an example that indicates why it is difficult to only use `indices` as the output of `shmem_wait_until_some`.

Another question is about the statement "The \VAR{ivars} and \VAR{status} arrays must not overlap in memory." Since `ivars` is an array of remotely accessible data object, why do we need to clarify that it can never be overlapped with the local memory array `status` ?

minsii:
@davidozog Thanks for the detailed explanation. My proposed change is a little different. To be clear:

Current APIs:
```
size_t shmem_test_some(TYPE * ivars, size_t nelems, size_t * indices, int * status, int cmp, TYPE cmp_value);
OUT indices
INOUT status

size_t shmem_test_any(TYPE * ivars, size_t nelems, int * status, int cmp, TYPE cmp_value);
INOUT status
```

Proposed change:
```
shmem_test_some: 
OUT indices
IN status

shmem_test_any:
IN status
```

Thus, the `IN status` parameter will be used to only describe whether user wants to check any of the elements in `ivars` array. The `OUT indices` in `shmem_test_some` and return value of `shmem_test_some` indicate the element(s) in the test set that satisﬁed the test condition in the current call.

This change can avoid unnecessary update to the `status` array inside the shmem call, if the user does not need it. Otherwise, the user can just update the `status` array in the user program. Moreover, if there is no specific reason to make `status` as `INOUT`, updating both `status` and `indices|return value` looks redundant to me.

I agree that the internal update overhead should be small, but since we are defining the API now, I think we should try to avoid any unnecessary overhead/redundancy  .

minsii:
@tonycurtis @davidozog The reason I feel a little odd when reading the overlap description is that, we do not have similar statement for other APIs (i.e., one array is symm object, and the other is likely a local array). Maybe we need to clean up the overlap statement globally ? I am not sure if any other APIs have this limitation.

davidozog:
If I recall correctly, the committee/WG has discussed transitioning towards including such a statement about overlap wherever necessary.  For now, we've introduced it in this PR as well as the put-with signal proposals, as something like a reminder.  For example [here](https://github.com/openshmem-org/specification/pull/218/files#diff-52e7c9c716aec88f1ace7dbd96ef3dddR69) and [here](https://github.com/openshmem-org/specification/pull/244/files#diff-fcca2f4cb166074d48841ada1e2a5699R72).

I don't believe the overlap statement is needed as pervasively as you might think.  Throughout the RMA routines for instance, overlapping `source` and `dest` in `shmem_put/get` is legal, right?  I believe the same might be true for the collectives buffers.  And AMOs don't have much opportunity for overlap that I can tell.  I could be wrong though, what do you think @minsii?

minsii:
@davidozog : The overlap statement in put-with-signal is slightly different, as both `dest` and `sig_addr` buffers are symmetric data object. So, I do not think the overlap statement is required there. In this PR, although I still feel it is a little odd to add the overlap statement, I have no objection to adding it if we think it is better than none.

minsii:
@davidozog Here is some update about the `status` issue in `wait|test_until_some|any` after discussed with other Argonne staff. Sorry for the late update.

First, we found a client-server usecase that we want to keep `status` as a constant mask. This scenario can be found in the genome assembly application.
1. Suppose the server PE maintains an `ivars[N]` array to check the query from multiple client PEs, where `N` is the number of clients.
2. The sever spawns `X` threads, each of the thread maintains the query from different group of the clients to avoid contention (e.g., T0 checks ivars[0-1], T1 checks ivars[2-3], if N=4, and X=2).
3. Once `ivars[i]` is updated, the same element might be updated again by client `PE i` for the next query, thus the `status` mask used by each thread should be constant.
4. After a period of time, the server may want to reassign the `ivars` partition to each thread, e.g., for load balancing purpose. The assignment pattern can be any format (e.g., round-robin: T0 handles ivars[0,2] and T1 handles ivars[1,3]).

In the above case, the reason why we do not want to replace the constant `status` mask with separate `ivars` arrays is that, the allocation overhead of `ivars` is expensive (i.e., requires all PEs to collectively allocate). But you have to reallocate the `ivars` array at step-4 if each thread uses different `ivars` array. Note that this problem does not exist in MPI, because the MPI `request` array is only a local variable, so the array recreation is cheap.

Second, about the API definition options (whether `IN status` or `INOUT status`), it is very hard to choose one of them, as the former is more flexible, but the latter is more convenient for common usecases. Actually, I still prefer the `IN status` option, but some others prefer `INOUT status`.

davidozog:
Thanks @minsii - that's a very interesting use case.  I think this is similar to our previous discussion about processing "boundary" values in a halo exchange, where a constant `status` mask would be useful.

So are you interested in `status` being "IN" or "INOUT" depending on the user's preference?  For example, would the following work?

- On input, elements of `status` set to 0 are included in the wait set, and are updated to 1 if the corresponding `ivars` value meets the condition (same as current proposal).
- On input, elements of `status` set to -1 (or some other macro/designated value) are included in the wait set and are **not** updated if the corresponding `ivars` value meets the condition.

I hesitate to make the API more complicated, but I'm curious if this is a change that would resolve the issue.  I'm also curious if rather make `status` an "IN" and remove the "INOUT" behavior.

Let's discuss this on the RMA WG call today.

davidozog:
Here is a summary of the latest changes.  There are 3 viable options with regard to the IN/OUT property of the `status` array:

1. Status is INOUT
2. Status is IN only
3. Status is either IN or INOUT, depending on the input values (see commit [9297080](https://github.com/openshmem-org/specification/pull/214/commits/9297080c6916f8c4fd1fa756d3bda3f06f2bdffc)).

There are several pros/cons across these options, but briefly:

1. Status is INOUT
    - Pros: Using `status` does not require updating the array after wait/test calls.
    - Cons: Some use cases may not require updating `status`, so it's unnecessary for the implementation to always do the write operations.

2. Status is IN only
    - Pros: This has the simplest API description with no major loss in functionality.  
    - Cons: Requires an extra line of code in the examples.

3. Status is either IN or INOUT
    - Pros: Best of both worlds in many respects.
    - Cons: Has a more complex API description, need new examples for coverage, places more requirements on implementations.

After some pondering and offline discussion, we've settled on #2 because its the simplest option and is still sufficiently versatile.

I plan on reading these changes at the upcoming meeting as a special ballot.  If you have any concerns about this or the changes themselves, please try to let me know before the end of the week.

davidozog:
Special ballot diff:
[724bd6f..d2b3405](https://github.com/davidozog/openshmem-specification/compare/724bd6f..d2b3405)

content/shmem_test_all.tex
```diff
@@ -0,0 +1,63 @@
...
+\begin{apiarguments}
+
+  \apiargument{IN}{ivars}{A pointer to an array of remotely accessible data
+    objects.}
+  \apiargument{IN}{nelems}{The number of elements in the \VAR{ivars} array.}
+  \apiargument{IN}{status}{An optional mask array of length \VAR{nelems}
+    that indicates which elements in \VAR{ivars} are excluded from the test set.}
+  \apiargument{IN}{cmp}{A comparison operator from Table~\ref{p2p-consts}
+    that compares elements of \VAR{ivars} with \VAR{cmp\_value}.}
+  \apiargument{IN}{cmp\_value}{The value to be compared with the objects
```
agrippa: I can see a variant of this API where we can provide a cmp_value per element as being useful.

That is, cmp_value would be a TYPE* with one value per element in ivars.

This would up the memory consumption, but in exchange for more flexibility.

I'm assuming you would agree that if we wanted to add that, it should be included in a separate follow on proposal?

content/shmem_test_all.tex
```diff
@@ -0,0 +1,63 @@
...
+    pointed to by \VAR{ivars}.}
+
+\end{apiarguments}
+
+\apidescription{
+    The \FUNC{shmem\_test\_all} routine behaves similarly to
+    \FUNC{shmem\_wait\_until\_all}, but it does not block and returns zero if
```
agrippa: I don't love cross-API references. As someone who uses the spec as an API reference, it would be annoying if I were trying to refresh my memory on shmem_test_all to then have to jump over to shmem_wait_until_all and read that, only to have to jump back. I'd much prefer shmem_test_all's description be self-contained as much as possible, even at the cost of redundant text with shmem_wait_until_all. Any objections?

agrippa: It also makes the spec harder to read as a new user.

content/shmem_test_all.tex
```diff
@@ -0,0 +1,63 @@
...
+    not all entries in \VAR{ivars} satisfied the test condition.  This routine
+    compares each of the \VAR{nelems} elements in the \VAR{ivars} array with
+    the value \VAR{cmp\_value} according to the comparison operator \VAR{cmp}
+    at the calling PE.
+    If \VAR{nelems} is 0, the wait set is empty and this routine returns 1.
+
+    The optional \VAR{status} array passed to \FUNC{shmem\_test\_all} has the
+    same behavior as in \FUNC{shmem\_wait\_until\_all} in
```
agrippa: Same comment as above.

content/shmem_test_any.tex
```diff
@@ -0,0 +1,78 @@
...
+    no entries in \VAR{ivars} satisfied the test condition.  This routine
+    compares each of the \VAR{nelems} elements in the \VAR{ivars} array with
+    the value \VAR{cmp\_value} according to the comparison operator \VAR{cmp}
+    at the calling PE.  The order in which these elements are tested is
+    unspecified.  If an entry $i$ in \VAR{ivars} within the test set satisfies
+    the test condition, a series of calls to \FUNC{shmem\_test\_any} must
+    eventually return $i$.
+
+    The optional \VAR{status} array passed to \FUNC{shmem\_test\_any} has the
+    same behavior as in \FUNC{shmem\_wait\_until\_any} in
```
agrippa: Another cross-API reference.

content/shmem_test_some.tex
agrippa: Heh looks like I made this same comment previously. I'm still in favor of my original opinion.

content/shmem_test_all.tex
davidozog: I believe I have heard some objections to full self-containment in the past, but perhaps not...  Might as well try taking a poll!
https://doodle.com/poll/q6chkaqv76psrttn

I'm actually slightly in favor of removing the cross-API references as well, but I would prefer to make the changes in a separate ticket.  Would that be acceptable?  If so, I can start preparing this before we do the voting on this ticket.

content/shmem_test_all.tex
davidozog: Yes - this has come up before, and it may have also been under the guise of the `cmp` operator as well as `cmp_value`.  But I imagine more relevant use-cases for `cmp_value`; for instance, in a work-sharing app where each `ivars` element has a corresponding (and varying) number of expected tasks.

In the past the consensus was to avoid the memory cost until we're certain that it's a useful feature.  Perhaps there's a clean way to make it optional, but I imagine its at least another argument to an already long list of arguments. 

I'd be happy to add this in if that's the actual consensus ([how about another Doodle poll?](https://doodle.com/poll/vfqvxfn59tnkz4hn)).  And yes, a follow on proposal is preferred if that's acceptable.

content/shmem_test_all.tex
nspark: I think cross references make more sense if one thing is defined in terms of extending the behavior of another. This is the reverse scenario, in which one thing is defined in terms of something else, but not exactly all that it does, which only serves to be confusing.

Specifically, the cross reference here would make more sense if `shmem_wait_until_all` was defined in terms of `shmem_test_all`, not vice versa.

content/shmem_test_all.tex
nspark: I don't understand the poll. There's no option to keep the API as currently proposed (i.e., `cmp` is `int` and `cmp_value` is `TYPE`); the "other" option is "it depends / don't care". I think such a change should be should be in another proposal.

I think making `cmp` and `cmp_value` both `nelems`-long arrays would add far too much memory pressure for this routine to be effective in the cases I have in mind. (It's seriously making me think of struct-of-arrays vs. array-of-structs layouts and the cache behavior implied here.)

content/shmem_test_all.tex
davidozog: I've always wondered why `shmem_test` was defined after `shmem_wait` in the spec...  Perhaps it's easier to absorb conceptually in that order as a user, but I'd certainly be open to switching.  As a developer, `shmem_wait` is easily implemented using `shmem_test` so why not define test first?  Or we could just not worry about it and embrace self-containment ;)

nspark: I think it's because `shmem_wait[_until]` predates OpenSHMEM 1.0, but `shmem_test` wasn't added until OpenSHMEM 1.4.

content/shmem_test_all.tex
davidozog: Ah, thanks @nspark.  I tried to clarify the poll with these options for which direction to take the API:

1) Only `cmp_value` is an array (of length `nelems`)
2) Only `cmp` is an array
3) Both `cmp_value` and `cmp` are arrays
4) Neither `cmp_value` and `cmp` are arrays (_as currently proposed_, length of 1 each)
5) Explore other options, like making 1 and/or 2 optional.

And someone please let me know if I'm a little too poll-happy ;)

davidozog: And yes, the poll is intended to guide a (possible) separate proposal.

content/shmem_test_all.tex
davidozog: Wow, somehow I didn't catch that `shmem_test` was new in 1.4...  It must have been proposed relatively early on.  Thanks.

content/shmem_test_all.tex
```diff
@@ -5,13 +5,14 @@
 \begin{apidefinition}
 
 \begin{C11synopsis}
-int @\FuncDecl{shmem\_test\_all}@(TYPE *ivars, size_t nelems, int cmp, TYPE cmp_value);
+int @\FuncDecl{shmem\_test\_all}@(TYPE *ivars, size_t nelems, const int *status, int cmp, TYPE cmp_value);
 \end{C11synopsis}
```
minsii: Apologize for my late question. I am not sure why is the `status` argument is also added for `test_all` and `wait_until_all`. Could you please explain the reason ?

davidozog: Sure - this came up during our discussion about whether `status` should be an "IN" only; but admittedly, the change to add `status` to the "all" routines is a little tangential to the IN/OUT issue.  Anyway, the relevant use case occurs when a user wants to call wait/test on a constant subset of _all_ values in `ivars`.  An example application might be communicating boundary cell data across a regular mesh, where each PE owns some part of the mesh, and that part of the mesh consists of `ivars`.  Every iteration the relevant boundary cells (a subset within `ivars`) stay the same (which a good argument for IN-only `status`) and the user wants to wait on _all_ the elements specified by `status`.  This sort of application does not fit the "some" idiom with `status`, because "some" would only indicate whether _at least one_ element satisfies the condition.  In apps like this, you often can't continue until all the boundary cells are updated (like in particle-in-cell apps, to be more specific).

It also makes the interface a little bit more consistent, which is nice.

minsii: Thanks, it makes sense to me.

jdinan:
Need doc edit approval from two committee members for the following patches: 05eed57, 5ca83e4, 
206b23d, 02ee56b.  Please respond with approval or +1 (:+1:) to approve.

jdinan:
@davidozog Could you please merge your docedits for review?  Don't worry about resolving the conflict in the changelog; I'll take care of that when I merge.  Thanks!

davidozog:
@jdinan, thanks for the reminder.

Commit https://github.com/openshmem-org/specification/pull/214/commits/347ffedfc0fe0a4a4d79f8c00fc6798d1a56d00f removes the cross-referencing noted by @agrippa and commit https://github.com/openshmem-org/specification/pull/214/commits/4196a3de3a9583f1de23c0ccfccc011ba06d9143 fixes a couple of typos I noticed.  Please note that the `SIZE_MAX` fix in https://github.com/openshmem-org/specification/pull/214/commits/4196a3de3a9583f1de23c0ccfccc011ba06d9143 might be considered a semantic change, but hopefully it's an acceptable doc edit, because we know that this routine needs to return `void`...

jdinan:
I would like to request committee review for the additional DocEdits in 347ffed and 4196a3d.  Please :+1: this comment or leave feedback here.  Thanks!

jdinan:
@minsii and @agrippa Could you please review the additional DocEdits mentioned above (347ffed and 4196a3d)?  Thanks!

PR #215, opened 2018-05-17
Remove interoperability annex

### LINKED ISSUE #191: Purge/Fix Appendix D
## Proposed Change

Either fix or remove Appendix D from specification document.  It adds nothing (well, perhaps confusion) to the specification and is slowly going out-of-date.

## Context

As noted by @jeffhammond:

Appendix D references SGI, which is renamed to HPE elsewhere.  Obviously, this is of no practical consequence, but I guess it should be fixed in the next version.

I'm also not clear what purpose Appendix D actually serves.  I'd much rather see us define something useful, such as preprocessor symbols the implementation can define to indicate whether or not SHMEM PEs are equivalent to MPI processes (i.e. ranks in MPI_COMM_WORLD), UPC threads, or any other unit of parallel execution that we think matters.

Finally, I'd like to see "the MPI rank within the MPI_COMM_WORLD environment variable" stricken or corrected, because that phrase makes no sense.

jdinan:
Annex D is slowly bit rotting and contains information that is duplicated from elsewhere in the spec. 
 Per discussion on #191, the proposed solution is to remove this annex.

Closes #191

### DIFF
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -215,51 +215,6 @@ The following cases summarize this behavior:
 
 
 
-\chapter{Interoperability with other Programming Models}\label{sec:mpi}
-
-\section{\ac{MPI} Interoperability}
-
-\begin{sloppypar} % to prevent constants from running into margins.
-%
-\openshmem routines may be used in conjunction with \ac{MPI} routines in the
-same program.  For example, on \ac{SGI} systems, programs that use both \ac{MPI} and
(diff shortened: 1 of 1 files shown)


jdinan:
Per @minsii's recommendation, we will proceed with a vote on this PR.  She will prepare a future proposal to introduce a new interoperability annex (additional details/discussion on #191).

Issue #206, opened 2018-02-28
shmem_put_signal


bcernohous:
Discuss a proposal for shmem_put_signal.

Transfers contiguous data from a local data object to a specified processing element (PE) without blocking the caller and sets a remote flag to signal completion.

Cray's current implementation:

```c
void shmem_put_signal[_nb](void *target, const void *source, size_t len, uint64_t *sig_target, uint64_t sig_val, int pe)
```

Semantically (almost) equivalent to :

- `shmem_put[_nb](data)`
- `shmem_fence()`
- `shmem_put[_nb](signal)`

except a `shmem_fence` is 'heavier' than the `shmem_put_signal` which only fences the data before the signal.

- The target buffer and signal location must reside in symmetric memory.
- All data size and type variants are supported.
- signal is always a `uint64_t`.

Attaching slides for RMA WG 03 01 2018
https://github.com/openshmem-org/specification/wiki/RMA-WG-03-01-2018

[OpenSHMEM_RMA_PUTSIGNAL.pptx](https://github.com/openshmem-org/specification/files/1768534/OpenSHMEM_RMA_PUTSIGNAL.pptx)

**Edit** (@nspark): Fixed some Markdown formatting issues.

naveen-rn:
I just realized, after today's OpenSHMEM spec meeting and @jdinan Active set changes discussion. If we didn't discuss about the put_with_signal with put size 0, we should have a quick discussion on it.

nspark:
I think that a `shmem_put_signal` call with a size of 0 should still send the signal update.

naveen-rn:
> I think that a shmem_put_signal call with a size of 0 should still send the signal update.

+1
Cray SHMEM currently sends a signal even for a 0 size puts. I'm OK with this semantics.

jdinan:
@bcernohous I talked with @shefty about implementation and he indicated that `FI_FENCE` does not guarantee that data written at the target will be visible in memory prior to that of the fenced operation (assuming here that the signal is sent with the FI_FENCE flag).  Sounds like this may not be the right way to implement put with signal on OFI.

bcernohous:
Why not?  Does FI_FENCE only guarantee FI_TRANSMIT_COMPLETE?  It doesn't specify that.  If you are using FI_DELIVERY_COMPLETE on the operations?  If FI_DELIVERY_COMPLETE doesn't guarantee visibility then nothing does and it's not really a FI_FENCE issue.


[...]

PR #218, opened 2018-05-25
Add support for put with signal operation

naveen-rn:
This is a work-in-progress proposal to add put with signal operations into the OpenSHMEM standards.
This is related to issue #206. Proposed feature is available as part of Cray  SHMEM implementation.

### DIFF
--- a/content/p2p_sync_intro.tex
+++ b/content/p2p_sync_intro.tex
@@ -3,8 +3,7 @@ for synchronization between two \acp{PE} based on the value of a symmetric data
 object.
 The point-to-point synchronization routines can be used to portably ensure
 that memory access operations observe remote updates in the order enforced by
-the initiator \ac{PE} using the put-with-signal(refer
-section~\ref{subsec:shmem_put_signal}, \FUNC{shmem\_fence} and
+the initiator \ac{PE} using the put-with-signal, \FUNC{shmem\_fence} and
 \FUNC{shmem\_quiet} routines.
 
 Where appropriate compiler support is available, \openshmem provides
--- a/content/p2p_sync_intro.tex
+++ b/content/p2p_sync_intro.tex
@@ -3,8 +3,9 @@ for synchronization between two \acp{PE} based on the value of a symmetric data
 object.
 The point-to-point synchronization routines can be used to portably ensure
 that memory access operations observe remote updates in the order enforced by
-the initiator \ac{PE} using the \FUNC{shmem\_fence} and \FUNC{shmem\_quiet}
-routines.
+the initiator \ac{PE} using the put-with-signal(refer
+section~\ref{subsec:shmem_put_signal}, \FUNC{shmem\_fence} and
+\FUNC{shmem\_quiet} routines.
 
 Where appropriate compiler support is available, \openshmem provides
 type-generic point-to-point synchronization interfaces via \Cstd[11] generic
--- a/content/shmem_put_signal.tex
+++ b/content/shmem_put_signal.tex
@@ -1,0 +1,98 @@
+\color{Green}
+\apisummary{
+    The put with signal routines provide a method for copying data from a
+    contiguous local data object to a data object on a specified \ac{PE}
+    and set a remote flag to signal completion.
+}
+
+\begin{apidefinition}
+
+\begin{C11synopsis}
+void @\FuncDecl{shmem\_put\_signal}@(TYPE *dest, const TYPE *source, size_t nelems, uint64_t *sig_addr, uint64_t signal, int pe);
+void @\FuncDecl{shmem\_put\_signal}@(shmem_ctx_t ctx, TYPE *dest, const TYPE *source, size_t nelems, uint64_t *sig_addr, uint64_t signal, int pe);
+\end{C11synopsis}
+where \TYPE{} is one of the standard \ac{RMA} types specified by Table \ref{stdrmatypes}.
+
+\begin{Csynopsis}
+void @\FuncDecl{shmem\_\FuncParam{TYPENAME}\_put\_signal}@(TYPE *dest, const TYPE *source, size_t nelems, uint64_t *sig_addr, uint64_t signal, int pe);
+void @\FuncDecl{shmem\_ctx\_\FuncParam{TYPENAME}\_put\_signal}@(shmem_ctx_t ctx, TYPE *dest, const TYPE *source, size_t nelems, uint64_t *sig_addr, uint64_t signal, int pe);
+\end{Csynopsis}
+where \TYPE{} is one of the standard \ac{RMA} types and has a corresponding \TYPENAME{} specified by Table \ref{stdrmatypes}.
+
+\begin{CsynopsisCol}
+void @\FuncDecl{shmem\_put\FuncParam{SIZE}\_signal}@(void *dest, const void *source, size_t nelems, uint64_t *sig_addr, uint64_t signal, int pe);
+void @\FuncDecl{shmem\_ctx\_put\FuncParam{SIZE}\_signal}@(shmem_ctx_t ctx, void *dest, const void *source, size_t nelems, uint64_t *sig_addr, uint64_t signal, int pe);
+\end{CsynopsisCol}
+where \SIZE{} is one of \CONST{8, 16, 32, 64, 128}.
+
+\begin{CsynopsisCol}
+void @\FuncDecl{shmem\_putmem\_signal}@(void *dest, const void *source, size_t nelems, uint64_t *sig_addr, uint64_t signal, int pe);
+void @\FuncDecl{shmem\_ctx\_putmem\_signal}@(shmem_ctx_t ctx, void *dest, const void *source, size_t nelems, uint64_t *sig_addr, uint64_t signal, int pe);
+\end{CsynopsisCol}
+
+\begin{apiarguments}
+    \apiargument{IN}{ctx}{The context on which to perform the operation.
+      When this argument is not provided, the operation is performed on
+      \CONST{SHMEM\_CTX\_DEFAULT}.}
+    \apiargument{OUT}{dest}{Data object to be updated on the remote \ac{PE}. This
+    data object must be remotely accessible.}
+    \apiargument{IN}{source}{Data object containing the data to be copied.}
+    \apiargument{IN}{nelems}{Number of elements in the \VAR{dest} and \VAR{source}
+    arrays. \VAR{nelems} must be of type \VAR{size\_t} for \Cstd. When using
+    \Fortran, it must be a constant, variable, or array element of default
+    integer type.}
+    \apiargument{OUT}{sig\_addr}{Signal data object to be updated on the remote
+    \ac{PE} to be updated as the signal. This signal data object must be
+    remotely accessible and it can be in the same or differnt memory segment
+    as the \VAR{dest} data object.}
+    \apiargument{IN}{signal}{Unsigned 64-bit value used to set the remote
+    \VAR{sig\_addr} signal data object.}
+    \apiargument{IN}{pe}{\ac{PE} number of the remote \ac{PE}. \VAR{pe} must be
+    of type integer. When using \Fortran, it must be a constant, variable,
+    or array element of default integer type.}
+\end{apiarguments}
+
+\apidescription{
+    The routines return after the data has been copied out of the \source{}
+    array on the local \ac{PE}. The delivery of \signal flag on the remote
+    \ac{PE} guarantees the delivery of data words into the data object on the
+    remote \ac{PE}. Furthermore, two successive put with signal routines or
+    a successive put followed by a put with signal routine may deliver data
+    out of order unless a call to \FUNC{shmem\_fence} is introduced between
+    the two calls and the delivery of the \signal flag on the remote \ac{PE}
+    guarantees only the delivery of its corresponding data object on the
+    remote \ac{PE}.
+ }
+
+\apidesctable{
+    The \dest{} and \source{} data objects must conform to certain typing
+    constraints, which are as follows:}
+    {Routine}{Data type of \VAR{dest} and \VAR{source}}
+    \apitablerow{shmem\_putmem}{Any data type. nelems is scaled in bytes.}
+    \apitablerow{shmem\_put8}{Any noncharacter type that
+        has a storage size equal to \CONST{8} bits.}
+    \apitablerow{shmem\_put16}{Any noncharacter type that
+        has a storage size equal to \CONST{16} bits.}
+    \apitablerow{shmem\_put32}{Any noncharacter type
+        that has a storage size equal to \CONST{32} bits.}
+    \apitablerow{shmem\_put64}{Any noncharacter type that
+        has a storage size equal to \CONST{64} bits.}
+    \apitablerow{shmem\_put128}{Any noncharacter type that has a
+        storage size equal to \CONST{128} bits.}
+
+\apireturnvalues{
+    None.
+}
+\apinotes{
+}
+
+\begin{apiexamples}
+
+\apicexample
+    { The following \FUNC{shmem\_put\_signal} example is for \Cstd[11] programs:}
+    {./example_code/shmem_put_signal_example.c}
+    {}
+\end{apiexamples}
+
+\end{apidefinition}
+\color{Black}
--- a/content/shmem_put_signal.tex
+++ b/content/shmem_put_signal.tex
@@ -78,8 +78,10 @@ void @\FuncDecl{shmem\_ctx\_putmem\_signal}@(shmem_ctx_t ctx, void *dest, const 
 \begin{apiexamples}
 
 \apicexample
-    { The following example shows a simple ring-based broadcast operation using
-    \FUNC{shmem\_put\_signal}:}
+    {The following example demonstrates the usage of \FUNC{shmem\_put\_signal}.
+    It shows the implementation of a broadcast operation from \ac{PE} 0 to
+    itself and all other \acp{PE} in the job as a simple ring-based algorithm
+    using \FUNC{shmem\_put\_signal}:}
     {./example_code/shmem_put_signal_example.c}
     {}
 \end{apiexamples}
--- a/content/shmem_put_signal.tex
+++ b/content/shmem_put_signal.tex
@@ -76,11 +76,11 @@ void @\FuncDecl{shmem\_ctx\_putmem\_signal}@(shmem_ctx_t ctx, void *dest, const 
     word in a sequence consisting of a put routine followed by a put-with-signal
     routine does not imply delivery of the put routine's data.
 
-    The put-with-signal routines are compatible with all point-to-point
-    synchronization interfaces. The delivery of \VAR{signal} flag on the remote
-    \ac{PE} must not cause partial updates. This requires the update on
-    \VAR{signal} flag to be an atomic operation, with atomicity guarantees
-    described in Section~\ref{subsec:amo_guarantees}.
+    The signal set by the put-with-signal routines is compatible
+    with all point-to-point synchronization interfaces. The delivery of
+    \VAR{signal} flag on the remote \ac{PE} must not cause partial updates. This
+    requires the update on \VAR{signal} flag to be an atomic operation, with
+    atomicity guarantees described in Section~\ref{subsec:amo_guarantees}.
 }
 
 \begin{apiexamples}
--- a/content/shmem_put_signal.tex
+++ b/content/shmem_put_signal.tex
@@ -36,7 +36,7 @@ void @\FuncDecl{shmem\_ctx\_putmem\_signal}@(shmem_ctx_t ctx, void *dest, const 
     \apiargument{OUT}{dest}{Data object to be updated on the remote \ac{PE}. This
     data object must be remotely accessible.}
     \apiargument{IN}{source}{Data object containing the data to be copied.}
-    \apiargument{IN}{nelems}{Number of elements in the \VAR{dest} and \VAR{source}
+    \apiargument{IN}{nelems}{Number of elements in the \dest{} and \source{}
     arrays. \VAR{nelems} must be of type \VAR{size\_t} for \Cstd.}
     \apiargument{OUT}{sig\_addr}{Data object to be updated on the remote
     \ac{PE} as the signal. This signal data object must be
@@ -61,12 +61,12 @@ void @\FuncDecl{shmem\_ctx\_putmem\_signal}@(shmem_ctx_t ctx, void *dest, const 
 }
 
 \apinotes{
-    The \VAR{dest} and \VAR{sig\_addr} data objects must both be remotely
-    accessible. The \VAR{sig\_addr} and \VAR{dest} could be of different kinds,
+    The \dest{} and \VAR{sig\_addr} data objects must both be remotely
+    accessible. The \VAR{sig\_addr} and \dest{} could be of different kinds,
     for example, one could be a global/static \Cstd variable and the other could
     be allocated on the symmetric heap.
 
-    The \VAR{sig\_addr} and \VAR{dest} may not be overlapping in memory.
+    The \VAR{sig\_addr} and \dest{} may not be overlapping in memory.
 
     The delivery of \VAR{signal} flag on the remote \ac{PE} indicates only the
     delivery of its corresponding \dest{} data words into the data object on
--- a/content/shmem_put_signal.tex
+++ b/content/shmem_put_signal.tex
@@ -77,6 +77,13 @@ void @\FuncDecl{shmem\_ctx\_putmem\_signal}@(shmem_ctx_t ctx, void *dest, const 
     routine does not imply delivery of the put routine's data.
 }
 
+\apiimpnotes{
+    Implementations must ensure that put-with-signal routines are compatible
+    with all point-to-point synchronization interfaces. The delivery of
+    \signal{} flag on the remote \ac{PE} must not cause partial updates. This
+    requires the update on \signal{} flag to be an atomic memory operation.
+}
+
 \begin{apiexamples}
 
 \apicexample
--- a/content/shmem_put_signal.tex
+++ b/content/shmem_put_signal.tex
@@ -1,4 +1,3 @@
-\color{Green}
 \apisummary{
     The put-with-signal routines provide a method for copying data from a
     contiguous local data object to a data object on a specified \ac{PE}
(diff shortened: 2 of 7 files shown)


naveen-rn:
Some PENDING ITEMS that needs further discussions:
- add nbi support - can we add as a single proposal?
- for zero size puts - send only the signal is the expected behavior - does Annex C already capture semantics?
- do we need explanation about consecutive puts_signals or put with another put_signal?

content/shmem_put_signal.tex
```diff
@@ -0,0 +1,83 @@
+\color{Green}
+\apisummary{
+    The put with signal routines provide a method for copying data from a
+    contiguous local data object to a data object on a specified \ac{PE}
+    and set a remote flag to signal completion.
```
nspark: Suggest:
* "put with signal" &#8594; "put-with-signal"
* "...and set a remote flag..." &#8594; "...and subsequently setting a remote flag..."

content/shmem_put_signal.tex
```diff
@@ -0,0 +1,83 @@
...
+\apinotes{
+    The \VAR{sig\_addr} data object can be placed in the symmetric data segment
+    or the symmetric heap which can be same or different from the \VAR{dest}
+    data object.
+
+    The delivery of \signal{} flag on the remote \ac{PE} guarantees only the
+    delivery of its corresponding \dest{} data words into the data object on
+    the remote \ac{PE}. For example, two successive put with signal routines
+    or a successive put followed by a put with signal routine may deliver data
+    out of order.
```
nspark: Suggest changing:

> For example, two successive put with signal routines or a successive put followed by a put with signal routine may deliver data out of order.

to something like:

> Without a memory-ordering operation, there is no implied ordering between the delivery of the signal word of a put-with-signal routine and another data transfer. For example, the delivery of the signal word in a sequence consisting of a put routine followed by a put-with-signal routine does not imply delivery of the put routine's data.

content/shmem_put_signal.tex
```diff
@@ -0,0 +1,83 @@
...
+    remotely accessible.}
+    \apiargument{IN}{signal}{Unsigned 64-bit value used to set the remote
+    \VAR{sig\_addr} signal data object.}
+    \apiargument{IN}{pe}{\ac{PE} number of the remote \ac{PE}.}
+\end{apiarguments}
+
+\apidescription{
+    The routines return after the data has been copied out of the \source{}
+    array on the local \ac{PE}. The delivery of \signal{} flag on the remote
+    \ac{PE} guarantees the delivery of its corresponding \dest{} data words
```
nspark: For some reason, the word "guarantees" doesn't quite feel right. I wonder whether something like "indicates" might be better.

content/shmem_put_signal.tex
```diff
@@ -0,0 +1,83 @@
...
+}
+
+\apireturnvalues{
+    None.
+}
+
+\apinotes{
+    The \VAR{sig\_addr} data object can be placed in the symmetric data segment
+    or the symmetric heap which can be same or different from the \VAR{dest}
+    data object.
```
nspark: As of 1.4, I think we only have one reference to "data segment" (in the threading text, no less), so I'm hesitant to add another without more clearly defining it. In &sect;3.0 (Memory Model), we say:

> In OpenSHMEM the following kinds of data objects are symmetric:
>
> * Global and static C and C++ variables. [...]
>
> [...] Symmetric data objects are stored on the symmetric heap or in the global/static memory section of each PE.

I also think this text could improved with something like:

> The `dest` and `sig_addr` data objects must both be remotely accessible, but may each be allocated from the symmetric heap or global/static memory.

content/shmem_put_signal.tex
```diff
@@ -0,0 +1,85 @@
...
+}
+
+\apireturnvalues{
+    None.
+}
+
+\apinotes{
+    The \VAR{dest} and \VAR{sig\_addr} data object must both be remotely
+    accessible, but may each be allocated from the symmetric heap or global/
+    static memory.
```
manjugv: Should you just call it “symmetric data object” ?

example_code/shmem_put_signal_example.c
```diff
@@ -0,0 +1,49 @@
...
+
+    signals[0] = 0;
+    for (i = 0; i < size; i++) {
+        message[i] = VAL_USED;
+        data[i]    = 0;
+    }
+    shmem_barrier_all();
+
+    if (me != 0) {
+        shmem_long_wait_until((long *)&signals[0], SHMEM_CMP_EQ, 1);
```
nspark: Feedback from July 23 reading: Use `shmem_wait_until` or `shmem_uint64_wait_until`.

example_code/shmem_put_signal_example.c
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    shmemx_putmem_signal(data, message, size*sizeof(uint64_t),
```
nspark: Feedback from July 23 reading: `shmem_putmem_signal` not `shmemx_*`

example_code/shmem_put_signal_example.c
```diff
@@ -0,0 +1,49 @@
...
+    shmem_init();
+
+    size_t    size    = MAX_SIZE;
+    int       me      = shmem_my_pe();
+    int       n       = shmem_n_pes();
+    int       pe      = (me + 1)%n;
+
+    uint64_t* message = malloc(size * sizeof(uint64_t));
+    uint64_t* data    = shmem_malloc(size * sizeof(uint64_t));
+    uint64_t* signals = shmem_malloc(sizeof(uint64_t));
```
nspark: Feedback from July 23 reading: Might it better emphasize the symmetric heap or static variable distinction to use a static variable here?

example_code/shmem_put_signal_example.c
```diff
@@ -0,0 +1,49 @@
...
+            &signals[0], 1, pe);
+
+    if (me == 0) {
+        shmem_long_wait_until((long *)&signals[0], SHMEM_CMP_EQ, 1);
+        printf("BCAST with put with signal is complete\n");
+    }
+
+    free(message);
+    shmem_free(data);
+    shmem_free(signals);
```
jamesaross: shmem_free(signals) can be removed if using `static uint64_t signal = 0;` above.

example_code/shmem_put_signal_example.c
```diff
@@ -0,0 +1,49 @@
+#include <stdio.h>
```
jamesaross: Formatting of the example code should be more consistent with other example code.

content/shmem_put_signal.tex
naveen-rn: This was added specifically to note that both dest and sig_addr need not be part of the same SDO type - dest can be from SHEAP and sig_addr from global/static memory. Generalizing it as SDO makes it bit confusing. Let me know, if we reframe it better.

content/shmem_put_signal.tex
```diff
@@ -0,0 +1,85 @@
...
+    \apiargument{OUT}{dest}{Data object to be updated on the remote \ac{PE}. This
+    data object must be remotely accessible.}
+    \apiargument{IN}{source}{Data object containing the data to be copied.}
+    \apiargument{IN}{nelems}{Number of elements in the \VAR{dest} and \VAR{source}
+    arrays. \VAR{nelems} must be of type \VAR{size\_t} for \Cstd.}
+    \apiargument{OUT}{sig\_addr}{Data object to be updated on the remote
+    \ac{PE} as the signal. This signal data object must be
+    remotely accessible.}
+    \apiargument{IN}{signal}{Unsigned 64-bit value used to set the remote
+    \VAR{sig\_addr} signal data object.}
```
agrippa: I found this wording a bit confusing. One of many suggested rewordings:

Unsigned 64-bit value that is assigned to the remote \VAR{sig\_addr} signal data object.

content/shmem_put_signal.tex
```diff
@@ -0,0 +1,85 @@
...
+    accessible, but may each be allocated from the symmetric heap or global/
+    static memory.
+
+    The delivery of \signal{} flag on the remote \ac{PE} indicates only the
+    delivery of its corresponding \dest{} data words into the data object on
+    the remote \ac{PE}. Without a memory-ordering operation, there is no implied
+    ordering between the delivery of the signal word of a put-with-signal
+    routine and another data transfer. For example, the delivery of the signal
+    word in a sequence consisting of a put routine followed by a put-with-signal
+    routine does not imply delivery of the put routine's data.
```
agrippa: This is probably a question that has been asked and answered in other contexts but: are there any atomicity guarantees on the delivery of the signal? Should there be a note here calling that out? If there are no guarantees, should/could there be a separate set of APIs that deliver the signal atomically (possibly as an atomic operation like AND and not just an atomic set)?

nspark: I think the answers to your questions lie with the resolution of #204, #229, and the overall memory model work. I think one goal of those efforts should be that `shmem_wait_until` (and family) can be safely triggered by atomic operations and the signal of a put-with-signal. If that requires the signal to be an atomic remote write, then so be it.

agrippa: To your knowledge, has there been investigation into something like put-with-atomic-op? Semantically a putmem followed by an atomic operation (e.g. AND).

Is there any benefit in hardware to doing something like that, or would implementations just support it as putmem-fence-atomic?

nspark: The closest I can think of is the "counting puts" proposal from @jdinan & Co. This is equivalent to a put+fence+atomic-add. That is certainly a useful idiom. I'm not sure whether the full generality of put+fence+atomic-op is useful for all operations (i.e., set, add, and, or, xor).

I definitely think there would be application benefit in having hardware support for these idioms. It is often the case that fence (ordering, not necessarily completion) requires completion, due to network API constraints. If that fence can be avoided because the interconnect manages the ordering and atomic op, then the user is introducing fewer latency-sensitive operations.

shamisp:
@nspark ordered network will help but if you take in account something like
adaptive routing it may break it. You really want to have the packet small
enough such that you can squeeze it in single MTU.

example_code/shmem_put_signal_example.c
naveen-rn: Can this be a doc edit and wait till the uniform format is decided from #237

content/shmem_put_signal.tex
naveen-rn: @manjugv ?

manjugv: Sorry missed it. In that case, my preference is to use two sentences.

nspark:
August F2F discussion:
* @minsii: Should we add clarifications suggesting `shmem_wait_until` or `shmem_test` to clarify portable practices for accessing the signal word?
* @anshumang: Do we explicitly prohibit the signal word from overlapping with the source or destination buffers?

manjugv:
August F2F : Ballot postponed.

naveen-rn:
@minsii and @shamisp, I suppose a simple reference to put-with-signal (as in 76a5015) in the list of shmem routines which enforces ordering through p2p_sync operations would clarify the portable practice for using put-with-signal. Let me know, If we need some more clarifications.

naveen-rn:
Based on discussions with @anshumang  and others - 
@nspark It doesn't look like we could get some form of compiler optimizations from the `restrict` qualifier and also the primary purpose of `restrict` is to guide users. So, I'm not sure whether we need the `restrict` qualifier in the API - can just the details in the description section be sufficient?

naveen-rn:
To update, based on RMA WG discussions, we are planning to drop `restrict` qualifier for put-with-signal and can add it when we go through the whole specification. This is because of the following reasons:
1. Adding `restrict` to signal buffer makes sense - but not to both destination and source buffer - that makes it too restrictive
2. Avoid unnecessary compilation warnings

minsii:
@naveen-rn Thanks for the clarification fix with point-to-point sync. I have one more question: why do you define `sig_addr` with `uint64_t` type ?

naveen-rn:
@minsii In general, I would prefer a fixed width data type for representing the size of the signal, rather than using generic types which varies with platform. If I'm correct the C standards limits the width to 64. Also, the current implementation in Cray hardware supports only 64 bits for signal - hence we matched that with ```uint64_t```. It can be ```int64_t`` - if we prefer the signal to be negative. Are you interested in something different? Say an opaque object like - ```shmem_sig_t``` and restrict the size of this opaque object to 64 bits?

shamisp:
Sometimes it is useful to have negative for detection overflow bugs.

Best,
Pasha

minsii:
I am not sure whether fixed 64bit width is required in spec. Maybe `shmem_sig_t` opaque object is a more portable approach and the implementation can define different types internally ?

jdinan:
@nspark If your review has been addressed, could you please dismiss or update your review (ideally before this afternoon's vote on this ticket)?  Thanks!

manjugv:
> I am not sure whether fixed 64bit width is required in spec. Maybe `shmem_sig_t` opaque object is a more portable approach and the implementation can define different types internally ?

FWIW, the InfiniBand spec supports only 32bit as immediate data. I'm not sure, if Cray uses similar technique for sending the signal information with the packet on Cray h/w.

shamisp:
@manjugv - IBTA I don't think immediate is any useful in this context, regardless of size. It is delivered as part of CQE...

@minsii @naveen-rn if you decide to go with opaque object, you would have to introduce additional API to pull out completion out of the object.

minsii:
Can we extend the `shmem_wait*` APIs to support the new opaque object ?

content/shmem_put_signal.tex
```diff
@@ -0,0 +1,87 @@
...
+void @\FuncDecl{shmem\_\FuncParam{TYPENAME}\_put\_signal}@(TYPE *dest, const TYPE *source, size_t nelems, uint64_t *restrict sig_addr, uint64_t signal, int pe);
+void @\FuncDecl{shmem\_ctx\_\FuncParam{TYPENAME}\_put\_signal}@(shmem_ctx_t ctx, TYPE *dest, const TYPE *source, size_t nelems, uint64_t *restrict sig_addr, uint64_t signal, int pe);
+\end{Csynopsis}
+where \TYPE{} is one of the standard \ac{RMA} types and has a corresponding \TYPENAME{} specified by Table \ref{stdrmatypes}.
+
+\begin{CsynopsisCol}
+void @\FuncDecl{shmem\_put\FuncParam{SIZE}\_signal}@(void *dest, const void *source, size_t nelems, uint64_t *restrict sig_addr, uint64_t signal, int pe);
+void @\FuncDecl{shmem\_ctx\_put\FuncParam{SIZE}\_signal}@(shmem_ctx_t ctx, void *dest, const void *source, size_t nelems, uint64_t *restrict sig_addr, uint64_t signal, int pe);
+\end{CsynopsisCol}
+where \SIZE{} is one of \CONST{8, 16, 32, 64, 128}.
```
jdinan: Given that OpenSHMEM supports the fixed-width integer types, do we want to introduce new SIZE bindings?

content/shmem_put_signal.tex
```diff
@@ -0,0 +1,87 @@
...
+
+\begin{CsynopsisCol}
+void @\FuncDecl{shmem\_putmem\_signal}@(void *dest, const void *source, size_t nelems, uint64_t *restrict sig_addr, uint64_t signal, int pe);
+void @\FuncDecl{shmem\_ctx\_putmem\_signal}@(shmem_ctx_t ctx, void *dest, const void *source, size_t nelems, uint64_t *restrict sig_addr, uint64_t signal, int pe);
+\end{CsynopsisCol}
+
+\begin{apiarguments}
+    \apiargument{IN}{ctx}{The context on which to perform the operation.
+      When this argument is not provided, the operation is performed on
+      \CONST{SHMEM\_CTX\_DEFAULT}.}
```
jdinan: Needs to be updated with new text being used for the `ctx` argument.

content/shmem_put_signal.tex
```diff
@@ -0,0 +1,87 @@
...
+}
+
+\apinotes{
+    The \VAR{dest} and \VAR{sig\_addr} data objects must both be remotely
+    accessible. The \VAR{sig\_addr} and \VAR{dest} could be of different kinds,
+    for example, one could be a global/static \Cstd variable and the other could
+    be allocated on the symmetric heap.
+
+    The restrict qualifier in \VAR{sig\_addr} expects the data object to be
+    distinct from \VAR{dest} and \VAR{source} data objects.
```
jdinan: I don't understand why `sig_add` and `source` can't be the same pointer value corresponding to separate objects that are on different PEs.

content/shmem_put_signal.tex
```diff
@@ -0,0 +1,87 @@
...
+    remote \VAR{sig\_addr} signal data object.}
+    \apiargument{IN}{pe}{\ac{PE} number of the remote \ac{PE}.}
+\end{apiarguments}
+
+\apidescription{
+    The routines return after the data has been copied out of the \source{}
+    array on the local \ac{PE}. The delivery of \signal{} flag on the remote
+    \ac{PE} indicates the delivery of its corresponding \dest{} data words
+    into the data object on the remote \ac{PE}.
+}
```
jdinan: The API description appears to be missing a description of what this routine actually does.  This appears to also be an issue with `shmem_put`, but we shouldn't propagate it here.  For example, `shmem_get` states: "The get routines provide a method for copying a contiguous symmetric data object from a different PE to a contiguous data object on the local PE."

example_code/shmem_put_signal_example.c
```diff
@@ -0,0 +1,42 @@
...
+#include <shmem.h>
+#include <stdlib.h>
+
+int main(void)
+{
+    int i, err_count  = 0;
+
+    shmem_init();
+
+    size_t    size    = (2<<10);
```
jdinan: `2<<10 == 1<<11 == 2048`.  Do you mean 1<<10, for 2^10?

jdinan: I would argue that writing `2 << 10` to get 2048 is poor programming practice and would rather see `1 << 11` or `2048`.

example_code/shmem_put_signal_example.c
```diff
@@ -0,0 +1,42 @@
...
+    int       me      = shmem_my_pe();
+    int       n       = shmem_n_pes();
+    int       pe      = (me + 1)%n;
+    uint64_t* message = malloc(size * sizeof(uint64_t));
+    uint64_t* data    = shmem_malloc(size * sizeof(uint64_t));
```
jdinan: Why not use `shmem_calloc`?  You can remove `data[i] = 0` as well as `shmem_barrier_all()` below.

jdinan: I would move the declaration of `sig_addr` above the `shmem_calloc` if you make this change.  I can't convince myself that isn't a race without consulting the C language spec.

example_code/shmem_put_signal_example.c
```diff
@@ -0,0 +1,42 @@
...
+        message[i] = me;
+        data[i]    = 0;
+    }
+    shmem_barrier_all();
+
+    if (me != 0) {
+        shmem_uint64_wait_until(&sig_addr, SHMEM_CMP_EQ, 1);
+    }
+
+    shmem_putmem_signal(data, message, size*sizeof(uint64_t),
```
jdinan: Please use `shmem_put_signal` generic and just `size` as the count argument.

jdinan: The wait_until calls would also need to be updated to use generics.  If you prefer not to use generics, please be consistent and use `shmem_uint64_put_signal` here.

example_code/shmem_put_signal_example.c
```diff
@@ -0,0 +1,42 @@
...
+            &sig_addr, 1, pe);
+
+    if (me == 0) {
+        shmem_uint64_wait_until(&sig_addr, SHMEM_CMP_EQ, 1);
+        printf("BCAST with put with signal is complete\n");
+    }
```
jdinan: The last PE broadcasts to PE 0?  Why?  Also, remove the printf?

shamisp:
> Can we extend the `shmem_wait*` APIs to support the new opaque object ?

@minsii IMHO a bit hacky. If @naveen-rn decides to take this path, I would suggest to add an API to poll the notification (completions), create the notification object (potentially address or completion queue/CQ ?) and exchange the object identifier (address, CQ #, etc.) between initiator and target. In this way you can abstract an address or some hardware object. My understanding that this was not the goal of the API. The intend was to keep it super simple, aka make 2 puts to two different vaddr ? My $0.02

manjugv:
> @manjugv - IBTA I don't think immediate is any useful in this context, regardless of size. It is delivered as part of CQE...

Agree. Talking to Naveen, I was under the impression it was his intention. 

Irrespective, I don't see how both data and signal can be sent in only one packet, at least for 
the IB hardware because it has two virtual addresses. That said, I still support this proposal because we could employ other optimizations.

content/shmem_put_signal.tex
naveen-rn: I would prefer the put-with-signal to be consistent with other APIs in the specification. Removing the support for SIZE based will drop uniformity. But, if everyone decides it is necessary - I dont have any strong requirements for SIZE based put-with-signal API.

content/shmem_put_signal.tex
```diff
@@ -0,0 +1,87 @@
+\apisummary{
+    The put-with-signal routines provide a method for copying data from a
+    contiguous local data object to a data object on a specified \ac{PE}
+    and subsequently setting a remote flag to signal completion.
+}
+
+\begin{apidefinition}
+
+\begin{C11synopsis}
+void @\FuncDecl{shmem\_put\_signal}@(TYPE *dest, const TYPE *source, size_t nelems, uint64_t *restrict sig_addr, uint64_t signal, int pe);
```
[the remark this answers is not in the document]
naveen-rn: I plan to drop ```restrict``` - this was decided only after the last RMA WG meeting - I wasn't sure we can make changes into the PR when it is open for reading/voting. Will change it after today's meeting.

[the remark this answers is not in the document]
naveen-rn: BTW, ```restrict``` doesn't make sense in this usage.

naveen-rn:
@shamisp @minsii @manjugv - I don't plan to add a new opaque object - it brings lots of other confusions - how can we use it for other RMA/AMO operations, perform memory ordering. I would prefer to stay with any fixed width types - either ```uint64_t``` or ```int64_t``` - If we have strong requirements for negative values in the signal - we can change to ```int64_t```.

shamisp:
> @shamisp @minsii @manjugv - I don't plan to add a new opaque object - it brings lots of other confusions - how can we use it for other RMA/AMO operations, perform memory ordering. I would prefer to stay with any fixed width types - either `uint64_t` or `int64_t` - If we have strong requirements for negative values in the signal - we can change to `int64_t`.

👍 IMHO there is a room for a simple API as you proposed and potentially more complicated one that can abstract hardware counters, queues, etc (I think @jdinan  called it counting put or something along the same concept.)

content/shmem_put_signal.tex
jdinan: @nspark indicated these API will get used.  We can consider this comment to be resolved.

content/shmem_put_signal.tex
jdinan: From 9/24 meeting: Suggestion was to define something like "nonoverlapping buffers" early in the spec so that the definition can also be used in other RMA, AMO, etc. operations.

naveen-rn:
Voting postponed in Sep specification meeting

naveen-rn:
Review comments from Sep specification meeting:

Resolved:
1. Need to update the description with the api summary - not worry about how other RMA operations are described - we can fix them in other PRs
2. No need to remove SIZE based APIs - as we have use cases with structures
3. remove ```restrict``` and fix the description

Pending:
1. Whether we need ```int64_t``` or ```uint64_t```? @nspark was fine with ```uint64_t```. @shamisp suggested the possible use case of negative values (I assume as the signal is not an increment counter - this use case might not be common - but still probable)  - do we need to use union with both ```int64_t``` and ```uint64_t```?

naveen-rn:
Based on [RMA WG discussion](https://github.com/openshmem-org/specification/wiki/RMA-WG-09-27-2018) it was clarified that there is no need for change in datatype (no ```int64_t``` or union) required for the put-with-signal proposal. ```uint64_t``` as defined in the current proposal is acceptable

example_code/shmem_put_signal_example.c
naveen-rn: @jamesaross any comments on this review comment? can i mark this resolved?

content/shmem_put_signal.tex
naveen-rn: Fixed the ctx arg explanation to reflect the recent changes.

example_code/shmem_put_signal_example.c
naveen-rn: fixed used 2048 directly

naveen-rn:
@nspark @jdinan @shamisp @manjugv If I understand correctly, it was accepted in the previous RMA WG meeting that we will not add restrict qualifier in the API and also we will not have any explanation on the API notes specifically explaining the overlapping semantics for put-with-signal. This is because of the ambiguity in explaining all the scenarios:
1. source/target buffer can overlap 
2. source/signal buffer can overlap 
3. target/signal buffer can overlap

I suppose only 3 is not allowed in put-with-signal and others are undefined. Hence, we decided to drop explaining this particular scenario and decided to deal with this in a separate PR where we explain the general overlapping semantics for all the RMA routines. Let me know, if my understanding is correct?

content/shmem_put_signal.tex
naveen-rn: Fixed - duplicated summary into the description

example_code/shmem_put_signal_example.c
naveen-rn: removed this bcast to PE 0 and restored the SHMEM bcast semantics

nspark:
@naveen-rn I agree that only Scenario 3 is prohibited. I recall the discussion about the overlap scenarios, but not our decision on when or where to address such prohibitions. I don't recall this affecting other routines, so -- unless we can identify other routines with related overlap restrictions -- I think it should (for now) live with this API listing.

Thinking again about Scenario 3, one _could_ argue that it should still be well defined for the `signal` and `target` buffers to overlap. Since the write of the `target` word(s) are defined to happen before the write of the `signal` word, there actually is (or should be) a well-defined ordering in how memory is updated.

naveen-rn:
>  I don't recall this affecting other routines, so -- unless we can identify other routines with related overlap restrictions -- I think it should (for now) live with this API listing.

As the ordering semantics is guaranteed, to me this overlapping restriction in put-with-signal seems to be little forced. We are trying to avoid users from a particular usage, though the implementations are capable of handling it. If there are any hardware issues that we anticipate with the particular target/signal overlapping - I suppose we could then do that. AFAIK, there shouldn't be any.

Also, it would be hard to make a distinct case where we avoid just the target/signal buffer while the other two scenarios are undefined. What happens if both the source and target PEs are the same? It looks we should either address all the overlapping scenarios or we could keep it undefined as we do for all other RMAs. If we plan to address all the overlapping scenarios, then we should make it uniform across all RMAs.

naveen-rn:
@nspark @jdinan @shamisp @manjugv 

I would like to add the following statement for handling the overlap semantics:

    The \VAR{sig\_addr} and \VAR{dest} may not be same or overlapping data
    objects.

Or do we need something more detailed:

    The \VAR{sig\_addr} data object is expected to be distinct from \VAR{dest} 
    and \VAR{source} data objects.

content/shmem_put_signal.tex
```diff
@@ -0,0 +1,91 @@
...
+}
+
+\apinotes{
+    The \VAR{dest} and \VAR{sig\_addr} data objects must both be remotely
+    accessible. The \VAR{sig\_addr} and \VAR{dest} could be of different kinds,
+    for example, one could be a global/static \Cstd variable and the other could
+    be allocated on the symmetric heap.
+
+    The \VAR{sig\_addr} and \VAR{dest} may not be overlapping in memory.
+
```
minsii: Does the term "overlapping" cover both (1) same memory range and (2) not the same but partially overlapping memory range ? I think both cases are illegal here.

The reason I think it might be a little confusing is that, the collectives section uses "overlapping" in "The source and dest arrays may be the *same* array, but they may not be *overlapping* arrays."

Can we say: 
`The \VAR{sig\_addr} and \VAR{dest} may not be the same data object or overlapping in memory.` 

However, I think @jdinan had some concerns about use of "data object".

naveen-rn:
Changes made after the October meeting:
https://github.com/openshmem-org/specification/compare/820a4c..1dcf7e

This can be used for unofficial discussions and update announcements during the November 5th specification meeting. 

@nspark @jdinan @jamesaross @agrippa @manjugv  All the open comments based on previous discussions are now outdated. I suppose, every review comments based on these conversations are resolved. Let me know, if we could mark all these conversations as resolved.

manjugv:
👍

content/shmem_put_signal.tex
naveen-rn: resolved - closing outdated changes.

content/shmem_put_signal.tex
naveen-rn: resolved - closing outdated changes.

content/shmem_put_signal.tex
naveen-rn: resolved - removed reference to "data segment" in the explanation.
closing outdated changes.

content/shmem_put_signal.tex
naveen-rn: resolved - closing outdated changes.
changed guarantees to indicated.

content/shmem_put_signal.tex
naveen-rn: resolved - closing outdated changes.
modified as suggested.

example_code/shmem_put_signal_example.c
naveen-rn: fixed. resolved - closing outdated changes.

example_code/shmem_put_signal_example.c
naveen-rn: resolved - closing outdated changes.
Example program went through multiple changes and the comments are resolved.

example_code/shmem_put_signal_example.c
naveen-rn: resolved - closing outdated changes.
made changes as suggested.

example_code/shmem_put_signal_example.c
naveen-rn: marking as resolved as this change can be added as doc edit or made as part of #237. We dont have any standard formating template.

example_code/shmem_put_signal_example.c
naveen-rn: resolved - we started using C11 generics throughout the example for all SHMEM APIs.
closing outdated changes.

content/shmem_put_signal.tex
naveen-rn: resolved - modified as per the suggestion. closing outdated changes.

content/p2p_sync_intro.tex
```diff
@@ -3,8 +3,9 @@
 object.
 The point-to-point synchronization routines can be used to portably ensure
 that memory access operations observe remote updates in the order enforced by
-the initiator \ac{PE} using the \FUNC{shmem\_fence} and \FUNC{shmem\_quiet}
-routines.
+the initiator \ac{PE} using the put-with-signal(refer
+section~\ref{subsec:shmem_put_signal}, \FUNC{shmem\_fence} and
```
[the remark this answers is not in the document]
naveen-rn: resolved - closing outdated changes.

content/shmem_put_signal.tex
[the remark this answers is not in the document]
naveen-rn: removed restrict - closing outdated changes.

content/shmem_put_signal.tex
naveen-rn: resolved - closing outdated changes.

content/shmem_put_signal.tex
naveen-rn: resolved - closing outdated changes. removed reference to restrict qualifier.

example_code/shmem_put_signal_example.c
naveen-rn: resolved - used C11 generic SHMEM APIs in the example. closing outdated changes.

example_code/shmem_put_signal_example.c
naveen-rn: resolved - made changes as per the suggestion. 
closing outdated changes.

naveen-rn:
Modified and marked many comments as resolved - as these comments were addressed and many files were previously marked by git as outdated.

naveen-rn:
Changes made after the October meeting:
https://github.com/openshmem-org/specification/compare/820a4c..1dcf7e

The following changes were made:

1. removed restrict qualifier
2. updated the description section to contain the same explanation from summary
3. updated context argument explanation
4. added better explanation for sig_addr and dest overlap
5. avoided using shmem_malloc and shifted to using shmem_calloc in example
6. changed the size parameter input in example
7. modified all type-based shmem routines to C11-generic routines in the example
8. removed print option from example
9. clarified that the signal update is an atomic operation

We could use these changes for Jan 2019 OpenSHMEM specification meeting.

naveen-rn:
Changes made after the October meeting:
https://github.com/openshmem-org/specification/compare/fd54dd3..441d681 

The following changes were made:

1. removed restrict qualifier
2. updated the description section to contain the same explanation from summary
3. updated context argument explanation
4. added better explanation for sig_addr and dest overlap
5. avoided using shmem_malloc and shifted to using shmem_calloc in example
6. changed the size parameter input in example
7. modified all type-based shmem routines to C11-generic routines in the example
8. removed print option from example
9. clarified that the signal update is an atomic operation

We could use these changes for Jan 2019 OpenSHMEM specification meeting.
**[EDIT]**: Updated after rebasing with master

manjugv:
January 2019 Meeting: Voting postponed

content/shmem_put_signal.tex
```diff
@@ -68,19 +75,21 @@
...
+    completion of the signal update in a sequence consisting of a put routine
+    followed by a put-with-signal routine does not imply delivery of the put
+    routine's data.
+
+    The signal update by the put-with-signal routines is compatible with all
+    point-to-point synchronization interfaces. The delivery of \VAR{signal} flag
+    based on the \VAR{sig\_op} signal operator on the remote \ac{PE} must not
+    cause partial updates. Only concurrent accesses on \VAR{sig\_addr} by
+    different signal update operations using the same signal update operator is
+    guaranteed to be exclusive.
```
shamisp: I would suggest to clarify what "exclusive" means in this context.

manjugv:
@naveen-rn @jdinan As I mentioned in the WG meeting, I think it would be useful for defining partial and complete updates. I created an issue to track this and referenced both tickets that gets impacted.

jdinan:
@naveen-rn Can this one also be closed?

naveen-rn:
Closing this PR as we have opened a new composed PR with both blocking and non-blocking put-with-signal: https://github.com/openshmem-org/specification/pull/275

Issue #221, opened 2018-06-06
Questions on shmem_ctx_t


nspark:
I'm starting to work with OpenSHMEM 1.4 and the new OpenSHMEM contexts features, but I have some questions on `shmem_ctx_t` that I wonder whether we need to answer/clarify in the specification.

Consider the following code, which is intended to be similar to the thread-registration extensions in Cray SHMEM:

```c
// Assume C11 support
_Thread_local shmem_ctx_t thread_ctx;

void mylib_thread_register(void) {
  shmem_ctx_create(SHMEM_CTX_PRIVATE, &thread_ctx);
}

void mylib_thread_unregister(void) {
  shmem_ctx_destroy(thread_ctx);
}

inline void mylib_putmem(void *dst, const void *src, size_t size, int pe) {
  shmem_putmem(thread_ctx, dst, src, size, pe);
}
```

### Default initializers

Currently, there is no compile-time constant (e.g., `SHMEM_CTX_NULL` or `SHMEM_CTX_INVALID`) to which `thread_ctx` can be initialized. Without such a feature, it seems difficult to make calls to `mylib_thread_register()` idempotent. For example, if I had such a feature, I would write:

```c
_Thread_local shmem_ctx_t thread_ctx = SHMEM_CTX_INVALID;
void mylib_thread_register(void) {
  if (thread_ctx == SHMEM_CTX_INVALID)
    shmem_ctx_create(SHMEM_CTX_PRIVATE, &thread_ctx);
}
```

Without such a feature, I cannot test whether `thread_ctx` has already been created by the calling thread.

### Safety of `shmem_ctx_destroy`

Such a feature also allows one to more safely indicate that a `shmem_ctx_t` object has been destroyed with an idiom like:

```c
void mylib_thread_unregister(void) {
  shmem_ctx_destroy(thread_ctx);
  thread_ctx = SHMEM_THREAD_INVALID;
}
```

In Cray SHMEM's extensions, a call to `shmemx_thread_unregister()` from an unregistered thread has no effect. I do not think that repeated calls to `shmem_ctx_destroy` with the same `shmem_ctx_t` object should be valid (much like successive`free()` calls with the same pointer are an error), however, a call of `shmem_ctx_destroy(SHMEM_CTX_INVALID)` could (and should, I think) be valid (much like `free(NULL)`).

### Miscellany

[...]

PR #222, opened 2018-06-08
Specify validity of contexts, SHMEM_CTX_INVALID

nspark:
## Summary of Changes

This PR addresses issue #221 with the following changes:

* Defines the validity of communication contexts and the handles that refer to them.
* Adds the `SHMEM_CTX_INVALID` compile-time constant
* Specifies the behavior of `shmem_ctx_*` routines on invalid contexts:
  * `shmem_ctx_{destroy, fence, quiet}` perform no operation
  * RMA and AMO `shmem_ctx_*` routines have undefined behavior
* Adds a change-log entry
* Revises use of `SHMEM_CTX_DEFAULT` vs. "the default context"

## Notes

* On (much) closer reading of "communication contexts" and "context handles", I wonder whether we should update the following argument-description text that permeates the `shmem_ctx_*` routines:

  > `ctx`: The context on which to perform the operation. When this argument is not provided, the operation is performed on `SHMEM_CTX_DEFAULT`.

  To something like:

  > `ctx`: **A context handle specifying** the context on which to perform the operation. When this argument is not provided, the operation is performed on **the default context** ~~`SHMEM_CTX_DEFAULT`~~.

  The distinction here is that multiple handles may refer to the same context. (This is separate from the shared vs. private distinction with contexts.) Further, `SHMEM_CTX_DEFAULT` is a _context handle_, not the context itself.

### DIFF
--- a/content/library_constants.tex
+++ b/content/library_constants.tex
@@ -41,9 +41,11 @@ See Section~\ref{subsec:thread_support} for more detail about its use.
 %%
 \LibConstDecl{\newtext{SHMEM\_CTX\_INVALID}} &
 \newtext{
-A value to which a context handle may be initialized, assigned, or
-equality-compared to indicate or test whether it is an invalid communication
-context.
+A value corresponding to an invalid communication context.
+This value can be used to initialize or update context handles to indicate
+that they do not reference a valid context.
+When managed in this way, applications can use an equality comparison
+to test whether a given context handle references a valid context.
 See Section~\ref{sec:ctx} for more detail about its use.
 }
 \tabularnewline \hline
--- a/content/library_constants.tex
+++ b/content/library_constants.tex
@@ -39,6 +39,15 @@ may be multithreaded and any thread may invoke the \openshmem interfaces.
 See Section~\ref{subsec:thread_support} for more detail about its use.
 \tabularnewline \hline
 %%
+\LibConstDecl{\newtext{SHMEM\_CTX\_INVALID}} &
+\newtext{
+A value to which a context handle may be initialized, assigned, or
+equality-compared to indicate or test whether it is an invalid communication
+context.
+See Section~\ref{sec:ctx} for more detail about its use.
+}
+\tabularnewline \hline
+%%
 \LibConstDecl{SHMEM\_CTX\_SERIALIZED} &
 The context creation option which specifies that the given context
 is shareable but will not be used by multiple threads concurrently.
--- a/content/shmem_atomic_add.tex
+++ b/content/shmem_atomic_add.tex
@@ -58,6 +58,10 @@ CALL @\FuncDecl{SHMEM\_INT8\_ADD}@(dest, value_i8, pe)
     The \FUNC{shmem\_atomic\_add} routine performs an atomic add operation. It adds
     \VAR{value} to \dest{} on \ac{PE} \VAR{pe} and atomically updates the \dest{}
     without returning the value.
+    \newtext{
+    If the context handle \VAR{ctx} specifies an invalid context,
+    the behavior is undefined.
+    }
  }
 
 \apidesctable{
--- a/content/shmem_atomic_add.tex
+++ b/content/shmem_atomic_add.tex
@@ -40,9 +40,9 @@ CALL @\FuncDecl{SHMEM\_INT8\_ADD}@(dest, value_i8, pe)
 \end{Fsynopsis}
 
 \begin{apiarguments}
-    \apiargument{IN}{ctx}{The context on which to perform the operation.
+    \apiargument{IN}{ctx}{\oldtext{The context on which to perform the operation.} \newtext{A context handle specifying the context on which to perform the operation.}
         When this argument is not provided, the operation is performed on
-        \CONST{SHMEM\_CTX\_DEFAULT}.}
+        \oldtext{\CONST{SHMEM\_CTX\_DEFAULT}} \newtext{the default context}.}
     \apiargument{OUT}{dest}{The remotely accessible integer data object to be
         updated  on the remote \ac{PE}.  When using \CorCpp, the type of
         \dest{} should match that implied in the SYNOPSIS section.}
--- a/content/shmem_atomic_add.tex
+++ b/content/shmem_atomic_add.tex
@@ -59,7 +59,7 @@ CALL @\FuncDecl{SHMEM\_INT8\_ADD}@(dest, value_i8, pe)
     \VAR{value} to \dest{} on \ac{PE} \VAR{pe} and atomically updates the \dest{}
     without returning the value.
     \newtext{
-    If the context handle \VAR{ctx} specifies an invalid context,
+    If the context handle \VAR{ctx} does not correspond to a valid context,
     the behavior is undefined.
     }
  }
--- a/example_code/shmem_ctx_invalid.c
+++ b/example_code/shmem_ctx_invalid.c
@@ -1,0 +1,67 @@
+#include <stddef.h>
+#include <shmem.h>
+#include <omp.h>
+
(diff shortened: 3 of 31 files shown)


main_spec.tex
```diff
@@ -128,13 +128,25 @@ \subsection{Communication Management Routines}
...
+Context handles are of type \CTYPE{shmem\_ctx\_t} and \newtext{may be used} \oldtext{are valid} for
+language-level assignment and equality comparison.
+\newtext{
+A context handle may be initialized to, assigned, or equality-compared with
+the value \CONST{SHMEM\_CTX\_INVALID} to indicate or test whether the handle
+represents an invalid context.
+A valid context is either the default context or one created by a successful
+call to \FUNC{shmem\_ctx\_create} that has not yet been destroyed.
+Any other context is invalid and the behavior of its use in \ac{RMA}, \ac{AMO},
+and memory ordering routines is undefined.
```
nspark: This is an error. The behavior of invalid contexts with the memory ordering routines (`shmem_ctx_{fence, quiet}`) is defined as a no-op.

nspark: Well, specifically, `shmem_ctx_{fence, quiet}` are a no-op if the context handle has the value `SHMEM_CTX_INVALID`. If the context handle has some other invalid value, the behavior would be undefined.

content/library_constants.tex
```diff
@@ -39,6 +39,15 @@
 See Section~\ref{subsec:thread_support} for more detail about its use.
 \tabularnewline \hline
 %%
+\LibConstDecl{\newtext{SHMEM\_CTX\_INVALID}} &
+\newtext{
+A value to which a context handle may be initialized, assigned, or
```
jdinan: The style here doesn't match the other table entries, which start with a sentence fragment.

I'm also concerned about the last part, "test whether it is *an* invalid communication context".  `SHMEM_CTX_INVALID` is one of many invalid handle values, but the statement suggests that it subsumes all invalid handle values.  I wonder whether something like `SHMEM_CTX_NULL` would be less prone to misinterpretation, since what this text describes is similar to how the null pointer value is used?

Here's a suggested update:
```
A value corresponding to an invalid communication context.  This value can be used to initialize or update context handles to indicate that they do not reference a valid context.  When managed in this way, applications can use an equality comparison to test whether a given context handle references a valid context.
```

content/shmem_atomic_add.tex
```diff
@@ -58,6 +58,10 @@
     The \FUNC{shmem\_atomic\_add} routine performs an atomic add operation. It adds
     \VAR{value} to \dest{} on \ac{PE} \VAR{pe} and atomically updates the \dest{}
     without returning the value.
+    \newtext{
+    If the context handle \VAR{ctx} specifies an invalid context,
+    the behavior is undefined.
```
jdinan: Not quite sure why, but I think I prefer "If the context handle \VAR{ctx} does not correspond to a valid context, the behavior is undefined."  Seems like there are more paths to this handle corresponding to something invalid, than there are to something valid.

main_spec.tex
```diff
@@ -118,8 +118,8 @@ \subsubsection{\textbf{SHPDEALLC}}\label{subsec:shpdeallc}
 
 \subsection{Communication Management Routines}
 \label{sec:ctx}
-All \openshmem RMA, AMO, and memory ordering routines are
-performed on a communication context.  The communication context defines an
+All \openshmem \ac{RMA}, \ac{AMO}, and memory ordering routines are
+performed on a \newtext{valid} communication context.  The communication context defines an
```
jdinan: Reads a bit funny with "valid", which is adding a restriction.  What about changing "are performed" to "must be performed"?

main_spec.tex
```diff
@@ -128,13 +128,25 @@ \subsection{Communication Management Routines}
 additionally provide thread isolation, eliminating overheads resulting from
 thread interference.
 
-Context handles are of type \CTYPE{shmem\_ctx\_t} and are valid for
-language-level assignment and equality comparison.  A handle to the desired context is
-passed as an argument in the \Cstd \CTYPE{shmem\_ctx\_*} and type-generic API
-routines.  API routines that do not accept a context argument operate on the
+\oldtext{A handle to the desired context}
+\newtext{A communication context is specified by a context handle, which} is
```
jdinan: s/specified/identified/?

main_spec.tex
```diff
@@ -128,13 +128,25 @@ \subsection{Communication Management Routines}
...
+routines.  \ac{API} routines that do not accept a context \newtext{handle} argument operate on the
 default context.  The default context can be used explicitly through the
 \LibHandleRef{SHMEM\_CTX\_DEFAULT} handle.
 
+Context handles are of type \CTYPE{shmem\_ctx\_t} and \newtext{may be used} \oldtext{are valid} for
+language-level assignment and equality comparison.
+\newtext{
+A context handle may be initialized to, assigned, or equality-compared with
+the value \CONST{SHMEM\_CTX\_INVALID} to indicate or test whether the handle
+represents an invalid context.
```
jdinan: Same concern with respect to "equalith-compared with ... to test whether the handle represents *an* invalid context."

main_spec.tex
```diff
@@ -128,13 +128,25 @@ \subsection{Communication Management Routines}
...
+A valid context is either the default context or one created by a successful
```
jdinan: Suggest splitting this into two sentences: "The default context is valid for the duration of the OpenSHMEM portion of an application" and "Contexts created by a successful call to ... remain valid until they are destroyed."  And then "A handle value that does not correspond to a valid context is considered to be invalid and its use in ... results in undefined behavior."

example_code/shmem_ctx_invalid.c
```diff
@@ -0,0 +1,67 @@
+#include <stddef.h>
+#include <shmem.h>
+#include <omp.h>
+
+_Thread_local shmem_ctx_t thread_ctx = SHMEM_CTX_INVALID;
+
+void lib_thread_register(void) {
+  if (thread_ctx == SHMEM_CTX_INVALID)
+    if (shmem_ctx_create(SHMEM_CTX_PRIVATE, &thread_ctx) &&
+        shmem_ctx_create(                0, &thread_ctx))
```
jdinan: I wonder if this might be too clever for an example.  ;)

example_code/shmem_ctx_invalid.c
```diff
@@ -0,0 +1,67 @@
...
+  const int n_pes = shmem_n_pes();
+  const int count = 1 << 15;
+
+  int *src_bufs[n_pes];
+  int *dst_bufs[n_pes];
+  for (int i = 0; i < n_pes; i++) {
+    src_bufs[i] = shmem_malloc(count * sizeof(*src_bufs[i]));
+    if (src_bufs[i] == NULL)
+      shmem_global_exit(3);
+    dst_bufs[i] = shmem_malloc(count * sizeof(*dst_bufs[i]));
```
jdinan: Calloc?

main_spec.tex
nspark: 🤔

nspark:
[PDF draft of this PR as of 9318079](https://github.com/openshmem-org/specification/files/2227972/openshmem-draft-9318079.pdf)

nspark:
[PDF draft of this PR as of b82300e](https://github.com/openshmem-org/specification/files/2228063/openshmem-draft-b82300e.pdf)

nspark:
[Changes for special ballot](https://github.com/openshmem-org/specification/pull/222/files/c6789f328de7da5d14697cf2978fcd65fedd9239..b82300e8eab41522c7c23efed74f473b2f1c5945) for August 22-23 committee meeting

PR #236, opened 2018-07-30
Add PDF auto-generation to PRs

nspark:
This PR adds auto-generation of the OpenSHMEM Specification drafts on PRs. It does so via Travis CI, which runs the existing PDF build test, then uploads the PDF to a (private) Dropbox account, creates a shared link, then adds a comment on the PR that links to the PDF.

### DIFF
--- a/.travis-deploy.py
+++ b/.travis-deploy.py
@@ -11,12 +11,15 @@ try:
     PR_number = os.environ['TRAVIS_PULL_REQUEST']
     if PR_number == 'false':
         # Skip PDF deployment on non-PR builds
-        raise KeyError()
+        raise RuntimeError('not a PR')
     repo_slug = os.environ['TRAVIS_REPO_SLUG']
     DB_ACCESS_TOKEN = os.environ['DROPBOX_ACCESS_TOKEN']
     GH_ACCESS_TOKEN = os.environ['GITHUB_ACCESS_TOKEN']
 except KeyError as e:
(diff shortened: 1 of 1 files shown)


nspark:
Hmm... the PDF deployment didn't work.

nspark:
Ah, [this security restriction of Travis CI](https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions) is the "problem".

nspark:
I'm not sure there's a good solution here. Given these restrictions, PDF generation can work for non-fork PRs, but that would require (some of) us to push directly to branches on `openshmem-org/specification` (vs. `<user>/specification`. If we do that, we'll have to maintain a list of users who can create and push to branches (we probably do that already), and we should probably make `master` a protected branch.

What do you think, @openshmem-org/specification-owners?

jdinan:
We can also create forks for the working groups under `openshmem-org` (e.g. `teams-wg`, `rma-wg`, etc.) and install the necessary keys into those forks.  That would allow us to maintain a clean central repository, but also limit the set of places where keys need to be installed.  Another advantage to WG forks is that each WG can have separate issues lists, projects, etc (I think the teams WG already does this).

Working group chairs (or their designees) would need to manage branches and permissions on the WG fork.  Someone will need to update each master branch when we merge changes upstream.  If we want, I'm happy to do this (since I get to push the big green button).  We can still accept PRs from individual forks in this scenario, we just won't get PDFs automatically.

Issue #216, opened 2018-05-21
Merge NBI operations into RMA routines section


jdinan:
Merge the NBI operations into the "RMA Routines" section.

Make sure to revisit introductory text to ensure that nonblocking implicit operations are accounted for in the RMA introduction.

shamisp:
Please take a look at #362.

shamisp:
Moved here https://github.com/shamisp/specification/pull/1

PR #238, opened 2018-08-08
Adding support for NBI put with signal operation

naveen-rn:
This PR extends the #218 proposal adding non-blocking support for put-with-signal operations. The nonblocking variant of the put-with-signal operation is semantically equivalent to:
```
shmem_put_nbi();
shmem_fence();
shmem_put_nbi();
```
Following the trend of no-example for any NBI operations, there is no example added for this proposed routine in this PR. And the placement of the routine, will also be modified based on Issue #216.

Expected DOC changes:
1. Since the change log entry framework and the keyword highlighting changes are added as part of #218 - there is no need to do it again in this PR. We will clean it up after reading.

### DIFF
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -164,9 +164,9 @@ or the input is not in accordance with the Specification, the behavior
 is undefined.
 
 \begin{longtable}{|>{\raggedright}p{0.3\textwidth}|>{\raggedright}p{0.6\textwidth}|}
-\hline 
+\hline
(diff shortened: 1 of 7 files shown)


naveen-rn:
I had to close this PR - because the history in this branch was modified and makes it hard to maintain a separate branch for blocking and nbi-put-with-signal.
I have created #244 for continuing the prpoposal to add NBI put-with-signal

PR #244, opened 2018-09-10
Add support for NBI put-with-signal operation

naveen-rn:
This PR extends the #218 proposal adding non-blocking support for put-with-signal operations. The nonblocking variant of the put-with-signal operation is semantically equivalent to:

shmem_put_nbi();
shmem_fence();
shmem_put_nbi();

Following the trend of no-example for any NBI operations, there is no example added for this proposed routine in this PR. And the placement of the routine, will also be modified based on Issue #216.

Expected DOC changes:
1. Since the change log entry framework and the keyword highlighting changes are added as part of #218 - there is no need to do it again in this PR. We will clean it up after reading.

### DIFF
--- a/content/shmem_put_signal_nbi.tex
+++ b/content/shmem_put_signal_nbi.tex
@@ -8,32 +8,32 @@
 \begin{apidefinition}
 
 \begin{C11synopsis}
-void @\FuncDecl{shmem\_put\_signal\_nbi}@(TYPE *dest, const TYPE *source, size_t nelems, uint64_t *restrict sig_addr, uint64_t signal, int pe);
-void @\FuncDecl{shmem\_put\_signal\_nbi}@(shmem_ctx_t ctx, TYPE *dest, const TYPE *source, size_t nelems, uint64_t *restrict sig_addr, uint64_t signal, int pe);
+void @\FuncDecl{shmem\_put\_signal\_nbi}@(TYPE *dest, const TYPE *source, size_t nelems, uint64_t *sig_addr, uint64_t signal, int pe);
+void @\FuncDecl{shmem\_put\_signal\_nbi}@(shmem_ctx_t ctx, TYPE *dest, const TYPE *source, size_t nelems, uint64_t *sig_addr, uint64_t signal, int pe);
 \end{C11synopsis}
 where \TYPE{} is one of the standard \ac{RMA} types specified by Table \ref{stdrmatypes}.
 
 \begin{Csynopsis}
-void @\FuncDecl{shmem\_\FuncParam{TYPENAME}\_put\_signal\_nbi}@(TYPE *dest, const TYPE *source, size_t nelems, uint64_t *restrict sig_addr, uint64_t signal, int pe);
-void @\FuncDecl{shmem\_ctx\_\FuncParam{TYPENAME}\_put\_signal\_nbi}@(shmem_ctx_t ctx, TYPE *dest, const TYPE *source, size_t nelems, uint64_t *restrict sig_addr, uint64_t signal, int pe);
+void @\FuncDecl{shmem\_\FuncParam{TYPENAME}\_put\_signal\_nbi}@(TYPE *dest, const TYPE *source, size_t nelems, uint64_t *sig_addr, uint64_t signal, int pe);
+void @\FuncDecl{shmem\_ctx\_\FuncParam{TYPENAME}\_put\_signal\_nbi}@(shmem_ctx_t ctx, TYPE *dest, const TYPE *source, size_t nelems, uint64_t *sig_addr, uint64_t signal, int pe);
 \end{Csynopsis}
 where \TYPE{} is one of the standard \ac{RMA} types and has a corresponding \TYPENAME{} specified by Table \ref{stdrmatypes}.
 
 \begin{CsynopsisCol}
-void @\FuncDecl{shmem\_put\FuncParam{SIZE}\_signal\_nbi}@(void *dest, const void *source, size_t nelems, uint64_t *restrict sig_addr, uint64_t signal, int pe);
-void @\FuncDecl{shmem\_ctx\_put\FuncParam{SIZE}\_signal\_nbi}@(shmem_ctx_t ctx, void *dest, const void *source, size_t nelems, uint64_t *restrict sig_addr, uint64_t signal, int pe);
+void @\FuncDecl{shmem\_put\FuncParam{SIZE}\_signal\_nbi}@(void *dest, const void *source, size_t nelems, uint64_t *sig_addr, uint64_t signal, int pe);
+void @\FuncDecl{shmem\_ctx\_put\FuncParam{SIZE}\_signal\_nbi}@(shmem_ctx_t ctx, void *dest, const void *source, size_t nelems, uint64_t *sig_addr, uint64_t signal, int pe);
 \end{CsynopsisCol}
 where \SIZE{} is one of \CONST{8, 16, 32, 64, 128}.
 
 \begin{CsynopsisCol}
-void @\FuncDecl{shmem\_putmem\_signal\_nbi}@(void *dest, const void *source, size_t nelems, uint64_t *restrict sig_addr, uint64_t signal, int pe);
-void @\FuncDecl{shmem\_ctx\_putmem\_signal\_nbi}@(shmem_ctx_t ctx, void *dest, const void *source, size_t nelems, uint64_t *restrict sig_addr, uint64_t signal, int pe);
+void @\FuncDecl{shmem\_putmem\_signal\_nbi}@(void *dest, const void *source, size_t nelems, uint64_t *sig_addr, uint64_t signal, int pe);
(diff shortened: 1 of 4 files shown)


naveen-rn:
Changes made after the October 2018 Meeting:
https://github.com/openshmem-org/specification/compare/8e386..1ce6d5

October 2018 is when we had the official reading of this ticket. These changes can be used for updates in the November 2018 meeting and used later for special ballot.

naveen-rn:
Changes made after the October 2018 Meeting:
https://github.com/openshmem-org/specification/compare/8e38627..4fcc2d6

The following changes were made similar to #218 (comment):
1. removed restrict qualifier
2. updated the description section to contain the same explanation from summary
3. updated context argument explanation
4. added better explanation for sig_addr and dest overlap
5. clarified NBI signal update as an atomic operation

October 2018 is when we had the official reading of this ticket. These changes can be used for updates in the January 2019 meeting and used for special ballot.

naveen-rn:
Changes made after the October 2018 Meeting:
https://github.com/openshmem-org/specification/compare/2b39c09..3b48d2b 

The following changes were made similar to #218 (comment):
1. removed restrict qualifier
2. updated the description section to contain the same explanation from summary
3. updated context argument explanation
4. added better explanation for sig_addr and dest overlap
5. clarified NBI signal update as an atomic operation

October 2018 is when we had the official reading of this ticket. These changes can be used for updates in the January 2019 meeting and used for special ballot.
[EDIT]: Updated after rebasing with master

content/shmem_put_signal_nbi.tex
```diff
@@ -0,0 +1,90 @@
...
+    The delivery of \VAR{signal} flag on the remote \ac{PE} indicates only the
+    delivery of its corresponding \dest{} data words into the data object on
+    the remote \ac{PE}. Without a memory-ordering operation, there is no implied
+    ordering between the delivery of the signal word of a nonblocking
+    put-with-signal routine and another data transfer. For example, the delivery
+    of the signal word in a sequence consisting of a put routine followed by a
+    nonblocking put-with-signal routine does not imply delivery of the put
+    routine's data.
+
+    The nonblocking put-with-signal routines are compatible with all
```
jdinan: Suggest: "The signal set by the nonblocking put-with-signal routines is compatible with..."

content/shmem_put_signal_nbi.tex
```diff
@@ -0,0 +1,90 @@
...
+    point-to-point synchronization interfaces. The delivery of \VAR{signal} flag
+    on the remote \ac{PE} must not cause partial updates. This requires the
+    update on \VAR{signal} flag to be an atomic operation, with atomicity
+    guarantees described in Section~\ref{subsec:amo_guarantees}.
```
jdinan: This text is too strong.  We don't want the signal update to be a SHMEM-level atomic.  If a network can implement this with a simple write, it should be allowed.  Suggest something like:

"The put-with-signal interfaces must be implemented such that a synchronization operation on the remote \ac{PE} does not observe partial updates to \VAR{signal}. On some platforms, this may require the update of the \VAR{signal} flag to be performed using an atomic operation."

naveen-rn: I'm fine with your suggestions. I will wait for @shamisp to comment on this. Initially I phrased it similar to your suggestion. @shamisp requested for a stronger semantics. 

I suppose @manjugv  is also looking into this on his side.

naveen-rn: @jdinan @manjugv @shamisp Though the comment is shown outdated - this one is not fixed. 
Need some feedback. I suppose with this one in pending - we wont be able to vote on the PRs.

To me the core of both @jdinan suggested changes and @shamisp review comments almost looks same to me. The issue is - we are trying to determine an implementation detail. 

If some implementation can implement ```shmem_atomic_set``` as simple ```shmem_p``` operation. It doesn't matter if this statement is too strong. Internally it is going to still go through the same path.

IMHO, we should completely remove the implementation detail. We should just say:

> The put-with-signal interfaces must be implemented such that a synchronization operation on the remote \ac{PE} does not observe partial updates to \VAR{signal}.

jdinan: I still prefer the minimally specified "is compatible with remote wait/test" text.  It's crucial for open standards to keep the implementation space open to support future and alternative architectures (e.g. Gen-Z, CCIX, NVLink, etc.) where the signal can be non-atomic.  SHMEM on shared memory (e.g. POSH) is another example, depending on the underlying architecture.

My understanding is that the weaker text is compatible with current architectures, while keeping the implementation space open.  I appreciate that the atomic text is clearer to present day implementors and, as a result, is more likely to be implemented correctly.  Apart from clarity, is there something else I'm missing?  If the preference for the atomic text is being driven by clarity, we can put more work into the wait-compatible version of the text to address this.

naveen-rn: Sorry, I couldn't reply on this earlier. During the last RMA WG meeting - there were multiple proposals from @anshumang and @shamisp that would enable us to remove this hard AMO constraints from put-with-signal.

As per their proposal,
1. plan to introduce a new routine called shmem_signal() which has the semantic requirements of atomic to itself guarantees, but not wrt to other OpenSHMEM atomics,
2. plan to modify the existing shmem_wait_until/test semantics - where it is expected to provide complete updation guarantees and no partial updates allowed. Though now it is only stated as notes to implementers - they want to make this a real requirement in the description.

If I understand correctly, we had overall consensus on (1) but (2) had disagreements in the group, as we need to go through the network and it makes shmem_wait_until/test a heavy API on some architectures. I don't see a huge change in AARCH64/x86_64 but GPU has issues.

With (1) we could remove the atomicity requirements from put-with-signal. But, to do that we have to wait on voting this proposal until the shmem_signal() is merged. 

That said, I'm certain that this is just from the RMA WG discussion - this will have to go through the larger committee before being finalized. We decided to discuss this with the larger group including the users and then decide on the outcome during the next OpenSHMEM specification meeting. These changes should also affect #214.

anshumang: Thanks @naveen-rn for the summary. 
One clarification that (2) does not make it any worse for NVIDIA GPUs than it already is if the implementation follows note to implementors and prohibits partial updates from being seen.

shamisp: @jdinan Correct, we would have two introduce two sets of wait/test APIs (at least this is the proposal that was floated around).

jdinan: @shamisp What I suggested above is essentially the dual-API, but without adding any new functions.  Or looking at it a different way, it fills in the definition for how the existing APIs work together in a way that balances backward compatibility with performance.  We would define wait/test with `SHMEM_CMP_NE` to be compatible with `shmem_p` and put-with-signal to allow them to update the signal word using ordinary puts (and also AMOs, of course, but not at the same time).  This requires stable updates, but not single-copy atomicity.  All other comparison operations would require single-copy atomicity to avoid returning a false positive from the wait/test routine; therefore, these comparisons would only have defined behavior with AMOs.  Interested to hear what you think of this idea.

anshumang: @jdinan I think having a non-AMO API that is compatible with `wait_until/test` across all predicates (`CMP_GE`, `CMP_LE` etc) will help performance on platforms which have a single-copy-atomic store faster than a read-modify-write-atomic cmpexchg. In terms of use, does defining non-single-copy-atomic `shmem_p` to be compatible with `wait_until/test` for only `CMP_NE` covers most uses and truly addresses backward compatibility?

bcernohous: Someone can correct me, but my understanding from the meeting was that the existing APIs will work with SIGNAL **or** AMOs.  Not mixed.  

We can expect shmem_waitxx to work fine when the remote source is using signals and not also, for example, incrementing the signal.   And shmem_waitxx will work fine on counters as long as signals are also written.  

There's just no guarantee that you can shmem_put_signal and shmem_long_add to the same location.

bcernohous: @jdinan , understood.  And there was some discussion of doubling the APIs to support both"no partial updates" and "partial update ok" versions of wait.  

IMO, before I've seen the upcoming "shmem signal" (not shmem_put_signal) proposal, I think SHMEM takes the hard path.  No partial updates.   

An app that wants performance and tolerates partial updates could use shmem_put/do while(flag!=0).  It's not clear to me that we need API's for that.  My 2 cents.

jdinan: @bcernohous The question @naveen-rn asked earlier in the thread was whether the signal update performed by put-with-signal should be specified as a SHMEM AMO or a "no partial updates" update.  What's your preference?

bcernohous: @jdinan , I'm waiting to see the new shmem_signal proposal.  I think we define them similarly and they both satisfy shmem_wait.  So it partly depends on what we decide for shmem_wait.

jdinan:
@naveen-rn Can this PR be closed now that it is merged with #275?

naveen-rn:
Closing this PR as we have opened a new composed PR with both blocking and non-blocking put-with-signal: #275

Issue #249, opened 2018-10-01
Adding Fortran bind(C) support for basic OpenSHMEM routines in a separate document


naveen-rn:
## Problem
As per #194, we are planning to separate the main Fortran replacement interface proposal into multiple smaller parts. This issue tracks the first part of the proposal - to add and support the initial list of Fortran bind(C) routines.

## List of supported OpenSHMEM interface
The following are the list of planned OpenSHMEM routines available in the Fortran bind(C) interface.

| | | | |
|-|-|-|-|
shmem_init () | shmem_ctx_create () | atomic_swap | shmem_wait_until() |
shmem_finalize () | shmem_ctx_destroy () | atomic_fetch_add | barrier |
shmem_my_pe () | shmem_putmem() | atomic_fetch_inc | sync |
shmem_n_pes () | shmem_putmem_nbi() | atomic_fetch_and | all-to-all |
shmem_global_exit () | shmem_getmem() | atomic_fetch_or | all-to-alls |
shmem_init_thread () | shmem_getmem_nbi() | atomic_fetch_xor | broadcast |
shmem_query_thread | atomic_fetch | shmem_quiet () | reduce |
shmem_malloc () | atomic_set | shmem_fence () | collect |
shmem_free () | atomic_cswap | shmem_test () | fcollect |

## Changes
All these routines will be added as part of a separate document and not be included in the main OpenSHMEM specification.

jeffhammond:
What's the value of standardizing this versus the interested parties providing an implementation to the community as OSS?

There is a real cost to standardization in the form of our time. I'd like to see the implementation in the wild and some evidence that we need more than one implementation of the same API. If there is only going to be one implementation, then there is no need to spend time on standardization.

naveen-rn:
If you look at the proposal - we have tried to split it as three different sub-levels:
1. RM all the existing FTN support - DONE
2. Introduce Fortran Bind(C) based base support - This #249 issue
3. Introduce the use of standard Coarray features with OpenSHMEM to make (2) more clean - Issue #252 

Though our goal is just to have (3) in the spec - introducing the changes as multiple smaller items makes it little easier for standardization. To answer why we need (3) as you have asked in https://github.com/openshmem-org/specification/issues/252#issuecomment-870184975:

[...]

PR #257, opened 2018-11-25
Adding Fortran bind(C) support for basic OpenSHMEM routines

naveen-rn:
As per Issue #249, we are creating a new separate document for adding separate language bindings which are not part of the main OpenSHMEM specification. We have created this separate document for adding Fortran bind(C) language bindings and this can be extended when we need other new language bindings in future.


language/ftn08/api.tex
```diff
@@ -0,0 +1,365 @@
+As the \Fortran[bind(C)] language bindings are implemented over the \openshmem
```
BryantLam: Please move all of these files from directory "ftn08" to directory "fortran2008" to more closely match the referred-to named "ISO/IEC 1539-1:2010 (Fortran 2008)". Or "fortran2018", or just "fortran" if you never anticipate ever having two separate sets of FortranXXXX language bindings.

language/lang_ext_overview.tex
```diff
@@ -0,0 +1,7 @@
+\openshmem provides ISO \Cstd bindings. \Cstd bindings are the primary language
```
BryantLam: The first and second sentences sound redundant. I would drop one of the sentences, or change/reword/reorder them if you were trying to impart specific information.

language/ftn08/api.tex
BryantLam: Nick had a comment that if these bindings are for Fortran 2018 because of its support for AMOs, collectives, etc., this document should make that comment somewhere so that users know why the C_BIND interface for AMOs, collectives are missing.

nspark:
Should the OpenSHMEM bindings for Fortran specify the relationship between...?

- `shmem_my_pe()` and `this_image()`
- `shmem_n_pes()` and `num_images()`

Issue #258, opened 2018-11-26
shmem_malloc Interface to Leverage Hierarchical & Heterogenous Memory Characteristics 


manjugv:
## Problem: 

A typical node in the current HPC systems is composed of variety of memories and organized into multiple hierarchies and/or have different affinities to the PEs and threads. The OpenSHMEM programming model and its memory allocation routines are oblivious to these variations. As a consequence, it is a challenge for the OpenSHMEM program to leverage memory characteristics and capabilities to achieve higher performance in a portable way.

## Proposal:

Introduce memory allocation interface that can pass hints to the OpenSHMEM implementations.
The memory hints are then utilized by the implementations to optimize memory for that behavior. 
For example, if the user specifies that a particular allocation is used as a pSync array, then the implementation can use the memory that is available on NUMA memory bank that is near to the network. In the cases where the memory is available on the network interface, it can allocate that memory for the pSync array. This can impact the latency characteristics.

## Impact on Users:

These interfaces should provide an opportunity to the user to provide usage information to 
the implementation, which implementations can then utilize to optimize for that behavior.  If the implementations 
optimize for that behavior, the programs should achieve higher performance and/or scalability. 

The OpenSHMEM programs not using these interfaces or using SHMEM_HINT_NONE is not 
impacted. 

## Impact on Implementations: 

This provides an opportunity for the implementations to optimize the behavior for particular usage. 
If an implementation does not support optimizations, it is allowed to default to shmem_malloc behavior.

## Useful References 
1. https://software.intel.com/sites/default/files/managed/5f/5e/MCDRAM_Tutorial.pdf
2. SharP Unified Memory Allocator: https://www.osti.gov/biblio/1468045
3. mbind: https://www.kernel.org/doc/html/v4.18/admin-guide/mm/numa_memory_policy.html

naveen-rn:
How is this different from #195? It looks like we are trying to address the same issue.

manjugv:
@naveen-rn 

Naveen, I knew you would ask this. :) 

The main difference is that there is only one symmetric heap here (no change to symmetric address model) and complexity of memory management is handled by the library. It is a small change (adding only one interface) and we can get most of the benefits. The approach in #195 is more explicit and here it is more implicit. That said, I feel there is value in having both solutions and they can co-exist.

jamesaross:
So this proposal is for something like this?
```c
void* shmem_malloc_hint(size_t size, int hint);
```
[...]

Issue #229, opened 2018-07-19
Memory model clarifications


anshumang:
Summary : ordering (inside a PE) + reads from (between two PEs) = happens before (across all PEs)

Following are the items that have been discussed in RMA WG on 6/21, 7/5 and 7/19 and are still open (except one marked by ^). They are grouped below 1) under ordering, 2) reads from and 3) happens before.

1) ordering
* Fetch AMOs are ordered (related: fence is not required to order fetch AMOs)
* Blocking get/g/iget are not ordered (related: fence orders blocking get/g/iget)
* fence also orders non-blocking get/g/iget
* Data-movement collective APIs using same psync are ordered in the order of their issue

2) reads from
* All communication APIs progress without requiring quiet or barrier
* Using shmem_put/iput/p to trigger shmem_wait_until is platform/implementation defined (related: shmem_atomic_set should be used to trigger shmem_wait_until)
* New APIs to trigger shmem_wait_until that need to be single-copy atomic but not read-modify-write atomic
* Trigger shmem_wait_until using the same type^
* wait_until on remote symmetric memory
* wait_until on local non-symmetric memory (related : when non-blocking fetch AMOs trigger wait_until, does it require read-modify-write-fetch to be atomic?)

3) happens before
None yet

anshumang:
@spotluri @jdinan @manjugv @nspark @minsii @khamidouche and others
Please feel free to add if anything is missing. Plan is to create separate issues for tracking each of the items above and then create PRs for the proposed changes to the spec.

shamisp:
1. Using shmem_wait_until in combination with AMO only is difficult one. 
It somewhat simple if you only looking at shared memory use case only. It is much more complicated if initiator of AMO is located in different coherency domain in respect to target running waituntil loop. The local load operation may not be atomic in respect to remote atomics.

2. AMO is expensive operation when it is compared regular PUT. It is slower and limited in number of outstanding operations. On the other hand waking up a remote PE with regular PUT even partial one can be perfectly fine way of notification.

anshumang:
Definition of concurrency used in #204 to be resolved in this.

anshumang:
Have renamed (can be improved) to distinguish from #172

minsii:
Is the related statement a mistake (get/g/iget ->put/p/iput) ?
> - Blocking get/g/iget are not ordered (related: fence orders blocking get/g/iget)

[...]

PR #201, opened 2018-02-05
Symmetric Heap: Clarify size and alignment semantics

jdinan:
In symmetric heap management routines, clarify requirements on size and alignment arguments and resulting behavior.

 * a `size` of zero results in no action and returns a null pointer
 * `alignment` must be a power-of-two multiple of `sizeof(void*)` or behavior is undefined
 * symmetric heap management routines that perform no action also perform no barriers

Closes #76 
Closes #78

PR #259, opened 2018-11-26
Adding shmem_malloc_with_hints interface

manjugv:
Proposal to address the issue https://github.com/openshmem-org/specification/issues/258 .

### DIFF
--- a/content/shmem_malloc_hints.tex
+++ b/content/shmem_malloc_hints.tex
@@ -27,6 +27,7 @@ void *@\FuncDecl{shmem\_malloc\_with\_hints}@(size_t size, long hints);
     In addition to the \VAR{size} argument, the \VAR{hints} argument is provided by the user. 
     The \VAR{hints} describes the expected manner in which the \openshmem program may use the allocated memory.
     The valid usage hints are described in Table~\ref{usagehints}. Multiple hints may be requested by combining them with a bitwise \CONST{OR} operation.
+	A zero option can be given if no options are requested.
     
     The information provided by the \VAR{hints} is used to optimize for performance by the implementation. 
     If the implementation cannot optimize, the behavior is same as \FUNC{shmem\_malloc}.
@@ -61,15 +62,20 @@ void *@\FuncDecl{shmem\_malloc\_with\_hints}@(size_t size, long hints);
 	\tabularnewline \hline
 	\endhead
 	%%
-	\LibConstDecl{SHMEM\_HINT\_NONE} &
+	\newline
+	\CONST{0} &
+	\newline
 	Behavior same as \FUNC{shmem\_malloc}
 	\tabularnewline \hline
+	
 		
 	\LibConstDecl{SHMEM\_HINT\_ATOMICS} &
+	\newline 
 	Memory used for \VAR{Atomic} operations
 	\tabularnewline \hline
 	
 	\LibConstDecl{SHMEM\_HINT\_SIGNAL} &
+	\newline
 	Memory used for \VAR{signal} operations
 	\tabularnewline \hline
 
--- a/content/shmem_malloc_hints.tex
+++ b/content/shmem_malloc_hints.tex
@@ -22,7 +22,7 @@ void *@\FuncDecl{shmem\_malloc\_with\_hints}@(size_t size, long hints);
     \VAR{size} bytes, which shall be suitably aligned so that it may be
     assigned to a pointer to any type of object.  This space is allocated from
     the symmetric heap (similar to \FUNC{shmem\_malloc}).  When the \VAR{size} is zero, 
-    the \FUNC{shmem\_malloc\_with\_hint} routine performs no action and returns a null pointer. 
+    the \FUNC{shmem\_malloc\_with\_hints} routine performs no action and returns a null pointer. 
     
     In addition to the \VAR{size} argument, the \VAR{hints} argument is provided by the user. 
     The \VAR{hints} describes the expected manner in which the \openshmem program may use the allocated memory.
@@ -33,7 +33,7 @@ void *@\FuncDecl{shmem\_malloc\_with\_hints}@(size_t size, long hints);
     If more than one hint is provided, the implementation will make the best effort to use one or more hints 
     to optimize performance. 
             
-    The \FUNC{shmem\_malloc\_with\_hint} routine is provided  so that multiple \acp{PE} in a program can allocate symmetric,
+    The \FUNC{shmem\_malloc\_with\_hints} routine is provided  so that multiple \acp{PE} in a program can allocate symmetric,
     remotely accessible memory blocks.  When no action is performed, these
     routines return without performing a barrier. Otherwise, the routine will call a barrier on exit.
     This ensures that all \acp{PE} participate in the memory allocation, and that the memory on other
@@ -65,22 +65,6 @@ void *@\FuncDecl{shmem\_malloc\_with\_hints}@(size_t size, long hints);
 	Behavior same as \FUNC{shmem\_malloc}
 	\tabularnewline \hline
 		
-	\LibConstDecl{SHMEM\_HINT\_LOW\_LAT\_MEM} &
-	Allocate memory on low-latency storage	
-	\tabularnewline \hline
-	
-	\LibConstDecl{SHMEM\_HINT\_HIGH\_BW\_MEM} &
-	Allocate memory on high-bandwidth storage	
-	\tabularnewline \hline
-
-	\LibConstDecl{SHMEM\_HINT\_PSYNC} &
-	Memory used as \CONST{PSYNC} array
-	\tabularnewline \hline
-		
-	\LibConstDecl{SHMEM\_HINT\_PWORK} &
-	Memory used as \CONST{PWORK} array
-	\tabularnewline \hline
-	
 	\LibConstDecl{SHMEM\_HINT\_ATOMICS} &
 	Memory used for \VAR{Atomic} operations
 	\tabularnewline \hline
--- a/content/shmem_malloc_hints.tex
+++ b/content/shmem_malloc_hints.tex
@@ -80,7 +80,7 @@ void *@\FuncDecl{shmem\_malloc\_with\_hints}@(size_t size, long hints);
 
 \apinotes{
 		The \openshmem programs should allocate memory with
-		\CONST{SHMEM\_HINT\_ATOMIC\_REMOTE}, when the majority of
+		\CONST{SHMEM\_HINT\_ATOMICS\_REMOTE}, when the majority of
 		operations performed on this memory are atomic operations, and origin
 		and target \ac{PE} of the atomic operations do not share a memory domain
 		.i.e., symmetric objects on the target \ac{PE} is not accessible using
--- a/content/shmem_malloc_hints.tex
+++ b/content/shmem_malloc_hints.tex
@@ -79,10 +79,6 @@ void *@\FuncDecl{shmem\_malloc\_with\_hints}@(size_t size, long hints);
 	Allocate memory on storage near to the network card	
 	\tabularnewline \hline
 	
-	\LibConstDecl{SHMEM\_HINT\_DEVICE\_GPU\_MEM} &
-	Allocate memory on the device	
-	\tabularnewline \hline
-	
 	\LibConstDecl{SHMEM\_HINT\_DEVICE\_NIC\_MEM} &
 	Allocate memory on the network interface card 
 	\tabularnewline \hline
--- a/content/shmem_malloc_hints.tex
+++ b/content/shmem_malloc_hints.tex
@@ -1,0 +1,116 @@
+
+\apisummary{
+    Collective memory allocation routine with support for providing hints.
+}
+
+\begin{apidefinition}
+
+\begin{Csynopsis}
+void *@\FuncDecl{shmem\_malloc\_with\_hints}@(size_t size, long hints);
+\end{Csynopsis}
+
+\begin{apiarguments}
+    \apiargument{IN}{size}{The size, in bytes, of a block to be
+        allocated from the symmetric heap. This argument is of type \CTYPE{size\_t}}
+    \apiargument{IN}{hints}{A bit array of hints provided by the user to the implementation}
+\end{apiarguments}
+
+
+\apidescription{
+
+    The \FUNC{shmem\_malloc\_with\_hint} like \FUNC{shmem\_malloc} routine returns a pointer to a block of at least
+    \VAR{size} bytes, which shall be suitably aligned so that it may be
+    assigned to a pointer to any type of object.  This space is allocated from
+    the symmetric heap (similar to \FUNC{shmem\_malloc}).  When \VAR{size} is zero, 
+    the \FUNC{shmem\_malloc\_with\_hint} routine performs no action and returns a null pointer. 
+    
+    In addition to the \VAR{size} argument, the \VAR{hint} argument is provided by the user. 
+    The \VAR{hint} could describe the trait of memory storage where memory is allocated or 
+    it could describe the expected manner in which the \openshmem program may use the allocated memory.
+    The valid storage trait hints are described in Table~\ref{traithints} and the valid usage hints 
+    are described in Table~\ref{usagehints}. Multiple hints are expressed as \CONST{OR} of \VAR{hints}.
+    
+     The information provided by the \VAR{hint} is used to optimize for performance by the implementation. 
+    If the implementation cannot optimize, the behavior is same as \FUNC{shmem\_malloc}.
+    If more than one hint is provided, the implementation will make the best effort to use one or more hints 
+    to optimize performance. 
+            
+    The \FUNC{shmem\_malloc\_with\_hint} routine is provided  so that multiple \acp{PE} in a program can allocate symmetric,
+    remotely accessible memory blocks.  When no action is performed, these
+    routines return without performing a barrier. Otherwise, the routine will call a barrier on exit.
+    This ensures that all \acp{PE} participate in the memory allocation, and that the memory on other
+    \acp{PE} can be used as soon as the local \ac{PE} returns. The implicit barrier performed by this routine will quiet the
+    default context.  It is the user's responsibility to ensure that no communication operations involving the given memory block are pending on
+    other contexts prior to calling the \FUNC{shmem\_free} and \FUNC{shmem\_realloc} routines.
+    The user is also responsible for calling these routines with identical argument(s) on all
+    \acp{PE}; if differing \VAR{size}, or \VAR{hint} arguments are used, the behavior of the call
+    and any subsequent \openshmem calls is undefined.
+}
+
+\apireturnvalues{
+    The \FUNC{shmem\_malloc\_with\_hint} routine returns a pointer to the allocated space;
+    otherwise, it returns a null pointer.
+}
+
+\apinotes{ 
+}
+
+\apiimpnotes{
+}
+    \begin{longtable}{|p{0.45\textwidth}|p{0.5\textwidth}|}
+	\hline
+	\textbf{Hints} & \textbf{Storage Trait}
+	\tabularnewline \hline
+	\endhead
+	%%
+	\LibConstDecl{SHMEM\_HINT\_NONE} &
+	Behavior same as shmem\_malloc
+	\tabularnewline \hline
+		
+	\LibConstDecl{SHMEM\_HINT\_LOW\_LAT\_MEM} &
+	Allocate memory on low-latency storage	
+	\tabularnewline \hline
+	
+	\LibConstDecl{SHMEM\_HINT\_HIGH\_BW\_MEM} &
+	Allocate memory on high-bandwidth storage	
+	\tabularnewline \hline
+
+	\LibConstDecl{SHMEM\_HINT\_NEAR\_NIC\_MEM} &
+	Allocate memory on storage near to the network card	
+	\tabularnewline \hline
+	
+	\LibConstDecl{SHMEM\_HINT\_DEVICE\_GPU\_MEM} &
+	Allocate memory on the device	
+	\tabularnewline \hline
+	
+	\LibConstDecl{SHMEM\_HINT\_DEVICE\_NIC\_MEM} &
+	Allocate memory on the network interface card 
+	\tabularnewline \hline
+        \TableCaptionRef{Storage trait hints}
+        \label{traithints}
+    \end{longtable}
+   
+
+    \begin{longtable}{|p{0.45\textwidth}|p{0.5\textwidth}|}
+	\hline
+	\textbf{Hints} & \textbf{Usage hint}
+	\tabularnewline \hline
+	\endhead
+	%%
+	\LibConstDecl{SHMEM\_HINT\_PSYNC} &
+	Memory used as \CONST{PSYNC} array
+	\tabularnewline \hline
+		
+	\LibConstDecl{SHMEM\_HINT\_PWORK} &
+	Memory used as \CONST{PWORK} array
+	\tabularnewline \hline
+	
+	\LibConstDecl{SHMEM\_HINT\_ATOMICS} &
+	 Memory used for \VAR{Atomic} operations	
+	\tabularnewline \hline
+	
+	\TableCaptionRef{Memory usage hints}
+        \label{usagehints}
+    \end{longtable}
+\end{apidefinition}
+\newpage
--- a/content/shmem_malloc_hints.tex
+++ b/content/shmem_malloc_hints.tex
@@ -25,12 +25,10 @@ void *@\FuncDecl{shmem\_malloc\_with\_hints}@(size_t size, long hints);
     the \FUNC{shmem\_malloc\_with\_hint} routine performs no action and returns a null pointer. 
     
     In addition to the \VAR{size} argument, the \VAR{hint} argument is provided by the user. 
-    The \VAR{hint} could describe the trait of memory storage where memory is allocated or 
-    it could describe the expected manner in which the \openshmem program may use the allocated memory.
-    The valid storage trait hints are described in Table~\ref{traithints} and the valid usage hints 
-    are described in Table~\ref{usagehints}. Multiple hints are expressed as \CONST{OR} of \VAR{hints}.
+    The \VAR{hint} describes the expected manner in which the \openshmem program may use the allocated memory.
+    The valid usage hints are described in Table~\ref{usagehints}. Multiple hints are expressed as \CONST{OR} of \VAR{hints}.
     
-     The information provided by the \VAR{hint} is used to optimize for performance by the implementation. 
+    The information provided by the \VAR{hint} is used to optimize for performance by the implementation. 
     If the implementation cannot optimize, the behavior is same as \FUNC{shmem\_malloc}.
     If more than one hint is provided, the implementation will make the best effort to use one or more hints 
     to optimize performance. 
@@ -57,9 +55,9 @@ void *@\FuncDecl{shmem\_malloc\_with\_hints}@(size_t size, long hints);
 
 \apiimpnotes{
 }
-    \begin{longtable}{|p{0.45\textwidth}|p{0.5\textwidth}|}
+	\begin{longtable}{|p{0.45\textwidth}|p{0.5\textwidth}|}
 	\hline
-	\textbf{Hints} & \textbf{Storage Trait}
+	\textbf{Hints} & \textbf{Usage hint}
 	\tabularnewline \hline
 	\endhead
 	%%
@@ -75,24 +73,6 @@ void *@\FuncDecl{shmem\_malloc\_with\_hints}@(size_t size, long hints);
 	Allocate memory on high-bandwidth storage	
 	\tabularnewline \hline
 
-	\LibConstDecl{SHMEM\_HINT\_NEAR\_NIC\_MEM} &
-	Allocate memory on storage near to the network card	
-	\tabularnewline \hline
-	
-	\LibConstDecl{SHMEM\_HINT\_DEVICE\_NIC\_MEM} &
-	Allocate memory on the network interface card 
-	\tabularnewline \hline
-        \TableCaptionRef{Storage trait hints}
-        \label{traithints}
-    \end{longtable}
-   
-
-    \begin{longtable}{|p{0.45\textwidth}|p{0.5\textwidth}|}
-	\hline
-	\textbf{Hints} & \textbf{Usage hint}
-	\tabularnewline \hline
-	\endhead
-	%%
 	\LibConstDecl{SHMEM\_HINT\_PSYNC} &
 	Memory used as \CONST{PSYNC} array
 	\tabularnewline \hline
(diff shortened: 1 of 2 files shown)


manjugv:
I think this is orthogonal to  PR #256 and there is value in co-existing with the PR #256 .

naveen-rn:
If I understand correctly in PR #256 we allocate the whole SHEAP using hints and then use ```shmem_kind_malloc``` for allocating SDO from the previously allocated SHEAP. In this PR, we are not controlling the SHEAP at all, it is just that during the ```shmem_malloc``` operation - we try to optimally place the SDO based on user hints.

manjugv:
@naveen-rn Correct. Please refer https://github.com/openshmem-org/specification/issues/258 .

content/shmem_malloc_hints.tex
```diff
@@ -0,0 +1,116 @@
...
+    \VAR{size} bytes, which shall be suitably aligned so that it may be
+    assigned to a pointer to any type of object.  This space is allocated from
+    the symmetric heap (similar to \FUNC{shmem\_malloc}).  When \VAR{size} is zero, 
+    the \FUNC{shmem\_malloc\_with\_hint} routine performs no action and returns a null pointer. 
+    
+    In addition to the \VAR{size} argument, the \VAR{hint} argument is provided by the user. 
+    The \VAR{hint} could describe the trait of memory storage where memory is allocated or 
+    it could describe the expected manner in which the \openshmem program may use the allocated memory.
+    The valid storage trait hints are described in Table~\ref{traithints} and the valid usage hints 
+    are described in Table~\ref{usagehints}. Multiple hints are expressed as \CONST{OR} of \VAR{hints}.
```
hoopoepg: @manjugv 
question regarding ```Multiple hints are expressed as OR of hints.```: how oshmem should react on case if supported less flags than requested? or supported all honts, but not at same buffer (and I believe this case - supported all but not at same buffer - is the most actual)?

it seems ```OR``` combination of ```traithints``` hints should be denied

manjugv: @hoopoepg That is correct. In the last sentence of this paragraph I say that. Some combination of hints does not make sense and some combination of hints might not be supported by a particular implementation. In either case, it can behave like shmem_malloc.

naveen-rn:
On the core functionality of this features - do we need the hints to have both the storage traits and possible usage pattern? Can we just stick to the possible usage pattern, say SHMEM_HINT_AMO, SHMEM_HINT_SIGNAL? Something to discuss further.

content/shmem_malloc_hints.tex
```diff
@@ -0,0 +1,97 @@
...
+    The valid usage hints are described in Table~\ref{usagehints}. Multiple hints are expressed as \CONST{OR} of \VAR{hints}.
+    
+    The information provided by the \VAR{hint} is used to optimize for performance by the implementation. 
+    If the implementation cannot optimize, the behavior is same as \FUNC{shmem\_malloc}.
+    If more than one hint is provided, the implementation will make the best effort to use one or more hints 
+    to optimize performance. 
+            
+    The \FUNC{shmem\_malloc\_with\_hint} routine is provided  so that multiple \acp{PE} in a program can allocate symmetric,
+    remotely accessible memory blocks.  When no action is performed, these
+    routines return without performing a barrier. Otherwise, the routine will call a barrier on exit.
```
jamesaross: > When no action is performed, these routines return without performing a barrier.

What does this mean? That the function returns NULL for all PEs, no memory has been allocated, and no implicit barrier has occurred?  Some applications may use `shmem_malloc` and friends as implicit barriers.  This proposal is for an optimization which seems to be a drop in replacement for vanilla `shmem_malloc`, but dropping the implicit barrier on function exit would change the behavior of the application.

content/shmem_malloc_hints.tex
```diff
@@ -0,0 +1,97 @@
...
+    
+    In addition to the \VAR{size} argument, the \VAR{hint} argument is provided by the user. 
+    The \VAR{hint} describes the expected manner in which the \openshmem program may use the allocated memory.
+    The valid usage hints are described in Table~\ref{usagehints}. Multiple hints are expressed as \CONST{OR} of \VAR{hints}.
+    
+    The information provided by the \VAR{hint} is used to optimize for performance by the implementation. 
+    If the implementation cannot optimize, the behavior is same as \FUNC{shmem\_malloc}.
+    If more than one hint is provided, the implementation will make the best effort to use one or more hints 
+    to optimize performance. 
+
```
jamesaross: Must all PEs allocate from the same special memory?  It's not clear if asymmetry can exist. Does this impose additional implicit synchronization for each subset configuration of hints if it cannot satisfy the entire hint list?

Also, what happens if you OR `SHMEM_HINT_NONE` with other hint behaviors?

content/shmem_malloc_hints.tex
manjugv: When no allocation is done, “dropping the implicit barrier” is the behavior we have for shmem_malloc in OpenSHMEM 1.4 - please refer page 26 line 40-41. The proposal is aiming to maintain the same behavior for that case.

content/shmem_malloc_hints.tex
manjugv: Good question. 

All use cases that I have thought of requires the memory to be symmetric (same kind of memory). 

Regarding extra synchronization, it depends on the implementation. If the implementations maintain asymmetric memory sizes (say each PE starts with different amount of special memory) on the PEs, you might need the extra synchronization for agreement. Otherwise, I do not see a need. In a way, it is similar to current DRAM allocations. Also, for the implementation we explored, we did not need extra synchronization.

I’m reluctant to add such a constraint. Without such constraint, the implementations are free to explore either approach (symmetric and asymmetric). 

Do see value in specifying one way or other?

content/shmem_malloc_hints.tex
jamesaross: Oops.  In OpenSHMEM 1.4, that was not the case.  This behavior was changed between 1.4 and now in #201.

content/shmem_malloc_hints.tex
jamesaross: To clarify, what would using `long hint = SHMEM_HINT_NONE | SHMEM_HINT_LOW_LAT_MEM | SHMEM_HINT_HIGH_BW_MEM` do?

Dropping `SHMEM_HINT_NONE`, what if the platform could provide `SHMEM_HINT_LOW_LAT_MEM` or `SHMEM_HINT_HIGH_BW_MEM` but not both simultaneously?  Will we see application code marked up like this because who doesn't want to use low latency and high bandwidth memory for their application? Does the "best effort" default to a platform-specific precedence? The only feedback is that an allocation succeeded or it did not.

manjugv: > To clarify, what would using long hint = SHMEM_HINT_NONE | SHMEM_HINT_LOW_LAT_MEM | SHMEM_HINT_HIGH_BW_MEM do?

Though this is a legal usage, it does not make sense to use. The implementations are allowed to default to shmem_malloc in this case. 

> Dropping SHMEM_HINT_NONE, what if the platform could provide SHMEM_HINT_LOW_LAT_MEM or SHMEM_HINT_HIGH_BW_MEM but not both simultaneously? Will we see application code marked up like this because who doesn't want to use low latency and high bandwidth memory for their application? Does the "best effort" default to a platform-specific precedence? The only feedback is that an allocation succeeded or it did not.

```
“If more than one hint is provided, the implementation will make the best effort to use one or more hints to optimize performance. 
“
```
My intention with this statement was to provide flexibility for the implementations to optimize as they wish when the user provides multiple hints. Obviously, some combinations of hints might not make sense. In such cases, If the implementations want to give precedence of one hint over others, the proposal allows it. That (assigning priorities to hints) is one way to implement it, but not the only way.

content/shmem_malloc_hints.tex
manjugv: Ah… Thanks for correcting that.  It has been so long that we debated about this I did not realize it is relatively new. I was looking at my git copy. :) 

Can we consider this issue resolved?

jamesaross:
Assuming that the #280 Teams Proposal makes it through for 1.5, won't all references to active set synchronization memory and reduction working memory be deprecated?  If so, would SHMEM_HINT_PSYNC, SHMEM_HINT_PWORK also be deprecated before they become part of the specification?  There seems to be a consistency problem with these two proposals.

Would SHMEM_HINT_SIGNAL (from #275) and SHMEM_HINT_ATOMIC have any different/distinct behavior in an implementation?  It seems most code (and I believe all of the examples in the specification) use global or static variables for atomics and signals rather than shmem_malloc'ed memory.  Should it be recommended to use shmem_malloc_with hints over static/global?

Often the lowest latency memory pools have the highest bandwidth (with some exceptions).  Are we relying on users to know whether their application needs lower latency or higher bandwidth and to understand which pools the flags implicitly allocate from?

What about very large memory pools of slow memory or mapped storage when latency/bandwidth aren't issues (i.e. SHMEM_HINT_BIGMEM)?

naveen-rn:
> Would SHMEM_HINT_SIGNAL (from #275) and SHMEM_HINT_ATOMIC have any different/distinct behavior in an implementation?

Yes, depends on the implementation - there might be a difference.

manjugv:
> Assuming that the #280 Teams Proposal makes it through for 1.5, won't all references to active set synchronization memory and reduction working memory be deprecated? If so, would SHMEM_HINT_PSYNC, SHMEM_HINT_PWORK also be deprecated before they become part of the specification? There seems to be a consistency problem with these two proposals.

Yes, I agree with your observation. One way to approach such inconsistencies is via chapter committee. In this particular instance, I’m ok not to introduce it or resolve it via chapter committee.

manjugv:
> Would SHMEM_HINT_SIGNAL (from #275) and SHMEM_HINT_ATOMIC have any different/distinct behavior in an implementation? 

Since the signal has a separate interface to fetch the signal value and it is atomic only with itself, it provides more\different opportunities for optimization. I can envision implementing them in different ways based on the capabilities of a particular network. On some systems, it might be the same.

manjugv:
> What about very large memory pools of slow memory or mapped storage when latency/bandwidth aren't issues (i.e. SHMEM_HINT_BIGMEM)?

In the usage scenarios that I have come across, we have not mapped symmetric memory to “slow” memory. So, I have not thought about such a hint. That said, it is a valid usage scenario. So, let's discuss this during the reading and see what rest of the committee thinks about adding it.

manjugv:
> It seems most code (and I believe all of the examples in the specification) use global or static variables for atomics and signals rather than shmem_malloc'ed memory. Should it be recommended to use shmem_malloc_with hints over static/global?

In general, managing global/static memory for global access via OpenSHMEM interfaces presents more constraints and thus fewer optimization opportunities. However, I’m still hesitant to provide a blanket recommendation because the performance depends on the system and implementation. 

We could add some examples where we can show atomics allocated via shmem_malloc interfaces, if it is useful.

manjugv:
> Often the lowest latency memory pools have the highest bandwidth (with some exceptions). Are we relying on users to know whether their application needs lower latency or higher bandwidth and to understand which pools the flags implicitly allocate from?

That is correct.

content/shmem_malloc_hints.tex
```diff
@@ -0,0 +1,97 @@
...
+	\textbf{Hints} & \textbf{Usage hint}
+	\tabularnewline \hline
+	\endhead
+	%%
+	\LibConstDecl{SHMEM\_HINT\_NONE} &
+	Behavior same as shmem\_malloc
+	\tabularnewline \hline
+		
+	\LibConstDecl{SHMEM\_HINT\_LOW\_LAT\_MEM} &
+	Allocate memory on low-latency storage
```
shamisp: Low-latency relatively to what ? Low latency for local access ? remote access ? both ? only one of those ? Why I would ever what to allocate memory with low-latency and (related to next flag) slow bandwidth ?

content/shmem_malloc_hints.tex
```diff
@@ -0,0 +1,97 @@
...
+		
+	\LibConstDecl{SHMEM\_HINT\_LOW\_LAT\_MEM} &
+	Allocate memory on low-latency storage	
+	\tabularnewline \hline
+	
+	\LibConstDecl{SHMEM\_HINT\_HIGH\_BW\_MEM} &
+	Allocate memory on high-bandwidth storage	
+	\tabularnewline \hline
+
+	\LibConstDecl{SHMEM\_HINT\_PSYNC} &
```
shamisp: why we introducing tuning for something that we are deprecating ?

content/shmem_malloc_hints.tex
```diff
@@ -0,0 +1,97 @@
...
+	Memory used as \CONST{PSYNC} array
+	\tabularnewline \hline
```
shamisp: why we introducing tuning for something that we are deprecating ?

content/shmem_malloc_hints.tex
```diff
@@ -0,0 +1,97 @@
...
+		
+	\LibConstDecl{SHMEM\_HINT\_PWORK} &
+	Memory used as \CONST{PWORK} array
+	\tabularnewline \hline
+	
+	\LibConstDecl{SHMEM\_HINT\_ATOMICS} &
```
shamisp: What happens with ATOMIC hint was not passed but atomic operation was used ?

content/shmem_malloc_hints.tex
```diff
@@ -0,0 +1,97 @@
...
+	Memory used for \VAR{Atomic} operations
+	\tabularnewline \hline
+	
+	\LibConstDecl{SHMEM\_HINT\_SIGNAL} &
+	Memory used for \VAR{signal} operations
```
shamisp: What happens when the flag was not used, but signal operation was used ?

content/shmem_malloc_hints.tex
```diff
@@ -0,0 +1,97 @@
...
+	\tabularnewline \hline
+
+	\TableCaptionRef{Memory usage hints}
+        \label{usagehints}
+    \end{longtable}
```
shamisp: Also we should to state that neither alignment requirements or memory properties, such as cache line size are not get impacted by the hint. We also shell state the memory semantics for local assess (load and store) are not impacted by the hint.

content/shmem_malloc_hints.tex
manjugv: > Low-latency relatively to what ? Low latency for local access ? remote access ? both ?

This is symmetric heap’s memory. So, it is for the memory accessed via shmem routines (which I would guess will be dominated by remote access).

manjugv: > Why I would ever what to allocate memory with low-latency and (related to next flag) slow bandwidth ?

Already discussed here. Please refer to this discussion (https://github.com/openshmem-org/specification/pull/259#discussion_r320468747)

content/shmem_malloc_hints.tex
manjugv: This is always a challenge with tickets (that are developed in parallel) that has impact on each other. 

Already discussed here. Please refer to this discussion (https://github.com/openshmem-org/specification/pull/259#issuecomment-530136683)

content/shmem_malloc_hints.tex
manjugv: Already discussed here. Please refer to this discussion (https://github.com/openshmem-org/specification/pull/259#issuecomment-530136683)

content/shmem_malloc_hints.tex
manjugv: I don't explicitly cover the bad behavior semantics. But, based on the use case experience we have currently, I'm inclined to say that It should not impact correctness. However, the programs might not be able to get the best performance characteristics for that environment.

Do you prefer that we should explicitly say this?

content/shmem_malloc_hints.tex
manjugv: Hmm .. Which flag are you referring to, can you please elaborate?

content/shmem_malloc_hints.tex
manjugv: AFAIU, for the shmem_malloc routine the implementation is free to allocate the memory, which is either cache aligned or not. One of the constraints is that it should be word-aligned. Similarly, the memory access model (which is yet be defined or clarified here https://github.com/openshmem-org/specification/issues/229) will provide certain access guarantees to the memory allocated by shmem_malloc. In both cases, the proposal intends to follow the semantics of shmem_malloc.

content/shmem_malloc_hints.tex
shamisp: Above `SHMEM_HINT_SIGNAL`

content/shmem_malloc_hints.tex
shamisp: The memory is access locally using load and store semantics without shmem. This is how you initialize  the memory with some data in a first place. It is also accessed remotely using load and store semantics without going through SHMEM API.  My point the we cannot claim that something is low latency (the same true for high BW and most of other hints) since it really depends on who is accessing the data and this runtime information. For example, for Nic it can be some local sram region on the die and for the socket it L1/L2/L3. If you want to allocate memory on the nic for low latency access, just call it NIC memory, GPU memory, L1 memory ,etc. The low latency and high bw definitions are misleading.

content/shmem_malloc_hints.tex
shamisp: My initial assumption is that if this is hint, implementation can completely ignore it. But we still have issue with this semantics. For HCA optimal location of the memory for Atomics will be on the HCA. For core it will be optimal to have in L1/L2 which is suboptimal for ATOMICs coming from the network and vice-versa. Essentially such hint may actually do more  harm to applications. My suggestion is to use explicit names - HCA1, HCA2, GPU, L2, LLC, etc. This will provide very clear description to user what actually happens.

content/shmem_malloc_hints.tex
shamisp: The memory gets eventually mapped into the core. The core architecture defined multiple way how it can be mapped. Each on of the mapping has own semantics and constrains. For example semantics between normal cacheable (write back) and non-cacheable (WC) is very different. Since user has direct access through the pointer, code that worked on one machine will break on another with exception or even data corruption.

shamisp: In order to support something like this, you have to remove shmem_ptr and prohibit any direct asses to the memory through load and store semantics. Next you would have to introduce shmem_memcpy function to copy-in-out shmem_malloced region.

content/shmem_malloc_hints.tex
```diff
@@ -0,0 +1,97 @@
...
+
+\apidescription{
+
+    The \FUNC{shmem\_malloc\_with\_hint} like \FUNC{shmem\_malloc} routine returns a pointer to a block of at least
+    \VAR{size} bytes, which shall be suitably aligned so that it may be
+    assigned to a pointer to any type of object.  This space is allocated from
+    the symmetric heap (similar to \FUNC{shmem\_malloc}).  When \VAR{size} is zero, 
+    the \FUNC{shmem\_malloc\_with\_hint} routine performs no action and returns a null pointer. 
+    
+    In addition to the \VAR{size} argument, the \VAR{hint} argument is provided by the user.
```
nspark:
```suggestion
    In addition to the \VAR{size} argument, the \VAR{hints} argument is provided by the user. 
```

content/shmem_malloc_hints.tex
```diff
@@ -0,0 +1,97 @@
...
+\begin{apiarguments}
+    \apiargument{IN}{size}{The size, in bytes, of a block to be
+        allocated from the symmetric heap. This argument is of type \CTYPE{size\_t}}
+    \apiargument{IN}{hints}{A bit array of hints provided by the user to the implementation}
+\end{apiarguments}
+
+
+\apidescription{
+
+    The \FUNC{shmem\_malloc\_with\_hint} like \FUNC{shmem\_malloc} routine returns a pointer to a block of at least
```
nspark:
```suggestion
    The \FUNC{shmem\_malloc\_with\_hint} routine, like \FUNC{shmem\_malloc}, returns a pointer to a block of at least
```

content/shmem_malloc_hints.tex
```diff
@@ -0,0 +1,97 @@
...
+    \VAR{size} bytes, which shall be suitably aligned so that it may be
+    assigned to a pointer to any type of object.  This space is allocated from
+    the symmetric heap (similar to \FUNC{shmem\_malloc}).  When \VAR{size} is zero, 
+    the \FUNC{shmem\_malloc\_with\_hint} routine performs no action and returns a null pointer. 
+    
+    In addition to the \VAR{size} argument, the \VAR{hint} argument is provided by the user. 
+    The \VAR{hint} describes the expected manner in which the \openshmem program may use the allocated memory.
```
nspark:
```suggestion
    The \VAR{hints} describes the expected manner in which the \openshmem program may use the allocated memory.
```

content/shmem_malloc_hints.tex
```diff
@@ -0,0 +1,97 @@
...
+    The valid usage hints are described in Table~\ref{usagehints}. Multiple hints are expressed as \CONST{OR} of \VAR{hints}.
+    
+    The information provided by the \VAR{hint} is used to optimize for performance by the implementation.
```
nspark:
```suggestion
    The information provided by the \VAR{hints} is used to optimize for performance by the implementation. 
```

content/shmem_malloc_hints.tex
```diff
@@ -0,0 +1,97 @@
...
+            
+    The \FUNC{shmem\_malloc\_with\_hint} routine is provided  so that multiple \acp{PE} in a program can allocate symmetric,
+    remotely accessible memory blocks.  When no action is performed, these
+    routines return without performing a barrier. Otherwise, the routine will call a barrier on exit.
+    This ensures that all \acp{PE} participate in the memory allocation, and that the memory on other
+    \acp{PE} can be used as soon as the local \ac{PE} returns. The implicit barrier performed by this routine will quiet the
+    default context.  It is the user's responsibility to ensure that no communication operations involving the given memory block are pending on
+    other contexts prior to calling the \FUNC{shmem\_free} and \FUNC{shmem\_realloc} routines.
+    The user is also responsible for calling these routines with identical argument(s) on all
+    \acp{PE}; if differing \VAR{size}, or \VAR{hint} arguments are used, the behavior of the call
```
nspark:
```suggestion
    \acp{PE}; if differing \VAR{size}, or \VAR{hints} arguments are used, the behavior of the call
```

content/shmem_malloc_hints.tex
```diff
@@ -0,0 +1,97 @@
...
+
+    The \FUNC{shmem\_malloc\_with\_hint} like \FUNC{shmem\_malloc} routine returns a pointer to a block of at least
+    \VAR{size} bytes, which shall be suitably aligned so that it may be
+    assigned to a pointer to any type of object.  This space is allocated from
+    the symmetric heap (similar to \FUNC{shmem\_malloc}).  When \VAR{size} is zero, 
+    the \FUNC{shmem\_malloc\_with\_hint} routine performs no action and returns a null pointer. 
+    
+    In addition to the \VAR{size} argument, the \VAR{hint} argument is provided by the user. 
+    The \VAR{hint} describes the expected manner in which the \openshmem program may use the allocated memory.
+    The valid usage hints are described in Table~\ref{usagehints}. Multiple hints are expressed as \CONST{OR} of \VAR{hints}.
```
nspark:
```suggestion
    The valid usage hints are described in Table~\ref{usagehints}. Multiple hints may be requested by combining them with a bitwise \CONST{OR} operation.
```

content/shmem_malloc_hints.tex
```diff
@@ -0,0 +1,97 @@
...
+\apiimpnotes{
+}
+	\begin{longtable}{|p{0.45\textwidth}|p{0.5\textwidth}|}
+	\hline
+	\textbf{Hints} & \textbf{Usage hint}
+	\tabularnewline \hline
+	\endhead
+	%%
+	\LibConstDecl{SHMEM\_HINT\_NONE} &
+	Behavior same as shmem\_malloc
```
nspark:
```suggestion
	Behavior same as \FUNC{shmem\_malloc}
```

content/shmem_malloc_hints.tex
```diff
@@ -0,0 +1,97 @@
...
+	\tabularnewline \hline
+		
+	\LibConstDecl{SHMEM\_HINT\_LOW\_LAT\_MEM} &
+	Allocate memory on low-latency storage	
+	\tabularnewline \hline
+	
+	\LibConstDecl{SHMEM\_HINT\_HIGH\_BW\_MEM} &
+	Allocate memory on high-bandwidth storage
```
jdinan: To capture my preference: "Allocate memory with high bandwidth for local access by the calling PE"

content/shmem_malloc_hints.tex
shamisp: Since in this case you optimize for remote PEs only, I think it should be named SHMEM\_HINT\_REMOTE\_ATOMICS to be clear that you don't optimize for local PEs. Without explicit specification local/remote the hint is not very useful.

manjugv:
Special ballot changes: 

1. Change the name of the hint from SHMEM_HINT_ATOMICS to SHMEM_HINT_ATOMICS_REMOTE (https://github.com/openshmem-org/specification/pull/259/commits/755a4ed9c879defd2772db3a3d18ddb62d648537)
2. Change option SHMEM_HINT_NONE to “0”, when no hint is desired. (https://github.com/openshmem-org/specification/pull/259/commits/e7d011c55170262b4356aff27a67bebe716cc034)
3. Trim the hints list - remove hint for PSYNC, PWORK, Low latency, and high bandwidth (https://github.com/openshmem-org/specification/pull/259/commits/bf4de2ebdc1d632afc1da11cd3d1cedeabf593c5)

Editorial changes: 
Fixing the typo by changing "hint" to “hints”, when relevant.

jdinan:
Comments from Intel review of PR:

  - The behavior for unrecognized bits in the hints argument should be more clearly specified -- unrecognized bits are ignored, behavior is undefined, routine returns NULL, machine catches fire, etc.
  - There is concern that the routine could be hard to use in situations where users require that hints are recognized.  That is, users can’t find out if hints were used successfully.  One suggestion was to add a `long *hints_out` argument that indicates which hints were utilized (allow NULL for this argument if the user doesn't care).
  - Can the routine return NULL if e.g. a memory allocation could not be performed?  From reviewing the text in `shmem_malloc`, the conditions in which NULL is returned are not clearly specified.
  - Should we add a hints bit SHMEM_HINT_USE_OR_NULL that requests this routine to return NULL if the hints could not be utilized?
  - What is behavior of shmem_realloc on a pointer return by shmem_malloc_with_hints?  Are the hints preserved?  This should be specified (there is precedent with realloc on aligned allocations).
  - Should we have calloc and realloc with hints operations?

jdinan:
@nspark Could you please update (or dismiss) your review so that the PR is no longer flagged as "changes requested"?  Thanks!

content/shmem_malloc_hints.tex
```diff
@@ -0,0 +1,90 @@
...
+        \label{usagehints}
+    \end{longtable}
+
+\apinotes{
+		The \openshmem programs should allocate memory with
+		\CONST{SHMEM\_HINT\_ATOMICS\_REMOTE}, when the majority of
+		operations performed on this memory are atomic operations, and origin
+		and target \ac{PE} of the atomic operations do not share a memory domain
+		.i.e., symmetric objects on the target \ac{PE} is not accessible using
+		load/store operations from the origin \ac{PE} or vice versa.
```
shamisp: Minor: I would consider to move the text under `SHMEM_HINT_ATOMICS_REMOTE` since this is critical information. BTW, probably the same true for remove `SHMEM_HINT_SIGNAL` vs local `SHMEM_HINT_SIGNAL` ? @naveen-rn

naveen-rn: If I understand the definition of remote vs local wrt this change, I'm not sure whether there is need to differentiate the local side to the signal. All signals from the source PE are passed-by-value to the remote signal address buffer.  A buffer created with SHMEM_HINT_SIGNAL will mostly be updated by some remote PE. 
```bash
void shmem_put_signal(shmem_ctx_t ctx, TYPE *dest, const TYPE *source, size_t nelems,
uint64_t *sig_addr, uint64_t signal, int sig_op, int pe);
```

manjugv:
> Comments from Intel review of PR:
> 
> * The behavior for unrecognized bits in the hints argument should be more clearly specified -- unrecognized bits are ignored, behavior is undefined, routine returns NULL, machine catches fire, etc.
> * There is concern that the routine could be hard to use in situations where users require that hints are recognized.  That is, users can’t find out if hints were used successfully.  One suggestion was to add a `long *hints_out` argument that indicates which hints were utilized (allow NULL for this argument if the user doesn't care).
> * Can the routine return NULL if e.g. a memory allocation could not be performed?  From reviewing the text in `shmem_malloc`, the conditions in which NULL is returned are not clearly specified.
> * Should we add a hints bit SHMEM_HINT_USE_OR_NULL that requests this routine to return NULL if the hints could not be utilized?
> * What is behavior of shmem_realloc on a pointer return by shmem_malloc_with_hints?  Are the hints preserved?  This should be specified (there is precedent with realloc on aligned allocations).
> * Should we have calloc and realloc with hints operations?

1) Only valid options to provide are the defined constants and I capture that in the text. It is similar to OpenSHMEM context options. I’m I missing something? Let us discuss on the call.

2) This usage I read semantically as constraints. I’m not sure if the Users should rely on any behavior like that. There are performance hints, really. The implementation provides no guarantee but does the best effort. 
Also, let's use the interface and gather experience, if Users think there is a need for such information. We can design an API for it. For example, a local query interface that provides information about memory’s attributes such as access model, size, alignment, and memory kind would be more useful, IMO. 

3) IIRC, we discussed this at F2F and we decided to do nothing on this .i.e., no hints need to be inherited. The library already has information about how the User intends to use the memory. If the library implementation can optimize it, it will.  

4) IMHO, it is not same alignment because alignment is more of a constraint than a hint. 

5) Yes. As far as fixing the shmem_malloc - I think we will be over-specifying if we specify the conditions when the memory allocation fails. It is probably outside the scope of this ticket anyways. 

6) For me it is not clear why would we need it? Is there a specific use case that would drive this? Also, I don’t want to add to this ticket. If I add, then I have to justify the need :)

I did not see them until this morning and I wanted to answer them before the vote :). So, the answers are a bit brief but hopefully, it makes the point. Also, if need be, we can discuss it during the vote.

jdinan:
Suggest that we rename `SHMEM_HINT_ATOMICS_REMOTE` to `SHMEM_MALLOC_ATOMICS_REMOTE` to maintain the convention that we namespace constants according to which family of API routines they are used with.  We follow this convention with `SHMEM_CTX_*` and `SHMEM_TEAM_NUM_CONTEXTS` constants.

jdinan:
@manjugv Please let me know when the PR is ready to merge.  We may need to use section committee review to approve edits prior to the merge.

manjugv:
@jdinan  The PR is ready to merge.

The two changes (1 and 2) we discussed today, I will bring it as separate PRs. So, it is easy to process. Both (1) and (2) we can do it as a chapter committee change. If the committee prefers to vote on (2) that is fine too.

(1)  Change SHMEM_HINT_SIGNAL to SHMEM_HINT_SIGNAL_REMOTE
(2) Text about shmem realloc behavior with hints

jdinan:
@manjugv Thanks -- will proceed with the merge.

Please bring into the discussion of (1) renaming these constants `SHMEM_MALLOC_*` to match the constant namespacing convention that we have established with contexts and teams constants.

I reviewed the existing text for (2) and I think I misremembered having text indicating how realloc deals with alignment.  We may need to deal with the alignment part as a separate ticket.  Something to keep in mind as you draft the text for how realloc deals with memory allocated by `shmem_malloc_with_hints`.

PR #269, opened 2019-03-27
Fix broken reference

jdinan:
We broke the following reference in the OpenSHMEM 1.1 changelog with #215:
```
LaTeX Warning: Reference `sec:mpi' on page 126 undefined on input line 836.
```


jdinan:
Need a couple +1s to merge this DocEdit.

tonycurtis:
xref to `\label{sec:openshmem_history}` instead?

jdinan:
That doesn't seem quite right, since the text refers to the interoperability annex.  I was also thinking that referencing a version of the specification, like "See OpenSHMEM 1.1, Annex D" might be a good way to handle links that are no longer valid.  But, I hesitated since that annex is there

tonycurtis:
Ah, did the annex letters change?  OK, I'll give a +1

manjugv:
👍

PR #280, opened 2019-08-28
Teams Proposal

gmegan:
This is the teams proposal that was sent out for reading on 8-8-2019 and read on 8-22-2019

PDF:
[main_spec.pdf](https://github.com/openshmem-org/specification/files/3553019/main_spec.pdf)


manjugv:
**Notes from Teams Proposal Reading (Aug 22nd)**

Page 29  
* Remove two sentences on the top of the page

Page 34
* All config_mask params should be same or not ? A note to implementers would be useful.

Page 34 
* Move line 8-9 to the intro section
* Clean up the line 16-18

Page 39
* Line 10 - shorten to remove default team

10.5.2
* Add a sentence about shmem_team_create_ctx is a local operation

Page 78
*  Remove line 5-6

10.9.3
* Line 17 - change “suspends PE” description to support the thread model. It should describe in terms of threads rather than just PE.

10.9.4
*  change “suspends PE” description to support the thread model. It should describe in terms of threads rather than just PE.

Page 88
* Line 3-4 and 7-8 - Delete fortran references

Table 8
* Identify all new operations that were added in this ticket and make it a separate ticket 

Page 94 
* Remove C/C++ interfaces in the example

Backmatter 
- Add change log 
- Add function interfaces to Annex E

RaymondMichael:
shmem_team_config_t only has one field in it. Do we anticipate wanting additional settings in the future? We won't be able to modify the struct due to backwards compatibility. If we can't modify the struct, then what is the use of the mask to shmem_team_split_()?

jamesaross:
I missed the first half of the reading, so maybe this was discussed.

The routine `void shmem_sync(int PE_start, int logPE_stride, int PE_size, long *pSync)` is marked for deprecation, but the same name is being used for C11 team synchronization: `int shmem_sync(shmem_team_t *team)`.  But then there is a new name for the C/C++ routine: `int shmem_team_sync(shmem_team_t *team)`.  Is this intentional?  I feel like we shouldn't be changing prototypes to a routine that hasn't been removed, just deprecated.  The C11 API should be the same for C/C++: `int shmem_team_sync(shmem_team_t *team)`.

gmegan:
> shmem_team_config_t only has one field in it. Do we anticipate wanting additional settings in the future? We won't be able to modify the struct due to backwards compatibility. If we can't modify the struct, then what is the use of the mask to shmem_team_split_()?

**Why only 1 field?**
When the struct was originally added, there were more fields for features that were considered too much to add in the first version of the teams API. The struct is expected to be expanded in the future to include the ability to request additional node resources or to disable them. For example, it may be that teams need additional resources to support non-blocking collectives, or team based heaps, which could be configured on a per-team basis.

**Can we add more fields?**
Additionally, we have left this struct open for implementation dependent fields to be added for experimentation or special purpose team configuration. We generally expect these fields to relate to some type of resource requests made at team creation.

As for backward compatibility, the mask fields are meant to allow the struct to change and add new fields as long as the old fields still exist.

**We defined a specific C struct... Oops maybe?**
You raise an interesting point though. We have specified the struct itself, when what we really meant to do was specify that there is a type ``shmem_config_t``. This type is a data structure implemented with a C struct (or potentially C++ class ?) with some number of publicly accessible member elements. We promise that one of these elements will be ``num_contexts`` of type ``int`` but we make no promise about where this element appears in the data structure or what else the data structure contains.

**Default values for configuration**
We also intend to specify that each member of the ``shmem_config_t`` data structure has a default value for new teams. This default value is not necessarily the configuration value for any predefined team. E.g., we don't tell implementations how to configure ``SHMEM_TEAM_WORLD`` since this team has a unique, predefined functionality in the program.

**Why the mask values?**
When a team is newly created, the configuration structure is optional, and only needed to override the default values. The program has to explicitly state which fields are being used by way of the mask values. This way if new fields are added to the data structure later, programs will not break.

```
shmem_config_t conf;
conf.num_context = 1;
shmem_split_strided(SHMEM_TEAM_WORLD, 0, 2, 10, &conf, SHMEM_TEAM_NUM_CONTEXTS, &new_team);
```

This code will work without modification even if next year we add a field ``heap_size``, which might end up with some garbage value in this code sample depending on how the compiler initializes conf. So even if we end up with ``conf.heap_size = 2094230948``, the implementation knows to ignore this value since the program only asked us to look at ``num_contexts``.

gmegan:
> I missed the first half of the reading, so maybe this was discussed.
> 
> The routine `void shmem_sync(int PE_start, int logPE_stride, int PE_size, long *pSync)` is marked for deprecation, but the same name is being used for C11 team synchronization: `int shmem_sync(shmem_team_t *team)`. But then there is a new name for the C/C++ routine: `int shmem_team_sync(shmem_team_t *team)`. Is this intentional? I feel like we shouldn't be changing prototypes to a routine that hasn't been removed, just deprecated. The C11 API should be the same for C/C++: `int shmem_team_sync(shmem_team_t *team)`.

What we expect from implementations here is that C11 programs just use shmem_sync for either active sets or teams. ``shmem_sync`` can use C11 generics to use the argument list to resolve into the appropriate API call. We provide the specific ``shmem_team_sync`` C/C++ interface to aid in debugging and profiling.

So this code is correct for C11:
```
shmem_sync(SHMEM_TEAM_WORLD);
shmem_sync(0, 1, shmem_n_pes(), pSync);
```

I would have to dig out the generic C11 code that I wrote to convince myself it is possible that this C11 code will resolve into:

```
shmem_team_sync(SHMEM_TEAM_WORLD);
shmem_sync(0, 1, shmem_n_pes(), pSync);
```

This way a profiler can be attached to look for the `shmem_sync` symbol only for active sets and the `shmem_team_sync` symbol only for teams.

Happy to go find the generic code if this would help. Let me know.

jamesaross:
I'm not sure offhand how to do it with a C11 _Generic macro.  Maybe you could dig that out.  But you can also do it with the C preprocessor.  At least with the GNU toolchain (see below).  The code below could actually support C89, C99, C11, or C++.  My point wasn't about whether it _could_, but whether the standard _should_.  Wouldn't this symbol name conflict make it more difficult for people building the tools?

I think I would prefer dropping the C11 interface (`shmem_sync(shmem_team_t *team)`) and just keeping the common C/C++ interface (`int shmem_team_sync(shmem_team_t *team)`).  Maybe a tools person can pipe up and tell me it's fine and the new C11 interface with the old name won't create headaches.

```c
/* compile as C89, C99, C11, or C++ */

#include <stdio.h>

#define __SHMEM_MACRO(_1,_2,_3,_4,NAME,...) NAME
#define shmem_sync(...) __SHMEM_MACRO(__VA_ARGS__, shmem_syncA, (void*)0, (void*)0, shmem_syncB)(__VA_ARGS__)

/*
 * Two prototypes with the same name supported:
 *    void shmem_sync(int PE_start, int logPE_stride, int PE_size, long *pSync);
 *    int shmem_sync(shmem_team_t *team);
 */

void shmem_syncA(int PE_start, int logPE_stride, int PE_size, long *pSync)
{
        printf("called: void shmem_sync(int PE_start, int logPE_stride, int PE_size, long *pSync)\n");
        return;
}

typedef struct {
        int whatever;
} shmem_team_t;

int shmem_syncB(shmem_team_t *team)
{
        printf("called: int shmem_sync(shmem_team_t *team)\n");
        return 0;
}


int main(void)
{
        shmem_team_t foo;

        shmem_sync(0, 0, 0, NULL);
        shmem_sync(&foo);
        shmem_sync(0);

        return 0;
}

```

RaymondMichael:
The problem is shmem_team_get_config(). If a program were compiled against an OpenSHMEM 1.5 header, then run with a 1.6 library in which we've added fields to the struct, then the size of struct will be different and the result will be memory corruption. A better solution is something like the MPI_Info API which uses a completely opaque object and modifies it through setters and getters.
 
> **We defined a specific C struct... Oops maybe?**
> You raise an interesting point though. We have specified the struct itself, when maybe what we really meant to do was specify that there is a type `shmem_config_t`. This type is a data structure implemented with a C struct (or potentially C++ class ?) with some number of publicly accessible member elements. We promise that one of these elements will be `num_contexts` of type `int` but we make no promise about where this element appears in the data structure or what else the data structure contains.

naveen-rn:
> If a program were compiled against an OpenSHMEM 1.5 header, then run with a 1.6 library in which we've added fields to the struct

If I understand correctly, this is a problem when the libraries are dynamically linked. AFAIU, OpenSHMEM is not dynamically backward compatible. This is not the first routine that creates this issue. shmem_get_version_info() is another similar one.

RaymondMichael:
shmem_info_get_version() returns a pair of integers. In 1.4 the definition of shmem_ctx_t with the accessible flags element could be a problem, but we don't have any APIs returning a shmem_ctx_t. There are no other routines / objects that might have this problem.

From my perspective dynamic compatibility is the default position. If vendors have to recompile every year or two, that's going to cause a problem. We have plenty of examples of customers expecting an application to work for 10+ years.

gmegan:
> I'm not sure offhand how to do it with a C11 _Generic macro. Maybe you could dig that out. But you can also do it with the C preprocessor. At least with the GNU toolchain (see below). The code below could actually support C89, C99, C11, or C++. My point wasn't about whether it _could_, but whether the standard _should_. Wouldn't this symbol name conflict make it more difficult for people building the tools?
> 
> I think I would prefer dropping the C11 interface (`shmem_sync(shmem_team_t *team)`) and just keeping the common C/C++ interface (`int shmem_team_sync(shmem_team_t *team)`). Maybe a tools person can pipe up and tell me it's fine and the new C11 interface with the old name won't create headaches.

As I remember, the rationale for doing this was that if the ultimate goal is to drop active set support and just have teams, then it would be nicer to avoid the *_team_* names where possible. We have other interfaces where we avoided _team, like using `shmem_and_reduce` for teams rather than `shmem_and_team_to_all`

We were able to avoid overloading any other functions because we were able to find different names... e.g. we deprecate `shmem_broadcast32` for active set and add `shmem_broadcast` for teams. But `shmem_sync` was the one function we didn't have a different name for, so we overloaded.

We got some informal feedback from implementors on the call at the time this was discussed, and the feedback at that time was that having distinct defined C++ symbols was very important so we have to have `shmem_team_sync` in any case. The overloading solution seemed fine to folks at the time but I don't remember any in depth discussion of potential problems with that. I would also like more feedback on whether the overloading of functions is a problem, since this might come up again in future.

shamisp:
> The problem is shmem_team_get_config(). If a program were compiled against an OpenSHMEM 1.5 header, then run with a 1.6 library in which we've added fields to the struct, then the size of struct will be different and the result will be memory corruption. A better solution is something like the MPI_Info API which uses a completely opaque object and modifies it through setters and getters.
> 
> > **We defined a specific C struct... Oops maybe?**
> > You raise an interesting point though. We have specified the struct itself, when maybe what we really meant to do was specify that there is a type `shmem_config_t`. This type is a data structure implemented with a C struct (or potentially C++ class ?) with some number of publicly accessible member elements. We promise that one of these elements will be `num_contexts` of type `int` but we make no promise about where this element appears in the data structure or what else the data structure contains.

@RaymondMichael good point. If we decide to address this problem, I think the solution is relatively trivial. We can add to `shmem_team_get_config()` additional input parameter called `field_mask` that will let user specify which fields of config it wants to "get". Since the "old" code does not have new indexes for `field_mask` it has no way to corrupt the structure. I'm making an assumption here that we never remove old fields from the config structure and all the new fields are appended to the end of the structure.

jdinan:
@gmegan Could you please remove `\oldtext{}`, `\newtext{}`, `\color{Green}`, and any other change highlighting?

jdinan:
Superseded by #297

PR #282, opened 2019-09-09
Teams proposal fixes

gmegan:
This PR includes non-semantic fixes and clarifications to the teams proposal in PR #280

The diff of the Teams proposal and fixes is here:
https://github.com/gmegan/specification/compare/teams-official-20190808...gmegan:teams-fixes-20190909


content/shmem_collect.tex
```diff
@@ -38,43 +66,88 @@
...
+The  \dest{}  and \source{} data  objects must conform to certain typing
+constraints, which are as follows:
+}{Routine}{Data type of \VAR{dest} and \VAR{source}}
+\apitablerow{\FUNC{shmem\_collectmem}, \FUNC{shmem\_fcollectmem}}{\Cstd: Any data  type. \VAR{nelems} is scaled in bytes.}%
+\apitablerow{\FUNC{shmem\_collect64}, \FUNC{shmem\_fcollect64}}%
+    {Any noncharacter type that has an element size of \CONST{64} bits. No \Fortran derived types nor
+    \CorCpp{} structures are allowed.}
+\apitablerow{\FUNC{shmem\_collect32}, \FUNC{shmem\_fcollect32}}%
+    {Any noncharacter type that has an element size of \CONST{32} bits. No \Fortran derived types nor
+    \CorCpp{} structures are allowed.}
```
jdinan: This API table was deleted as part of the Fortran removal.

content/shmem_sync.tex
```diff
@@ -1,16 +1,39 @@
 \apisummary{
+    \newtext{Registers the arrival of a \ac{PE} at a synchronization point and suspends
+    execution until all other \acp{PE} in a given \openshmem team or active set
```
jdinan:
```suggestion
    execution of the calling thread until all other \acp{PE} in a given \openshmem team or active set
```

content/shmem_sync.tex
```diff
@@ -1,16 +1,39 @@
...
+    arrive at a synchronization point. For multithreaded programs, execution is suspended
+    as specified by the threading model (Section \ref{subsec:thread_support}).}
```
jdinan: Propsed change:
"Registers the arrival of a PE at a synchronization point.  This routine does not return until all other PEs in a given OpenSHMEM team or active set arrive at a synchronization point."
The "For multithreaded programs, ..." sentence is removed.

content/shmem_team_create_ctx.tex
```diff
@@ -0,0 +1,63 @@
+\apisummary{
+  Create a communication context from a team locally.
```
jdinan:
```suggestion
  Create a communication context from a team.
```

content/shmem_team_create_ctx.tex
```diff
@@ -0,0 +1,63 @@
...
+    The set of options requested for the given context.
+    Multiple options may be requested by combining them with a bitwise OR
+    operation; otherwise, \CONST{0} can be given if no options are requested.}
+  \apiargument{OUT}{ctx}{A handle to the newly created context.}
+\end{apiarguments}
+
+\apidescription{
+  The \FUNC{shmem\_team\_create\_ctx} routine creates a new communication
+  context and returns its handle through the \VAR{ctx} argument.
+  This context is created from the team specified by the \VAR{team} argument.
```
jdinan:
```suggestion
  This context is created from the team specified by the \VAR{team} argument; however, it is not collective.
```

example_code/shmem_team_context.c
```diff
@@ -0,0 +1,112 @@
...
+  // We will add up some results on pe 4 of team_3s using ctx_2s
+  if ((team_3s != SHMEM_TEAM_INVALID) && (team_2s != SHMEM_TEAM_INVALID)) {
+    int pe4_of_3s_in_2s = my_ctx_translate_pe(ctx_3s, 4, ctx_2s);
+
+    if (pe4_of_3s_in_2s < 0) {
+      fprintf (stderr, "Fail to translate pe 4 from 3s context to 2s context\n");
+    }
+    else {
+      // Add up the results on pe 4 of the 3s team, using the 2s team context
+      shmem_ctx_int_atomic_add(ctx_2s, &isum, ival2 + ival3, _pe4_of_3s_in_2s);
```
nspark:
```suggestion
      shmem_ctx_int_atomic_add(ctx_2s, &isum, ival2 + ival3, pe4_of_3s_in_2s);
```

jdinan:
FIXME: `shmem_sync_example.c` can index outside of the twos and threes teams in the call to `shmem_p` when `npes` is not a multiple of 2 and 3.  Would be good to demonstrate the use of the translate function in these examples instead.

jdinan:
FIXME: The undefined behavior annex covers using an invalid PE id.  Add a row to cover the case where the PE id is valid, but not a member of the given team (e.g. in and RMA operations, as the root in a broadcast, etc.).

jdinan:
Follow-up: Do we want ---

(a) for the relative order of PEs in the child team to be the same as the relative order of those PEs in the parent team (current text), or

(b) for the relative order of PEs in the child team to be implementation defined, but the spec will strongly suggest (via advice to implementors) that using the same ordering as the parent team leads to more scalable, lower overhead implementations.

jdinan:
Merged into #280

jdinan:
Edits moved to #298

PR #283, opened 2019-09-09
Teams Proposal Remove Error Handling

gmegan:
This removes the separate error handling section from the teams proposal in PR #280 

Two paragraphs from the removed section are added back to the collectives introduction with one extra statement added there for clarification.

The diff of the teams proposal as read and this PR is here:
https://github.com/gmegan/specification/compare/teams-official-20190808...gmegan:teams-remove-errorhandle


content/collective_intro.tex
```diff
@@ -1,49 +1,137 @@
...
+
+Collective operations involving many \acp{PE} may return values indicating success while other \acp{PE} are still executing the collective operation. Return values indicating success of a collective routine on one \ac{PE} do not indicate that all \acp{PE} involved in the collective operation will return from the routine successfully. In the case where successful local completion of a collective implies the success of some global operation, such as team creation resulting in a valid team on all involved \acp{PE}, the implementation should not return \CONST{0} from the routine if the implied or stated global guarantees of the routine are not met.
```
jdinan:
```suggestion
Collective operations involving many \acp{PE} may return values indicating success while other \acp{PE} are still executing the collective operation. Return values indicating success or failure of a collective routine on one \ac{PE} may not indicate that all \acp{PE} involved in the collective operation will return the same value. Some operations, such as team creation, must return consistent return codes across multiple \acp{PE}.
```

jdinan: Not sure that's perfect, but maybe a bit closer to what we were going for in today's discussion.

jdinan: Also, hard wrap the lines, please.

content/collective_intro.tex
```diff
@@ -1,49 +1,137 @@
...
+call to \FUNC{shmem\_team\_sync} on the default team.
+
+The \FUNC{shmem\_barrier\_all} routine synchronizes all \acp{PE} in
+the default team and ensures completion of all local and remote memory
+updates issued via the default context.  This routine is equivalent to
+a call to \FUNC{shmem\_ctx\_quiet} on the default context followed by a
+call to \FUNC{shmem\_team\_sync} on the default team.
+
+\subsubsection*{Error codes returned from collectives}
+\CorCpp routines that return an integer error code follow the convention that \CONST{0} indicates successful local completion of the operation. This is considered a best effort of the implementation to indicate that all required local operations have been performed correctly inside the routine and the internal \openshmem state on the calling \ac{PE} is consistent with the description of the routine and its arguments upon completion. Implementations may use an integer return value from a routine to define integer error codes specific to the implementation as long as those codes are not already explicitly defined for that routine by the \openshmem specification.
```
jdinan:
```suggestion
\CorCpp routines that return an integer error code follow the convention that \CONST{0} indicates successful local completion of the operation. Local completion indicates that the calling \ac{PE} has completed its involvement in the routine and the internal \openshmem state on the calling \ac{PE} is consistent with the description of the routine and its arguments. Implementations may use an integer return value from a routine to define integer error codes specific to the implementation as long as those codes are not already explicitly defined for that routine by the \openshmem specification.
```

jdinan:
Merged into #280

PR #290, opened 2019-09-25
Fix example intro affected by Fortran removal

jdinan:


jdinan:
I'd like to handle this as a document edit.  Can I get sign-offs from two committee members (+1 this comment or add a comment below)?  Thanks!

jdinan:
Approved at F2F meeting.

Issue #253, opened 2018-10-11
What does "same ... and ... not overlapping" mean?


jdinan:
In reductions, OpenSHMEM 1.4 states:

"The source and dest arrays may be the same array, but they may not be overlapping arrays."

Does this mean that:
1. The source and dest pointers must either be identical or correspond to different objects, or
1. The source and dest pointers may point to elements in the same array allocation, but they must not overlap in memory, or
1. Something else?

spophale:
The sentence introduced in 1.4 suddenly moves away from symmetric objects and wants to be able to work with an object in fractions. It is okay if that is the route we want to go. But it needs to be consistent across the spec. 

When I read that in the spec I mentally corrected it to "The source and dest arrays may be the same array, but they may not be overlapping **sub-**arrays."  From reading of that sentence alone I would go with option 2.

nspark:
Thinking about #266 also made me realize that we don't say anything about overlapping RMA operations. For example, which of the following RMA operations is well-defined...?

```c
const size_t nelems = ...;
int *sym_addr = shmem_calloc(nelems, sizeof(*sym_addr));
shmem_put(sym_addr    , sym_addr, nelems    , shmem_my_pe()); // Case 0
shmem_put(sym_addr + 2, sym_addr, nelems - 2, shmem_my_pe()); // Case 1
```

Now, throw `shmem_ptr` into the mix and...?
```c
const int nbr = (shmem_my_pe() + 1) % shmem_n_pes();
const size_t nelems = ...;
int *sym_addr = shmem_calloc(nelems, sizeof(*sym_addr));
int *nbr_addr = shmem_ptr(sym_addr, nbr);
if (nbr_addr)
  shmem_put(sym_addr, nbr_addr, nelems, nbr); // Case 3
```

nspark:
It's worth noting that the statement for reductions that:

> `source` and `target` may be the same array, but they must not be overlapping arrays

goes all the way back to OpenSHMEM 1.0.

davidozog:
> Thinking about #266 also made me realize that we don't say anything about overlapping RMA operations. For example, which of the following RMA operations is well-defined...?
> 
> ```c
> const size_t nelems = ...;
> int *sym_addr = shmem_calloc(nelems, sizeof(*sym_addr));
> shmem_put(sym_addr    , sym_addr, nelems    , shmem_my_pe()); // Case 0
[...]

PR #276, opened 2019-05-06
Update Symmetric Pointer Text

jdinan:
Clarify that OpenSHMEM routine arguments contain either symmetric or local addresses, where each is used, and restrictions on pointer arguments.

PR #291, opened 2019-09-25
Clarify reduction src/dest overlap text

jdinan:
The current text has been around since at least as far back as the T3E documentation, but it's hard (at least for me) to grok.  This PR updates the description of src/dest overlap using the language from the original SHMEM collectives document, ["SHMEM User’s Guide for C" by Barriuso and Knies '94](https://www.cs.cmu.edu/afs/cs/project/cmcl/link.iwarp/OldFiles/archive/fx-papers/cri-shmem-users-guide.ps).

### DIFF
--- a/content/shmem_reductions.tex
+++ b/content/shmem_reductions.tex
@@ -244,10 +244,11 @@ where \TYPE{} is one of the integer, real, or complex types supported for the PR
     perform.  The \source{} array on all \acp{PE} participating in the reduction
     provides one element for each reduction.  The results of the reductions are placed in the
     \dest{} array on all \acp{PE} participating in the reduction.
-    
-    The \source{} and \dest{} arrays must not point to partially overlapping
-    memory regions.  That is, they must point to the same region in memory or
-    to completely disjoint regions in memory.
+
+    The \source{} and \dest{} arguments must either be the same symmetric
+    address, or two different symmetric addresses corresponding to buffers that
+    do not overlap in memory. That is, they must be completely overlapping or
+    completely disjoint.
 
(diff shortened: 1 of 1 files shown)


jdinan:
Closes #253

content/shmem_reductions.tex
```diff
@@ -125,8 +125,9 @@
     \dest{} array on all \acp{PE} in the active set.  The active set is defined
     by the \VAR{PE\_start}, \VAR{logPE\_stride}, \VAR{PE\_size} triplet.
 
-    The \source{} and \dest{} arrays may be the same array, but they may not be
-    overlapping arrays.
+    The \source{} and \dest{} arrays must not point to partially overlapping
```
agrippa: I feel like "region" isn't a precisely defined term, and it isn't very commonly used in the spec (looks like in 1.4 it's only used in shmem_calloc and in the locking APIs).

After reading the changes you've made to the symmetric pointer text, I'm wondering if it would be better to have this change build on that and say something like:

The \source{} and \dest{} arguments must either be passed the same symmetric address, or two different symmetric addresses at least \nreduce{} elements apart. That is, they must be completely overlapping or completely disjoint.

jdinan: I like this change.  The `nreduce` bit doesn't completely address the case where a pointer queried via `shmem_ptr` aliases the other pointer.  We could borrow the overlapping buffer text from #276.  What do you think?
> The \source{} and \dest{} arguments must either be the same symmetric address, or two different symmetric addresses corresponding to buffers that do not overlap in memory. That is, they must be completely overlapping or completely disjoint.

jdinan:
Special ballot: 6f9d8f4

Issue #243, opened 2018-09-04
Interoperability with other programming models 


minsii:
**Goal**: define a new interoperability section

**Context**: Appendix D has been removed because the contents are outdated. We want to make a new interoperability section that defines the interoperability of OpenSHMEM with other programming models such as MPI, UPC, CAF, in order to help portable hybrid OpenSHMEM + X programs.

The section is planed to cover: 
- Standardization of features / useful APIs for portable hybrid program
- Clarification of undefined behavior caused by mixed use of different models
- Advice to users about the ways to guarantee program portability  
- Advice to runtime developers about the ways to optimize performance

A PR is created for MPI interoperability. See https://github.com/minsii/openshmem-specification/pull/1

minsii:
Slides used in 20180904's Thread WG call:
[SHMEM_Interoperability_20180904.pptx](https://github.com/openshmem-org/specification/files/2350079/SHMEM_Interoperability_20180904.pptx)

spophale:
Notes form WG Meeting Tue Sep 4th:

- Atomicity guarantees between MPI and SHMEM operations will not be part of the Spec - user responsibility.
- Removed section on data allocation by one programming model and used by the other. 
- Open Question : How to interpret SHMEM_PROGRESS_UNIFIED and SHMEM_PROGRESS_SEPARATE

bcernohous:
Try to explain myself after a bit more thought.

>Open Question : How to interpret SHMEM_PROGRESS_UNIFIED and SHMEM_PROGRESS_SEPARATE

I think SHMEM queries SHMEM attributes, not MPI/UPC/Fortran.  

I think there are 3 attributes (+detail/sub attributes if desired).

1) SHMEM make progress on xxx :

- SHMEM_PROGRESS_ALL

    or individually
    - SHMEM_PROGRESS_MPI
    - SHMEM_PROGRESS_UPC
    .. to whatever level of detail

2) SHMEM needs other programming models, or the user applications, to make progress:

- SHMEM_REQUIRES_PROGRESS (or SHMEM_PROGRESS_MANUAL)

3) SHMEM always makes progress (no dependency on other programming models or application)

- SHMEM_PROGRESS_AUTO

For 2), I don't think we can define if MPI progress (MPI_TEST) make progress on SHMEM, only that SHMEM needs progress.

e.g. 

Cray XC supports Cray MPI, Intel MPI, OpenMPI (and sometime MVAPICH).  

[...]

PR #304, opened 2019-10-24
Rewriting annex - interoperability with other programming models

minsii:
This PR rewrites `Annex D - Interoperability with other Programming Models`. It clarifies MPI Interoperability and defines an OpenSHMEM extended API for querying interoperability features. The interoperability with other communication libraries and languages will be addressed in future work.

Related discussion can be found at [ticket-243](https://github.com/openshmem-org/specification/issues/243) and [pr-1](https://github.com/minsii/openshmem-specification/pull/1)


minsii:
Diff from September F2F meeting: [843d8bb..d560514](https://github.com/minsii/openshmem-specification/compare/843d8bb..d560514)

Summary of changes:
- 1.2 Dynamic Process Creation: delete "MPMD" from the section title
- 1.3 Thread Safety:
  - For the additional thread safety rules, mention them as suggestions to ensure portable hybrid program because they are required only for implementations that provide both OpenSHMEM and MPI.
  - Improve the description of additional rule for `THREAD_SERIALIZED` safety.
- 1.6 Communication Progress:
   - More specific guidance to implementors: use software progress thread if the implementation supports asynchronous communication for both MPI and OpenSHMEM.
   - Add example in `2.1 SHMEM_QUERY_INTEROPERABILITY` to showcase how the progress support can help program to avoid deadlock.
- Others:
  - Example cleanup (header file, unneeded function)
  - Grammar fixes

Performance impact when implementation supports progress for MPI:
- It should only impact on implementations that supports both MPI and OpenSHMEM
- For implementations that already uses a software thread for OpenSHMEM progress, the performance impact should be very small.
- For implementations that provide full hardware progress for OpenSHMEM but has to create a software thread to support MPI progress, performance may be degraded. However, the implementation can optimize it. For example, it can reduce the frequency of thread polling to minimize thread contention&oversubscription overhead. Or it can use other approaches for MPI progress such as Casper: https://github.com/pmodels/casper.
- For implementations that has to involve extra heavy overhead to support MPI progress, it can choose to disable the feature.

Nevertheless, I feel that the optimizations of asynchronous progress is irrelevant to this proposal. Thus, I do not discuss them in detail in the proposal.

minsii:
Diff from September F2F meeting: [843d8bb..d560514](https://github.com/minsii/openshmem-specification/compare/843d8bb..d560514)

Summary of changes:
- 1.2 Dynamic Process Creation: delete "MPMD" from the section title
- 1.3 Thread Safety:
  - For the additional thread safety rules, mention them as suggestions to ensure portable hybrid program because they are required only for implementations that provide both OpenSHMEM and MPI.
  - Improve the description of additional rule for `THREAD_SERIALIZED` safety.
- 1.6 Communication Progress:
   - More specific guidance to implementors: use software progress thread if the implementation supports asynchronous communication for both MPI and OpenSHMEM.
   - Add example in `2.1 SHMEM_QUERY_INTEROPERABILITY` to showcase how the progress support can help program to avoid deadlock.
- Others:
  - Example cleanup (header file, unneeded function)
  - Grammar fixes

Performance impact when implementation supports progress for MPI:
- It should only impact on implementations that supports both MPI and OpenSHMEM
- For implementations that already uses a software thread for OpenSHMEM progress, the performance impact should be very small.
- For implementations that provide full hardware progress for OpenSHMEM but has to create a software thread to support MPI progress, performance may be degraded. However, the implementation can optimize it. For example, it can reduce the frequency of thread polling to minimize thread contention&oversubscription overhead. Or it can use other approaches for MPI progress such as Casper: https://github.com/pmodels/casper.
- For implementations that has to involve extra heavy overhead to support MPI progress, it can choose to disable the feature.

Nevertheless, I feel that the optimizations of asynchronous progress is irrelevant to this proposal. Thus, I do not discuss them in detail in the proposal.

minsii:
@jdinan @nspark I just moved the PR to again the main spec. Could you please check if all your comments are addressed?

minsii:
@manjugv For the performance concern about private context, I do not think adding support for MPI progress (e.g., using a software thread) can degrade the performance of SHMEM operations that use a private context. The additional thread should only touch the "context" used for MPI and does not access the resource isolated for any private context. What do you think?

content/interoperability.tex
```diff
@@ -0,0 +1,191 @@
...
+and provides a mechanism to establish communication
+between the newly created processes and the existing MPI application (see
+MPI standard version 3.1, Chapter 10).
+Unlike MPI, \openshmem starts all processes at once and requires all PEs to
+collectively allocate and initialize resources (e.g., symmetric heap) used by
+the \openshmem library before any other \openshmem routine may
+be called. Communicating with a dynamically created process in the \openshmem
+environment may result in undefined behavior.
+Hence, users should not use \openshmem and MPI dynamic process models
+in the same program.
```
jdinan: I'm not sure this is the right conclusion.  Connect/accept can be used to connect a job to a service, e.g. connect a SHMEM job to a database or I/O service.  Or to connect two different SHMEM jobs as a workflow.  These should all work fine, provided the communication is done with MPI and not SHMEM.  A better conclusion may be:
```
\openshmem does not support communication with dynamically created or connected processes. 
In such a scenario, MPI must be used communicate with these processes.
```

content/interoperability.tex
```diff
@@ -0,0 +1,191 @@
...
+
+
+\subsection{Thread Safety}
+\label{subsec:interoperability:thread}
+Both \openshmem and MPI define the interaction with user threads in a program
+with routines that can be used for initializing and querying the thread
+environment. In a hybrid program, the user may request different thread levels
+at the initialization calls of \openshmem and MPI environments; however, the
+returned support level provided by the \openshmem library might be different
+from that returned in an \openshmem-only program. For instance, the former
```
jdinan: Suggest:
```
returned support level provided by the \openshmem or MPI library might be different
from that returned in an non-hybrid program. For instance, the first
```

content/interoperability.tex
```diff
@@ -0,0 +1,191 @@
...
+initialization call in a hybrid program may initialize a resource with the
+user-requested thread level, but the supported level cannot be updated by the latter
```
jdinan:
```suggestion
user-requested thread level, but the supported level cannot be updated by a subsequent
```

content/interoperability.tex
```diff
@@ -0,0 +1,191 @@
...
+\end{itemize}
+
+\subsection{Mapping Process Identification Numbers}
+\label{subsec:interoperability:id}
+
+Similar to the PE identifier in \openshmem, MPI defines rank as the
+identification number of a process in a communicator. Both the \openshmem PE
+and the MPI rank are unique integers assigned from zero to one less than the total
+number of processes. In a hybrid program, the \openshmem
+PE and the MPI rank in \VAR{MPI\_COMM\_WORLD} of a process can be equal.
```
jdinan:
```suggestion
PE number in \LibHandleRef{SHMEM\_TEAM\_WORLD} and the MPI rank in \VAR{MPI\_COMM\_WORLD} of a process can be equal.
```

content/interoperability.tex
```diff
@@ -0,0 +1,191 @@
...
+This feature, however, may be provided by only some of the \openshmem and MPI
+implementations (e.g., if both environments share the same underlying process
+manager) and is not portably guaranteed. A portable program should always
+use the standard functions in each model, namely, \FUNC{shmem\_my\_pe} in \openshmem
```
jdinan:
```suggestion
use the standard functions in each model, namely, \FUNC{shmem\_team\_my\_pe} in \openshmem
```

content/interoperability.tex
```diff
@@ -0,0 +1,191 @@
...
+\openshmem and MPI each define similar one-sided communication models;
+however, a portable program should not assume interoperability between these
+models.
+For instance, \openshmem guarantees the atomicity only of concurrent \openshmem AMO operations
+that operate on symmetric data with the same datatype. Access to the same symmetric
+object with MPI atomic operations, such as an \FUNC{MPI\_Fetch\_and\_op}, may
+result in an undefined result. Furthermore,
+because most RMA programs can be written by using either \openshmem or MPI RMA,
+users should choose only one of the RMA models in the same program, whenever
+possible, for performance and code simplicity.
```
jdinan: If a user is using a library that uses MPI RMA, they may not have an either-or choice.  Phase-based access (e.g. MPI I/O and SHMEM RMA phases) using the different models should still be fine.  Suggest instead:
```
Users should avoid situations where MPI and \openshmem operations perform concurrent
accesses the same memory locations.
```

content/interoperability.tex
```diff
@@ -0,0 +1,191 @@
...
+The program should always check the \VAR{provided} thread level returned
+at the corresponding initialization call or query the level of thread support
+after initialization to portably ensure thread support in each communication
+environment.
+
+Both \openshmem and MPI define similar thread levels, namely, \VAR{THREAD\_SINGLE},
+\VAR{THREAD\_FUNNELED}, \VAR{THREAD\_SERIALIZED}, and \VAR{THREAD\_MULTIPLE}.
+When requesting threading support in a hybrid program, however,
+the following additional rules are applied if the implementations of \openshmem
+and MPI share the same internal communication resource.
```
jdinan: Is it possible to query whether synchronization requirements are combined for both MPI and SHMEM or separate (i.e. separate serialization domains)?

content/shmem_query_interoperability.tex
```diff
@@ -0,0 +1,47 @@
...
+
+\begin{Csynopsis}
+int @\FuncDecl{shmem\_query\_interoperability}@(int property);
+\end{Csynopsis}
+
+\begin{apiarguments}
+    \apiargument{IN}{property}{The interoperability property queried by the user.}
+\end{apiarguments}
+
+% compiling error ?
```
jdinan: Is this comment resolved?

content/shmem_query_interoperability.tex
```diff
@@ -0,0 +1,47 @@
...
+% \apidescription{
```
jdinan: Do we need this?

content/shmem_query_interoperability.tex
```diff
@@ -0,0 +1,47 @@
...
+is supported by the \openshmem library. One of the following properties can be 
+queried in an \openshmem program after finishing the
+initialization call to \openshmem and that of the relevant programming models
+being used in the program. An \openshmem library implementation may extend the
+available properties.
+
+\begin{itemize}
+\item \VAR{SHMEM\_PROGRESS\_MPI} Query whether the \openshmem
+implementation makes progress for the MPI communication used in the user program.
+\end{itemize}
```
jdinan:
```
\item \VAR{SHMEM\_SERIALIZATION\_SEPARATE} Query whether the \openshmem
and MPI have separate thread serialization domains.
```

content/interoperability.tex
minsii: I like the suggested conclusion. I disallowed it because I think it is hard to precisely cover all combinations of `SHMEM+MPI dynamic process` (see below). 

- case 1:
After MPI initialization and before any PEs start OpenSHMEM initialization, the program spawns(connects to) a new MPI process. It is unclear whether P1 is included in the original `TEAM_WORLD`.
```
P0                  Spawned P1
MPI_Init(..)
spawn();
shmem_init();       shmem_init();
```

- case 2:
After MPI and SHMEM initialization, the program spawns(connects to) a new process and the new process wants to communicate with PEs in the original `TEAM_WORLD`.
```
P0                  Spawned P1
MPI_Init(..)
shmem_init();
spawn();
                    shmem_init();
```

- case 3:
After MPI and SHMEM initialization, the program spawns(connects to) some new processes (e.g., P1 and P2 in the example). Can `P1` and `P2` communicate with each other through OpenSHMEM? 
```
P0                  Spawned P1      Spawned P2
MPI_Init(..)
shmem_init();
spawn();
                    shmem_init();   shmem_init();
```

I think the new conclusion clearly defines all the above cases!

content/shmem_query_interoperability.tex
minsii: I fixed the compiling issue by adding `\apinotes{None.}` after `\apidescription{...}`. It is now consistent with other API definition.

content/interoperability.tex
minsii: Also replaced the old API with teams-API in the example code.

minsii: Also replaced the old API with teams-API in the example code.

content/interoperability.tex
minsii: Makes sense.

content/interoperability.tex
minsii: Good idea. We may also add a query property to indicate whether the same master thread is required for the `THREAD_FUNNELED` safety. But I need time to think about a use case which can benefit from the feature. Maybe it is better to first wrap up the current proposal and propose the other properties separately. What do you think?

minsii:
@jdinan The last push includes all changes. Could you please review again?

content/shmem_query_interoperability.tex
```diff
@@ -0,0 +1,50 @@
+\apisummary{
+    Determines whether an interoperability feature is supported by the \openshmem
```
naveen-rn: Is there a general prototype of this routine? Has this been tested? Sorry, I haven't been focusing on this proposal before - in general, I'm bit confused on the implementation of such a routine.

Lets consider a OpenSHMEM implementation, with the following scenario:
1. It can guarantee SHMEM_PROGRESS_MPI for an MPI implemetation MPI:1
2. It can't guarantee SHMEM_PROGRESS_MPI for an MPI implementation MPI:2
Also, both MPI:1 and MPI:2 are third party implementations and the OpenSHMEM implementation have no control over it. We can fill any known MPI and OpenSHMEM implementations in this scenario.

AFAIU, reading environment variables from the MPI implementations is incorrect and there are no standard MPI routines to determine the version info like ```shmem_info_get_name```.  So, how would the OpenSHMEM implementation determine the linked MPI implementation? Please clarify.

content/interoperability.tex
```diff
@@ -0,0 +1,189 @@
...
+\openshmem and MPI are two commonly used parallel programming models for
+distributed-memory systems. The user can choose to utilize both models in the same program
+to efficiently and easily support various communication patterns.
+
+A vendor may implement the \openshmem and MPI libraries in different ways. For
+instance, one may implement both \openshmem and MPI as standalone libraries,
+each of which allocates and initializes fully isolated communication
+resources.
+Another common approach
+is to implement both \openshmem and MPI interfaces within the
```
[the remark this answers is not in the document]
minsii: Fixed

content/interoperability.tex
```diff
@@ -0,0 +1,189 @@
...
+
+\subsection{Dynamic Process Creation}
+\label{subsec:interoperability:mpmd}
+
+MPI defines a dynamic process model that allows creation of processes after
+an MPI application has started (e.g., by calling \FUNC{MPI\_Comm\_spawn}) and
+connection to independent processes (e.g., through \FUNC{MPI\_Comm\_accept}
+and \FUNC{MPI\_Comm\_connect})
+and provides a mechanism to establish communication
+between the newly created processes and the existing MPI application (see
```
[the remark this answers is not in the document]
minsii: Fixed

content/interoperability.tex
```diff
@@ -0,0 +1,189 @@
...
+MPI standard version 3.1, Chapter 10).
+Unlike MPI, \openshmem starts all processes at once and requires all PEs to
+collectively allocate and initialize resources (e.g., symmetric heap) used by
+the \openshmem library before any other \openshmem routine may
+be called. \openshmem does not support communication with dynamically created
+or connected processes. In such a scenario, MPI must be used to communicate
+with these processes.
```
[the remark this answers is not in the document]
minsii: I am actually a little cautious in making the possible communication approaches more restricted, because users may also use other approaches such as socket. We only want to prohibit OpenSHMEM in such scenarios. How about we say?
```In such a scenario, MPI can be used to communicate with these processes.```

File: content/shmem_query_interoperability.tex (hunk 11 above)
Comment (minsii) [reply content/shmem_query_interoperability.tex]: CASE-1: If MPI is provided by a third party implementation, I do not think the OpenSHMEM implementation can guarantee progress for MPI. In such a case, the return value of `SHMEM_PROGRESS_MPI` should be `0` in general. However, this feature may be supported for such a scenario if the implementations agree on a special synchronization protocol (**vendor-specific**). For example:
 - The MPI and OpenSHMEM implementations share the same process manager.
 - They handshake at initialization through the process manager and the MPI implementation exposes a progress hook.
 - If the OpenSHMEM impl already uses a software thread, it can trigger the MPI progress hook to ensure MPI progress.
 - All the above conditions can be confirmed once both MPI and OpenSHMEM are initialized, thus the query API can return the corresponding value.

CASE-2: The support of `SHMEM_PROGRESS_MPI` can be easier if MPI and OpenSHMEM are provided by the same runtime system. The implementation can ensure that the user is using its MPI and OpenSHMEM once the program called `MPI_Init` and `shmem_init` . For example:
- The runtime can set internal flags `is_my_mpi_initialized=1` at  `MPI_Init` and set `is_my_shmem_initialized=1` at `shmem_init`.
- When the user queries  `SHMEM_PROGRESS_MPI`, return `is_my_mpi_initialized && is_my_shmem_initialized ? 1:0`.

I have tested the API in OSHMPI which is similar to CASE-2.

File: content/interoperability.tex [hunk 15]
```diff
@@ -144,10 +144,8 @@ \subsection{RMA Programming Models}
 For instance, \openshmem guarantees the atomicity only of concurrent \openshmem AMO operations
 that operate on symmetric data with the same datatype. Access to the same symmetric
 object with MPI atomic operations, such as an \FUNC{MPI\_Fetch\_and\_op}, may
-result in an undefined result. Furthermore,
-because most RMA programs can be written by using either \openshmem or MPI RMA,
-users should choose only one of the RMA models in the same program, whenever
-possible, for performance and code simplicity.
+result in an undefined result. Users should avoid situations where MPI and
+\openshmem operations perform concurrent accesses to the same memory location.
```
minsii: Make a pass to minimize usage of `users`

content/interoperability.tex
```diff
@@ -111,7 +111,7 @@ \subsection{Thread Safety}
 \subsection{Mapping Process Identification Numbers}
 \label{subsec:interoperability:id}
 
-Similar to the PE identifier in \openshmem, MPI defines rank as the
+Similar to the PE number in \openshmem, MPI defines rank as the
 identification number of a process in a communicator. Both the \openshmem PE
```
minsii: Use \ac{PE}.

nspark: FWIW, I think we can fix all the missing `\ac{...}` macros afterward as editorial changes. (MPI itself should be an acronym, no?)

content/shmem_query_interoperability.tex
minsii: Discussion at today's thread WG meeting:
It is more suitable to expose the progress support of MPI through an MPI API rather than an OpenSHMEM API. Because we do not have a useful case for the query API now, I will postpone the query API part and only go with the first section of the proposal.

content/interoperability.tex
minsii: MPI stands for `Message Passing Interface`. Should we define the acronym macro for it?

content/interoperability.tex
minsii: Over the WG call, we decided to postpone the query API into a separate proposal and we will work on the thread safety properties there.

content/interoperability.tex
minsii: I just found that \ac{MPI} is already defined. Replaced `MPI` with `\ac{MPI}` in the proposal.

minsii:
The PDF version for official reading at Nov 13, 2019 specification meeting:
[openshmem-spec-interoperability-20191029.pdf](https://github.com/openshmem-org/specification/files/3785659/openshmem-spec-interoperability-20191029.pdf)

content/interoperability.tex
```diff
@@ -0,0 +1,163 @@
...
+number of processes. In a hybrid program, the \openshmem
+\ac{PE} number in \LibHandleRef{SHMEM\_TEAM\_WORLD}
+and the \ac{MPI} rank in \VAR{MPI\_COMM\_WORLD} of a process can be equal.
+This feature, however, may be provided by only some of the \openshmem and \ac{MPI}
+implementations (e.g., if both environments share the same underlying process
+manager) and is not portably guaranteed. A portable program should always
+use the standard functions in each model, namely, \FUNC{shmem\_team\_my\_pe} in \openshmem
+and \FUNC{MPI\_Comm\_rank} in \ac{MPI}, to query the process identification numbers
+in each communication environment and manage the mapping of identifiers in the
+program when necessary.
```
davidozog: Maybe it's implicit or maybe I'm missing something, but do we need to also specify that the size of `MPI_COMM_WORLD` is the same as `SHMEM_TEAM_WORLD`?  Otherwise, the example code below might not work correctly, right?  If we do need to specify the sizes are equal, do we have to also account for the usage of `MPI_Comm_Spawn`, which IIUC has a separate `MPI_COMM_WORLD` on the children ranks?

minsii: Good question. You are right that there is no guarantee that size of `MPI_COMM_WORLD` is same as size of `SHMEM_TEAM_WORLD`. I think it depends on two factors:
1.  How the program initializes SHMEM and MPI (e.g., some processes call `MPI_Init` but do not call `shmem_init` )
2. How the implementations define accessible processes (e.g., although all launched processes call both SHMEM and MPI init, some of them might be visible in MPI but not visible in SHMEM).

For the example code, it forces each process to initialize both MPI and SHMEM, thus factor-(1) is confirmed. For factor-(2), I think only the processes who are accessible in both models can execute the code (e.g., any hidden process should not be able to execute the code after init calls). Thus, the size of each "WORLD" should be equal and the program works correctly.

For the program using `MPI_Comm_Spawn`, it should not use SHMEM to communicate with dynamically created processes. We have explicitly defined it in section "Dynamic Process Creation".

Please let me know if I miss any case.

davidozog: Yes, I think you're correct - thanks!

jdinan:
Would it be helpful for the example to also show how to create an MPI communicator that has the same rank and OpenSHMEM PE numbering?  I think the easiest way is something like:
```
MPI_Comm_split(MPI_COMM_WORLD, 0, shmem_my_pe(), &shmem_comm);
```

content/interoperability.tex
```diff
@@ -0,0 +1,163 @@
...
+location; otherwise, the behavior is undefined.
+
+\subsection{Communication Progress}
+\label{subsec:interoperability:progress}
+
+\openshmem promises the progression of communication both with and without
+\openshmem calls and requires the software progress mechanism in the implementation
+(e.g., a progress thread) when the hardware does not provide asynchronous communication
+capabilities. In \ac{MPI}, however, a weak progress semantics is applied. That is,
+an \ac{MPI} communication call is guaranteed only to complete in finite time. For
```
minsii: Add reference to section 4.1.

minsii:
> Would it be helpful for the example to also show how to create an MPI communicator that has the same rank and OpenSHMEM PE numbering? I think the easiest way is something like:
> 
> ```
> MPI_Comm_split(MPI_COMM_WORLD, 0, shmem_my_pe(), &shmem_comm);
> ```

Agreed. Then the user program can use `shmem_comm` instead of `MPI_COMM_WORLD` in the program and safely assume that the rank and PE are equal. I will add the example.

minsii:
The last push contains two changes:
1. Add reference to SHMEM progress definition section  (4.1)
2. Add example that manages rank-PE mapping by using `MPI_Comm_split`.

@manjugv @jdinan Can you please review?

minsii:
Special ballot changes: [ec2ca5fe...9f11fd4b](https://github.com/minsii/openshmem-specification/compare/ec2ca5fe...9f11fd4b)

RaymondMichael:
Instead of:

that contains all
processes in \VAR{MPI\_COMM\_WORLD} and uses the same \ac{MPI} rank and
\openshmem \ac{PE} numbering.

I think the following might read a little easier:

that contains all
processes in \VAR{MPI\_COMM\_WORLD} and each process has the same \ac{MPI} rank number as its \openshmem \ac{PE} number.

minsii:
@RaymondMichael Thanks! Fixed.

minsii:
@jdinan I just noticed that I forget to remove the green color highlights in this PR. Sorry for this mistake. Can I create a different PR to add the fix?
```
\color{ForestGreen}
\input{content/interoperability}
\color{black}
```

jdinan:
Good catch -- DocEdit posted in #327

PR #305, opened 2019-10-29
Revise error handling w.r.t. invalid teams

### LINKED ISSUE #312: Behavior of shmem_team_create_ctx with SHMEM_TEAM_INVALID
Currently undefined

nspark:
This PR includes changes to error handling with respect to team creation and invalid teams. Highlights include:

- `shmem_team_split_2d` will now always succeed or always fail for all child teams on all PEs (in the parent team)
- Clarified the handling of invalid teams vs. `SHMEM_TEAM_INVALID` as team arguments (thanks @davidozog!)
- Clarifying invalid teams with `shmem_team_create_ctx`; closes #312 

Special Ballot change: ac0ff9a


nspark:
[PDF draft as of `befb323`](https://github.com/openshmem-org/specification/files/3781362/openshmem-draft-befb323.pdf)

content/shmem_team_split_strided.tex
```diff
@@ -74,20 +74,20 @@
...
 will be created and \VAR{new\_team} will be assigned the value
-\LibConstRef{SHMEM\_TEAM\_INVALID}; otherwise, the behavior is undefined.
+\LibConstRef{SHMEM\_TEAM\_INVALID}.  If \VAR{parent\_team} is otherwise invalid, the behavior is undefined.
 
 If an invalid \ac{PE} triplet is provided, then the \VAR{new\_team}
 will not be created.
 
 If \VAR{new\_team} cannot be created, then it will be assigned the value
-\LibConstRef{SHMEM\_TEAM\_INVALID}.
+\LibConstRef{SHMEM\_TEAM\_INVALID} and \FUNC{shmem\_team\_split\_strided} will return a nonzero value.
```
davidozog: Related question: `new_team` will  be `SHMEM_TEAM_INVALID` on PEs that are in the parent team but not in the child team, so do those PEs return nonzero?  Seems to me that these PEs should return zero, at least if it was a successful team creation operation across the child team (assuming we have a reduction to assure equal return values across all parent PEs).  If we agree, maybe we could specify that near [line 65](https://github.com/openshmem-org/specification/pull/305/files#diff-16302be989ad41eb9353b7a8359e728cR63-R65) (or here)?

content/shmem_team_translate_pe.tex
```diff
@@ -11,20 +11,23 @@
...
 Specifically, given the \VAR{src\_pe} in \VAR{src\_team}, this routine returns that
 \ac{PE}'s number in \VAR{dest\_team}. If \VAR{src\_pe} is not a member of both the
 \VAR{src\_team} and \VAR{dest\_team}, a value of \CONST{-1} is returned.
 
-
-If either of the \VAR{src\_team} or \VAR{dest\_team} handle is invalid, the behavior is undefined.
+If at least one of \VAR{src\_team} and \VAR{dest\_team} compares equal
+to \LibConstRef{SHMEM\_TEAM\_INVALID}, then a nonzero value is
+returned.
+If either of the \VAR{src\_team} or \VAR{dest\_team} handles are
```
minsii: Should we replace "nonzero value" with "-1" when at least one of the teams is `SHMEM_TEAM_INVALID`? At return value description, it says
```
or a value of -1 if any team handle arguments are invalid or the src_pe is not in both the source and destination teams.
```

content/collective_intro.tex
```diff
@@ -123,6 +123,11 @@ \subsubsection*{Team-implicit collectives}
...
 
-Collective operations involving many \acp{PE} may return values indicating success while other \acp{PE} are still executing the collective operation. Return values indicating success of a collective routine on one \ac{PE} do not indicate that all \acp{PE} involved in the collective operation will return from the routine successfully. In the case where successful local completion of a collective implies the success of some global operation, such as team creation resulting in a valid team on all involved \acp{PE}, the implementation should not return \CONST{0} from the routine if the implied or stated global guarantees of the routine are not met.
+Collective operations involving multiple \acp{PE} may return values
+indicating success while other \acp{PE} are still executing the
+collective operation. Return values indicating success or failure of a
+collective routine on one \ac{PE} may not indicate that all \acp{PE}
+involved in the collective operation will return the same value. Some
+operations, such as team creation, must return consistent return codes
+across multiple \acp{PE}.
```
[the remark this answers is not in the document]
nspark: > Does "must return consistent return values" require all PEs to return _equal_ return values?

Yes, I think to say "equal" is clearer to users and implementers.

> For example, during team creation - if PEs in the active set discover they cannot successfully create a team, do those PEs need to notify the other PEs in the parent team (i.e., outside the active set) that they need to return a nonzero value?

Yes, this is required and, I would argue, the natural behavior an application developer would expect. If team creation could fail on a subset of PEs in the child team, then a portable application is forced to use the parent team to check for success across the child team. There's no guarantee that a collective on the child team would ever be successful otherwise. Executing the reduction inside the implementation is worth the cost to the end user.

> I recall discussing this at the F2F, but don't recall if we decided the cost of a reduction is worth having equal return values across PEs during team creation.

At the F2F, I recall discussing `shmem_team_split_2d` as one of the "important" cases to consider here. While we initially thought there would be a reduction involved, I think we rationalized to ourselves that if any one team on any PE in the parent team fails to create any child team, then all PEs experience the cascading failure.

content/shmem_team_split_strided.tex
nspark: I'm not sure we've ever really settled on a clear behavior here. I think there are two options to consider:

> If creation of the child team is successful:
> 
> 1. All PEs in the parent team are returned zero
> 2. All (and only) PEs in the child team are returned zero; PEs in the parent team, but not in the child team are returned nonzero
> 
> In either case, the `new_team` handle references a valid team for those PEs in the child team and those PEs not in the child team have `new_team == SHMEM_TEAM_INVALID`.
> 
> If creation of the child team is not successful, then all PEs in the parent team should receive a nonzero return value and  `new_team == SHMEM_TEAM_INVALID`.

I think Option 1 is the most clear and user-friendly. It's also most consistent with the return behavior discussed in your earlier question. My rationale here is that the PEs in the parent team may need to do something collectively if the child team is not created. This would be an action of the parent team, not the (nonexistent) child team, so all PEs in the parent team should know whether the team creation operation was successful.

nspark: @davidozog What do you think of the following?

shmem_team_split_strided

content/collective_intro.tex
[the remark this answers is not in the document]
davidozog: Agreed, thanks.

I may not fully understand how a "cascading failure" for 2D-splits obviates a reduction... but I'll think about it.  As of now, I think the SOS prototype might need a reduction in `split_strided`, which is called by `split_2d` at least a couple times on each PE.

content/shmem_team_split_strided.tex
davidozog: This looks great, as long as it's also clear that an unsuccessful creation returns equal nonzero return values.

nspark:
[PDF draft as of `3020e9e`](https://github.com/openshmem-org/specification/files/3843396/openshmem-draft-3020e9e.pdf)

PR #306, opened 2019-10-29
Define atomicity domains w.r.t. teams

nspark:
This PR defines atomicity domains as discussed in prior F2F and WG meetings.


nspark:
[PDF draft as of `69e541a`](https://github.com/openshmem-org/specification/files/3781690/openshmem-draft-69e541a.pdf)

naveen-rn:
If I understand this correctly, this change just allows two atomicity domains. What is the rationale behind not creating an atomicity domain per team?

naveen-rn:
> This PR only defines two atomicity domains. We could add extra ones in the future.
I haven't looked at the various pros and cons of one atomicity domain per team. I agree we could relax the proposal in future and add more domains.

As I read the new changes, the major issue is that this 2 domain semantic brings in another new dimension to the OpenSHMEM Teams - *hierarchy*. We haven't had more discussions on this. It would be safe, if we don't have any parent-child team relation once the team is created. I'm afraid that the general progression of introducing hierarchy, would slowly seep into other features that we plan to introduce (like SHEAP per team).

nspark:
Team hierarchy here is a relatively weak concept. We only use the "hierarchy" to identify the associated atomicity domain. (If the term itself is problematic, I could probably reword the text to avoid it.) As an example of its weakness, we don't destroy "child" teams when the "parent" team is destroyed. Accessibility w.r.t. team-based symmetric heaps is a separate issue&mdash;one for which I think we (collectively) could decide either way.

nspark:
@naveen-rn Here's an attempt to avoid the word "hierarchy":

> An OpenSHMEM _atomicity domain_ is a set of communication contexts whose associated teams (see Section ...) are all split by (possibly recursive) calls to a `shmem_team_split_*` routine from a common predefined team. OpenSHMEM provides two such predefined teams, `SHMEM_TEAM_WORLD` and `SHMEM_TEAM_SHARED` (see Section ...).
>
> _...footnote..._
> Although all PEs in `SHMEM_TEAM_SHARED` are also in `SHMEM_TEAM_WORLD`, and a PE's number can be translated from `SHMEM_TEAM_SHARED` to `SHMEM_TEAM_WORLD`, the `SHMEM_TEAM_SHARED` team is defined as not having been created by a call to `shmem_team_split_*` on `SHMEM_TEAM_WORLD`. Therefore, the two teams are distinct predefined teams forming separate atomicity domains.

nspark:
[PDF draft as of `96dc1c9`](https://github.com/openshmem-org/specification/files/3786565/openshmem-draft-96dc1c9.pdf)

example_code/amo_scenario_3.c
```diff
@@ -1,16 +1,21 @@
...
-     * increment of symmetric variable 'x'. */
-    if (shmem_my_pe() > 0)
-      shmem_int_atomic_inc(&x, 0);
-    else
-      x++;
+    shmem_int_atomic_inc(&x, (shmem_my_pe()+1) % shmem_n_pes());
+    /* Undefined behavior: The following reduction operation performs accesses to symmetric
+     * variable 'x' that are concurrent with previously issued atomic increment operations
+     * on the same variable. */
+    shmem_int_sum_to_all(&y, &x, 1, 0, 0, shmem_n_pes(), pwrk, psync);
```
minsii: Do we want to use the team-based collective API?

nspark: We could&mdash;and maybe should&mdash;make this change, but I didn't actually change this example in this PR. It's hard to see in the GitHub diff, but I structured the commits to help show that I renamed the examples to put the atomicity domain example first. So, I didn't actually change this example, though the diff on the PR shows otherwise.

PR #316, opened 2019-11-24
Per PE Quiet interface

### LINKED ISSUE #317: Per PE Quiet
## Problem :
To complete all RMA operations that are posted from a particular PE to a particular target PE, the OpenSHMEM programs can call shmem_quiet. The shmem_quiet operation, however, completes the RMA operations from the calling PE to all PEs. For system configuration, where shmem_quiet is expensive this is unnecessary overhead. 

## Proposal : 
Introduce PE specific completion routines that complete RMA operations from the calling PE to the target PE

`shmem_pe_quiet(target_pe)` 
`shmem_ctx_pe_quiet(ctx, target_pe)`

[PR](https://github.com/openshmem-org/specification/pull/316) describing the interface and semantics
 
## Impact on Users:
This provides a light-weight alternative for shmem_quiet if the OpenSHMEM programs desire to complete operations from the calling PE to the target PE. There is no change in semantics for the current shmem_fence interface.
 
## Impact on implementation:
Implementations will have to implement the new interface shmem_pe_quiet(target_pe) and its context variants shmem_ctx_quiet(ctx, int target_pe)

manjugv:
Closes #317

### DIFF
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -673,11 +673,10 @@ The following list describes the specific changes in \openshmem[1.6]:
     \openshmem[1.5] Table 10, and clarified the types, names, and supporting
     operations for team-based reductions.
 \ChangelogRef{teamreducetypes}%
-
-\item Added fine grained completion routine: \FUNC{shmem\_pe\_quiet}
+%
+\item Added fine grained completion routine: \FUNC{shmem\_pe\_quiet}.
 \ChangelogRef{subsec:shmem_pe_quiet}%
-
-
+%
 \item Split the listings for the \FUNC{shmem\_\{malloc, free, realloc, align\}}
   functions from a single entry in \openshmem[1.5] into separate entries.
 \ChangelogRef{subsec:shmem_malloc, subsec:shmem_free, subsec:shmem_realloc,
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -668,6 +668,10 @@ The following list describes the specific changes in \openshmem[1.6]:
     \openshmem[1.5] Table 10, and clarified the types, names, and supporting
     operations for team-based reductions.
 \ChangelogRef{teamreducetypes}%
+
+\item Added fine grained completion routine: \FUNC{shmem\_pe\_quiet}
+\ChangelogRef{subsec:shmem_pe_quiet}%
+
 \end{itemize}
 
 \section{Version 1.5}
--- a/content/shmem_pe_quiet.tex
+++ b/content/shmem_pe_quiet.tex
@@ -9,8 +9,8 @@
 \begin{apidefinition}
 
 \begin{Csynopsis}
-void @\FuncDecl{shmem\_pe\_quiet}@(int *target_pes, size_t npes);
-void @\FuncDecl{shmem\_ctx\_pe\_quiet}@(shmem_ctx_t ctx, int *target_pes, size_t
+void @\FuncDecl{shmem\_pe\_quiet}@(const int *target_pes, size_t npes);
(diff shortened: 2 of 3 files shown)


agrippa:
As a non-implementer, I may need some help with this.

Is there a performance benefit to this routine over maintaining 1 context per PE I communicate with? At the application level that might undesirably lead to memory consumption scaling with # of PEs, so are there tricks in the runtime layer that mean it won't?

Could the options parameter to shmem_ctx_create be used to set a property on contexts (e.g. SHMEM_ONLY_USED_WITH_ONE_PE) that could have a similar effect rather than introducing a new API?

manjugv:
@agrippa That is definitely one way to do it. As a matter of fact, with that option, one can do more optimizations. However, it is much stronger semantic than the proposal here. The context is creating a communication stream. With your proposed change, you are creating a stream per PE and you have an ability to couple that stream with a thread.

shamisp:
@manjugv UCX support this, but if you look at the code it comes with the cost. It is not huge but you have to track outstanding messages on the PE. If this semantics is useful for OpenSHMEM users, I'm all for it.

nspark:
WG feedback: Clarify that the `target_pes` pointer isn't accessed if `npes` is zero.

content/shmem_pe_quiet.tex
```diff
@@ -0,0 +1,50 @@
...
+    objects issued by the calling \ac{PE} to the target \acp{PE} and on the
+    given context.
+	
+	The completion and visibility semantics of these operations are the same as the
+	\FUNC{shmem\_quiet} routine. However, it applies only to the target
+    \acp{PE}, i.e., the routines to the target \acp{PE} are guaranteed to be
+    complete and visible to all \acp{PE} when \FUNC{shmem\_pe\_quiet} returns.
+    When calling with the \FUNC{shmem\_ctx\_pe\_quiet} interface, the updates are
+    guaranteed to be complete and visible only to the \acp{PE} in the team used to
+    create the context.
```
nspark, line 40: WG Discussion: Consider adding an `options` flag to API; remove this restriction; and make this restriction one of the optional values.

manjugv: Will remove this (as per WG discussion) and address it via another PR. Issue is captured here : https://github.com/openshmem-org/specification/issues/487

manjugv:
@jdinan Do you need anything else on this before merging?

content/backmatter.tex
```diff
@@ -673,7 +673,11 @@ \section{Version 1.6}
     \openshmem[1.5] Table 10, and clarified the types, names, and supporting
     operations for team-based reductions.
 \ChangelogRef{teamreducetypes}%
-%
+
+\item Added fine grained completion routine: \FUNC{shmem\_pe\_quiet}
+\ChangelogRef{subsec:shmem_pe_quiet}%
+
+
```
jdinan, line 680:
```suggestion
%
\item Added fine grained completion routine: \FUNC{shmem\_pe\_quiet}.
\ChangelogRef{subsec:shmem_pe_quiet}%
%
```

jdinan:
@manjugv I added a few doc editor updates for you to review and apply. I also see that there are unresolved comments on this PR. Please resolve these, e.g. by adding doc edit changes to this PR, filing new issues, new PRs, etc.

manjugv:
@jdinan I accepted the commit. Thanks.

jdinan:
@manjugv Could you please resolve the open comments on this ticket before merging? It looks like some of these should have issues created for follow-up before the next spec release.

manjugv:
@jdinan I have resolved all comments. Thanks.

PR #322, opened 2019-12-08
Adding realloc semantics for shmem_malloc_with_hints routine

manjugv:
The description of the behavior of shmem_realloc when ptr allocated with shmem_malloc_with_hints routine is passed to it.

### DIFF
--- a/content/shmem_malloc_hints.tex
+++ b/content/shmem_malloc_hints.tex
@@ -44,6 +44,17 @@ void *@\FuncDecl{shmem\_malloc\_with\_hints}@(size_t size, long hints);
     The user is also responsible for calling these routines with identical argument(s) on all
     \acp{PE}; if differing \VAR{size}, or \VAR{hints} arguments are used, the behavior of the call
     and any subsequent \openshmem calls is undefined.
+
+	Calling the routine \FUNC{shmem\_realloc(void *ptr, size\_t size)} on the
+	ptr allocated by \FUNC{shmem\_malloc\_with\_hints} has a behavior similar to
+	calling \FUNC{shmem\_realloc} on the ptr allocated by \FUNC{shmem\_malloc}.
+	If the new size is larger, the memory is allocated with the same hints that
+	were passed when calling \FUNC{shmem\_malloc\_with\_hints}, and the newly
+	allocated portion of the block is uninitialized. If ptr is a null pointer,
+	the \FUNC{shmem\_realloc} routine behaves like the \FUNC{shmem\_malloc}
+	routine for the specified size. If the size is 0 and ptr is not a null pointer,
+	the block to which it points is freed. If space cannot be allocated, the
+	block to which ptr points is unchanged.
 }
 
 \apireturnvalues{


manjugv:
I'm adding this PR as discussed at the November 2019 meeting. I will read this ticket at the January meeting. If we think the chapter committee approval is good and does not require full committee approval, we can move it to the chapter committee.

content/shmem_malloc_hints.tex
```diff
@@ -44,6 +44,17 @@
     The user is also responsible for calling these routines with identical argument(s) on all
     \acp{PE}; if differing \VAR{size}, or \VAR{hints} arguments are used, the behavior of the call
     and any subsequent \openshmem calls is undefined.
+
+	Calling the routine \FUNC{shmem\_realloc(void *ptr, size\_t size)} on the
+	ptr allocated by \FUNC{shmem\_malloc\_with\_hints} has a behavior similar to
+	calling \FUNC{shmem\_realloc} on the ptr allocated by \FUNC{shmem\_malloc}.
+	If the new size is larger, the memory is allocated with the same hints that
+	were passed when calling \FUNC{shmem\_malloc\_with\_hints}, and the newly
```
jdinan: What about if the new size is smaller?

jdinan: Everything below this comment is duplicated from the `shmem_realloc` section. I think I am in favor of only keeping the first two sentences and moving them to where realloc is specified.

jdinan: I think we can add to the end of the `shmem_realloc` paragraph. I tried consider the case where realloc is used several times on the same buffer:
> If \VAR{ptr} corresponds to memory initially allocated by \FUNC{shmem\_malloc\_with\_hints}, the resulting memory will comply with the hints used in the initial allocation.

manjugv: > What about if the new size is smaller?

I did not define that behavior since we don't define it for shmem_realloc when ptr is allocated via shmem_malloc. I did not want to introduce new semantics. 

However,  it makes sense to define it as "shrink the size of memory". This change should be for shmem_realloc in general, not just wrt memory allocated via shmem_malloc_with_hints.

manjugv:
As per the discussion at the spec meeting, this PR will move to the chapter committee.

jdinan:
@manjugv I took an AR to propose an update to `shmem_malloc.tex` instead of the proposed update to `shmem_malloc_hints.tex` on this PR. With apologies for the slow response, here is the proposed patch:

```
diff --git a/content/shmem_malloc.tex b/content/shmem_malloc.tex
index 17a004d1..90cd4a7b 100644
--- a/content/shmem_malloc.tex
+++ b/content/shmem_malloc.tex
@@ -45,13 +45,16 @@ void *@\FuncDecl{shmem\_align}@(size_t alignment, size_t size);
            
     The \FUNC{shmem\_realloc} routine changes the size of the block to which
     \VAR{ptr} points to the size (in bytes) specified by \VAR{size}.  The contents
-    of the block are unchanged up to the lesser of the new and old sizes. If the new
+    of the block are unchanged up to the lesser of the new and old sizes.
+    The \FUNC{shmem\_realloc} routine preserves allocation hints, e.g. if `ptr`
+    was allocated by \FUNC{shmem\_malloc\_with\_hints}. If the new
     size is larger, the newly allocated portion of the block is
     uninitialized.  If \VAR{ptr} is a null pointer, the
     \FUNC{shmem\_realloc} routine behaves like the \FUNC{shmem\_malloc} routine for
     the specified size.  If \VAR{size} is \CONST{0} and \VAR{ptr} is not a
     null pointer, the block to which it points is freed. If the space cannot
-    be allocated, the block to which \VAR{ptr} points is unchanged.
+    be allocated or if hints cannot be preserved, the block to which \VAR{ptr}
+    points is unchanged.
     
     The \FUNC{shmem\_malloc}, \FUNC{shmem\_align}, \FUNC{shmem\_free}, and \FUNC{shmem\_realloc} routines
     are provided  so that multiple \acp{PE} in a program can allocate symmetric,
```

manjugv:
@jdinan Look good for me! Thanks.

manjugv:
@swpoole @agrippa @jamesaross @tonycurtis @minsii Please approve or disapprove and suggest any changes. Thanks.

minsii:
@manjugv @jdinan One question for Jim's patch: If hints cannot be preserved for realloc-block, should the block be unchanged? Or should it be resized with no hint?

tonycurtis:
Looks ok...

manjugv:
> @manjugv @jdinan One question for Jim's patch: If hints cannot be preserved for realloc-block, should the block be unchanged? Or should it be resized with no hint?

@minsii It is unchanged. In a way, it is in the same spirit as realloc.

jdinan:
@minsii My preference is for unchanged. A reallocation without hints can't be distinguished from a reallocation that preserved hints in the current API.

minsii:
@manjugv @jdinan The reasons make sense to me. My only concern was that it somehow takes the hints as a "requirement" of `realloc` but not only a "hint." But I agree that there seems no better option than `unchanged`.

Approve the change.

manjugv:
Ping @swpoole @agrippa @jamesaross

jamesaross:
@jdinan's patch is better.

Codes really shouldn't be reallocating (at least expanding) heap memory.  OpenSHMEM should have used brk/sbrk from the start :-)

Side note: It's odd that I do not receive mentions in email or otherwise.  I have to look through the notifications, which is why I've been delayed.

content/shmem_malloc_hints.tex
```diff
@@ -44,6 +44,17 @@
     The user is also responsible for calling these routines with identical argument(s) on all
     \acp{PE}; if differing \VAR{size}, or \VAR{hints} arguments are used, the behavior of the call
     and any subsequent \openshmem calls is undefined.
+
+	Calling the routine \FUNC{shmem\_realloc(void *ptr, size\_t size)} on the
```
agrippa: Twice here it says "the ptr allocated", I think it would be better to say "a ptr allocated".

content/shmem_malloc_hints.tex
```diff
@@ -44,6 +44,17 @@
...
+	ptr allocated by \FUNC{shmem\_malloc\_with\_hints} has a behavior similar to
+	calling \FUNC{shmem\_realloc} on the ptr allocated by \FUNC{shmem\_malloc}.
+	If the new size is larger, the memory is allocated with the same hints that
+	were passed when calling \FUNC{shmem\_malloc\_with\_hints}, and the newly
+	allocated portion of the block is uninitialized. If ptr is a null pointer,
```
agrippa: Why not just make this wording consistent with C realloc and say "the contents of the new part of the array are undefined"?

manjugv:
@agrippa You are looking at old changes. As discussed during the specification meeting, @jdinan changes to shmem_malloc.tex will replace the changes to shmem_malloc_hints.tex. 

Please look here https://github.com/openshmem-org/specification/pull/322#issuecomment-577356017

jdinan:
@manjugv Would it be helpful to apply that change to this PR so the "Files Changed" tab accurately reflects what is being proposed? I'm going to ask you to do this anyway before I can click the merge button. :neckbeard:

manjugv:
@jdinan Please look here: https://github.com/manjugv/specification/pull/12

manjugv:
@agrippa Please approve this https://github.com/manjugv/specification/pull/12

swpoole:
manjugv:

approved

manjugv:
Replaced by manjugv#12

PR #327, opened 2020-01-10
Remove change highlighting from #304

jdinan:


jdinan:
@minsii Please review -- thanks!

PR #330, opened 2020-01-14
Fix LaTeX to support and remove Notes/None

nspark:
This PR:

- Fixes the LaTeX environment to allow removing `\apinotes{}`
- Removes all instances of `\apinotes{ None. }`

Replaces #87

### DIFF
--- a/utils/defs.tex
+++ b/utils/defs.tex
@@ -399,6 +399,7 @@
 \vspace{-2em}
 }
 {
+\end{description} % ends \begin{description} from \apidescription
 \end{description}
 }
 
@@ -511,7 +512,6 @@
 \item[Notes] \hfill \\
     #1
 \hfill \\
-\end{description}
 }
 
 \newcommand{\apiimpnotes}[1]{
--- a/utils/defs.tex
+++ b/utils/defs.tex
@@ -515,14 +515,14 @@
 }
 
 \newcommand{\apiimpnotes}[1]{
-\item[Note to implementors] \hfill \\
+\item[Note to implementers] \hfill \\
     #1
 \hfill \\
 }
 
 \newcommand{\parimpnotes}[1]{
 \begin{description}
-\item[Note to implementors] \phantom{a} \newline
+\item[Note to implementers] \leavevmode \\
     #1
 \end{description}
 }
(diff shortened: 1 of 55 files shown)


utils/defs.tex
```diff
@@ -399,6 +399,7 @@
 \vspace{-2em}
 }
 {
+\end{description} % ends \begin{description} from \apidescription
```
jdinan: This change appears to mess up the indenting of the "Notes to Implementors", e.g. in shmem_malloc and friends. Wondering if it would be more robust to have start and end macros for apidescription, since the last thing in the API description varies?

nspark: This is easy to fix, but breaks one other case:

https://github.com/nspark/specification/blob/7e5cd8f9f17de95eafaec0250009c620f9c622dc/content/interoperability.tex#L37-L45

Other than that usage, `\apiimpnotes` is not used outside of a `description` environment anywhere in the Specification, despite the appearance in the Execution Model:

https://github.com/nspark/specification/blob/7e5cd8f9f17de95eafaec0250009c620f9c622dc/content/execution_model.tex#L44

I'm inclined (and have a WIP) to go with the easy fix (that corrects the styling errors you note) and "fix" the use of `\apiimpnotes` in the Interoperability section with something comparable to what's in the Execution Model section. What do you think? This would be _much less_ churn than making "SYNOPSIS" and "DESCRIPTION" proper environments (which would be a better overall solution, but would touch every API source file).

jdinan: Makes sense to me. IIUC, `\apiimpnotes` should only be used within an API description block. Would be good to introduce a different implementation note macro that can be used outside of API descriptions.

nspark: :+1: I was working on that, actually, but stopped when I had some whitespace issues in the usage under "Execution Model". I'll pick that back up either this week or next.

nspark:
Ugh. Here's one of those "it works on my machine" problems.

nspark:
Renders of the PDF as of e12a415:

screenshot-note-to-impl

nspark:
@manjugv @jdinan This PR is ready for review; I'd like to have an informal review at the F2F

Issue #328, opened 2020-01-10
Fix restriction on xrange in shmem_team_split_2d


nspark:
We need to revise `shmem_team_split_2d` as follows:

> The value of `xrange` must be ~~nonnegative~~ **positive** and all PEs in the parent team must pass the same value for `xrange`.

Issue #326, opened 2020-01-07
2D team split behavior when xrange > npes


davidozog:
@jdinan and I ran into a minor implementation problem when testing a 2D team split with `xrange > npes` here:
https://github.com/davidozog/sandia-shmem/pull/9

Perhaps I'm missing something in my interpretation of the API description, but I'm not sure it's explicit about this case.  Do we create an "unfilled" xteam (similar to the unfilled yteams when `npes(parent_team) % xrange > 0`), or should we require that `xrange` be less than or equal to the number of PEs in the parent team?

For what it's worth, I think it's a bit more self-consistent to create the unfilled xteam; but it's also simpler to just require `xrange <= npes`.  Any thoughts?

naveen-rn:
duplicate of https://github.com/gmegan/specification/issues/122

jdinan:
@naveen-rn Thanks for the reference. Now that teams is merged, I suggest that we move issues to the openshmem-org repository, so that we can better track them relative to the OpenSHMEM 1.5 release.

nspark:
I'm happy to work with @davidozog on suggesting a clarifying note to the Teams section committee.

PR #331, opened 2020-01-16
Clarify 2D split behavior when xrange > parent size

davidozog:
This addresses issues #326 and #328.

As discussed on the last call, this is likely ok as a doc edit.  Is that correct @jdinan?

Also, we could maybe use `$N$` instead of "size of the parent team" if that's preferred...


jdinan:
Looks good to me. This should be handled as a section committee edit -- @agrippa @gmegan @wrrobin -- please review. Thanks!

jdinan:
@agrippa @gmegan @wrrobin Please review.

jdinan:
Please move this PR to the relevant section committee's branch.

jdinan:
@davidozog Looks like we missed this in the section committee. Updating to be processed at the FRM.

PR #350, opened 2020-02-05
Remove line numbers from the PDF

nspark:
This PR removes the line numbering on the pages of the Specification. These numbers are not referenced in the document itself, nor do they actually align with the lines of text. They cause copy-paste errors for examples (see https://github.com/openshmem-org/specification/pull/237#issuecomment-582513835).


minsii:
Agreed.

PR #353, opened 2020-02-05
Undefined behavior clause added for invalid ctx value

wrrobin:
Signed-off-by: Md <md.rahman@intel.com>


agrippa:
This looks good to me, but I think this PR should be opened against Jim's chapter committee branch.

PR #354, opened 2020-02-05
Clarify shmem_team_split_strided behavior for a zero size triplet

agrippa:
I've taken a first stab at this clarification, but I'm open to disagreement on this.


agrippa:
Accidentally opened against main repo, need to open against Jim's fork for chapter committee.

Issue #278, opened 2019-07-22
Enumerate list of routines handled by shmem_fence and shmem_quiet


naveen-rn:
We have too many items in the shmem_fence and shmem_quiet routines explanations - we can convert them into a list and then clean the texts. This was based on the comments during the put-with-signal routine reading during the July F2F meeting.

naveen-rn:
Closed as part of  https://github.com/naveen-rn/specification/pull/6

PR #356, opened 2020-02-05
Clean list of Operations supported by fence and quiet

naveen-rn:
Fix for https://github.com/openshmem-org/specification/issues/278

### DIFF
--- a/content/shmem_quiet.tex
+++ b/content/shmem_quiet.tex
@@ -1,6 +1,6 @@
 \apisummary{
-    Waits for completion of all outstanding operations as listed in
-    Table~\ref{quietops} to symmetric data objects issued by a \ac{PE}.
+    Waits for completion of outstanding operations on symmetric data objects
+    issued by a \ac{PE}.
 }
 
 \begin{apidefinition}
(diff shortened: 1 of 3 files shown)


naveen-rn:
@nspark @BryantLam - Can you verify whether this change makes sense? This is the initial take on the
enumeration issue - we can update it as required.

content/shmem_quiet.tex
```diff
@@ -38,23 +34,21 @@
...
     symmetric data objects on all \acp{PE}.
 
     In an \openshmem program with multithreaded \acp{PE}, it is the
-    user's responsibility to ensure ordering between operations issued by the threads
-    in a \ac{PE} that target symmetric memory (e.g. \PUT{}, \ac{AMO}, \OPR{put-with-signal}, memory stores,
-    and nonblocking routines) and calls by threads in that \ac{PE} to
+    user's responsibility to ensure ordering between operations issued by the
+    threads in a \ac{PE} that target symmetric memory
+    (e.g. \PUT{}, \ac{AMO}, \OPR{put-with-signal}, memory stores,
+    and nonblocking routines)
```
BryantLam: What's the reason for why this list of operations in parentheses were left in when the others were removed?

main_spec.tex
```diff
@@ -475,6 +475,30 @@ \subsection{Memory Ordering Routines}\label{subsec:memory_order}
 \OPR{put-with-signal}, \GET{}, and \ac{AMO} routines to symmetric data
 objects.
```
BryantLam: This description is now redundant with the table. Can you shorten it similar to how you shorted the phrasing in the other changes?

BryantLam: Also, there might be other places like this. It looks like you caught most of them, but you could further check by e.g., grep'ing for things like "memory store".

jdinan:
Please move this PR to the relevant section committee's branch.

Issue #357, opened 2020-02-05
Remove instance of "partial update"


nspark:
`shmem_signal_wait_until` includes:

> **Note to implementors**
> > Implementations must ensure that `shmem_signal_wait_until` do not return before the update of the memory indicated by `sig_addr` is fully complete. Partial updates to the memory must not cause `shmem_signal_wait_until` to return.

We've tried to remove references to "partial updates", and this is one lingering issue. Suggest:

> Implementations must ensure that **calls to** `shmem_signal_wait_until` do not return before the update of the memory indicated by `sig_addr` is ~~fully~~ complete. ~~Partial updates to the memory must not cause `shmem_signal_wait_until` to return.~~

jdinan:
#358 was reverted, making a note here to cherry-pick the relevant commit from openshmem-org:master and merge into the relevant section committee's branch.

jdinan:
@naveen-rn Has this issue been addressed?

naveen-rn:
Issue closed as part of https://github.com/naveen-rn/specification/pull/5

PR #358, opened 2020-02-05
Update full/partial completion to signal_wait_until

naveen-rn:
We were addressing the partial update completions in the apinotes
for the shmem_wait_until_signal op. To be consistent with other
sync operations - moving to the description section.

Addresses issue #357

### DIFF
--- a/content/shmem_signal_wait_until.tex
+++ b/content/shmem_signal_wait_until.tex
@@ -33,18 +33,15 @@ uint64_t @\FuncDecl{shmem\_signal\_wait\_until}@(uint64_t *sig_addr, int cmp, ui
     blocks until the value of \VAR{sig\_addr} at the calling \ac{PE} satisfies
     the wait condition specified by the comparison operator, \VAR{cmp}, and
     comparison value, \VAR{cmp\_value}.
+
+    Implementations must ensure that \FUNC{shmem\_signal\_wait\_until} do not
+    return before the update of the memory indicated by \VAR{sig\_addr} is
+    fully complete.
 }
 
 \apireturnvalues{
     Return the contents of the signal data object, \VAR{sig\_addr}, at the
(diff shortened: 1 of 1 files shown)


naveen-rn:
@nspark and @BryantLam - this change is in sync with other ```sync``` routines description

naveen-rn:
If I get approval from @BryantLam  - we can merge this one.

jdinan:
@naveen-rn I am reverting these changes from openshmem-org:master. Please re-open a PR to incorporate them into the relevant section committee branch. This is needed so that we can they can be tracked and processed according to our procedures. It will be a revert commit, you can cherry-pick the commits again from master.

jdinan:
A comment on the technical bits of this change -- maybe I am missing something since I'm only looking at part of the test, but IIUC signaling operations should have atomic semantics, otherwise signal add doesn't work. Please consider defining these operations as atomic with respect to signal updates and wait/test according to the atomicity semantics in front matter.

naveen-rn:
@jdinan My mistake - I didn't notice the merge branch. WIll fix. 

Regarding the technical bit - we have the following statement in the put-with-signal intro:

```tex
All signaling operations on the signal data object completes as if performed
atomically with respect to the following:
\begin{itemize}
    \item other blocking or nonblocking variant of the put-with-signal routine
    that updates the signal data object using the same signal update operator;
    \item signal-fetch routine that fetches the signal data object; and
    \item any point-to-point synchronization routine that accesses the signal
    data object.
\end{itemize}
```

Is it sufficient?

jdinan:
Thanks - I think it's fine. It matches what we also have in the wait/test APIs. Please ignore my earlier comment.

naveen-rn:
@jdinan are these changes reverted from openshmem-org:master?

jdinan:
Yes, reverted by #367

Issue #293, opened 2019-09-26
Teams: Vendor additions to shmem_team_config_t


BryantLam:
Related https://github.com/gmegan/specification/issues/153

From discussion of Teams at Sep 2019 F2F, this issue notes that the Teams feature needs to clarify the behavior of implementation-specific options in the `shmem_team_config_t` struct.

shamisp:
@BryantLam Is there community agreement on this ? I was not present at the meeting ?

BryantLam:
@shamisp - From the full-group discussion at the Sep 2019 F2F, I filed this issue as an action item for the Teams working group. I don't think a conclusion was reached that day.

naveen-rn:
Duplicate issue https://github.com/openshmem-org/specification/issues/342 - closing the other issue - as this has more discussion traffic

shamisp:
@BryantLam - I think we have two separate issues here.
1. Do we allow vendor specific extension within existing text. I don't think we prohibit this explicitly but I don't think we have spell this out in the text.

2. Current ```int shmem_team_get_config(shmem_team_t team, shmem_team_config_t *config)```
does not pass config mask as a separate argument and as a result binary backward compatibility is not possible. We have two options here: 
* integrate the mask into shmem_team_config_t. We woold have to remove config_mask from all the rest team create/split APIs

* add mask only to the shmem_team_get_config

We would have to community to agree on one of the changes. Either way it is change for API.

naveen-rn:
> integrate the mask into shmem_team_config_t. We woold have to remove config_mask from all the rest team create/split APIs

+1

BryantLam:
@shamisp - Oops. You're correct. This issue is specifically about vendor extensions. I personally would favor consistency and if the other APIs already have `config_mask`, adding it here to `shmem_team_get_config` is reasonable.

shamisp:
Proposed #363 please take a look and comment.

jdinan:
If we feel that users setting the config mask is too bothersome, we can add a compile time constant `SHMEM_TEAM_CONFIG_MASK_ALL` that users can pass to query all fields. Since it is a compile time constant from the SHMEM library used to build the application, implementations can set the bits so that it is forward compatible with future ABI as we have discussed.

shamisp:
Patch moved here https://github.com/jdinan/specification/pull/9

PR #363, opened 2020-02-06
TEAMS: Adding config_mask to shmem_team_get_config

shamisp:
Signed-off-by: Pavel Shamis (Pasha) <pasharesearch@gmail.com>

### DIFF
--- a/content/shmem_team_get_config.tex
+++ b/content/shmem_team_get_config.tex
@@ -5,18 +5,21 @@
 \begin{apidefinition}
 
 \begin{Csynopsis}
-int @\FuncDecl{shmem\_team\_get\_config}@(shmem_team_t team, shmem_team_config_t *config);
+int @\FuncDecl{shmem\_team\_get\_config}@(shmem_team_t team, long config_mask, shmem_team_config_t *config);
 \end{Csynopsis}
 
 \begin{apiarguments}
   \apiargument{IN}{team}{An \openshmem team handle.}
+  \apiargument{IN}{config\_mask}{
+    The bitwise mask representing the set of configuration parameters to fetch from the given team.
+    }
   \apiargument{OUT}{config}{
     A pointer to the configuration parameters for the given team.}
 \end{apiarguments}
 
 \apidescription{
 \FUNC{shmem\_team\_get\_config} returns through the \VAR{config} argument
-the configuration parameters of the given team, which were assigned according
+the configuration parameters as described by the mask, which were assigned according
 to input configuration parameters when the team was created.
 
 If \VAR{team} compares equal to \LibConstRef{SHMEM\_TEAM\_INVALID},


shamisp:
Fixing #293

content/shmem_team_get_config.tex
```diff
@@ -5,18 +5,21 @@
 \begin{apidefinition}
 
 \begin{Csynopsis}
-int @\FuncDecl{shmem\_team\_get\_config}@(shmem_team_t team, shmem_team_config_t *config);
+int @\FuncDecl{shmem\_team\_get\_config}@(shmem_team_t team, long config_mask, shmem_team_config_t *config);
 \end{Csynopsis}
 
 \begin{apiarguments}
   \apiargument{IN}{team}{An \openshmem team handle.}
+  \apiargument{IN}{config\_mask}{
```
naveen-rn: This should be a OUT

shamisp: Since the memory for config is allocated by user, it has to be in ?

shamisp:
@jdinan I will keep the PR on this tree as for now. Once we done with discussion, I will open an PR on your tree.

content/shmem_team_get_config.tex
BryantLam: @naveen-rn -- The description for this mask reads like an input. Why do you think it should be `OUT`?

naveen-rn: Initially, I was under the assumption that the memory of the ```shmem_config_t``` mask is allocated by the implementation. But, after discussing with Pasha, it looks like the implementation should just update the values, and the allocation should be handled by the users. 
Since, this is the case, I think IN is fine,

shamisp: @BryantLam OpenSHMEM does not define release function for ```shmem_config_t```, as a result openshmem library should not be allocation it all together. User has to allocate memory for the struct and pass it to the library. So, the structure is always in.

content/shmem_team_get_config.tex
```diff
@@ -5,18 +5,21 @@
...
+    }
   \apiargument{OUT}{config}{
     A pointer to the configuration parameters for the given team.}
 \end{apiarguments}
 
 \apidescription{
 \FUNC{shmem\_team\_get\_config} returns through the \VAR{config} argument
-the configuration parameters of the given team, which were assigned according
+the configuration parameters as described by the mask, which were assigned according
 to input configuration parameters when the team was created.
```
jdinan: According to this, the implementation will always need to return the exact values provided when the team was created. It is not able to return the values actually used by the implementation. Are we /really/ sure this is what we want? There is no opportunity for the implementation to provide feedback in this case.

shamisp: I don't think the new description changes previous meaning, at least as I read it.  So, I don't think there is semantics change.

jdinan:
Please move this PR to the teams section branch.

content/shmem_team_get_config.tex
jdinan: Agree you are not changing semantics, but we also did not spend any time discussing this specific semantic in the teams proposal. Is what we have now the right thing?

shamisp:
@jdinan - Sorry , where is section branch.

content/shmem_team_get_config.tex
shamisp: I think it is useful to return the values actually used by implementation. As I read it, I don't think the above prevents implementation from returning actually used values. Do you suggest to be more explicit about this ?

jdinan: Given the pain we just went through in the MPI Forum to clarify this exact semantic -- yes, I strongly encourage not leaving this open to interpretation. You could look at the text I wrote for MPI info in the [2019 draft](https://www.mpi-forum.org/) on page 271 under `MPI_Comm_get_info`. The text took several revision cycles before we got it right.

jdinan:
@shamisp jdinan:sec/ctx-teams

shamisp:
@jdinan seems like GitHub does not have an option to move PR to another fork. So I guess the right thing todo is to close this one and reopen one on your tree ?

jdinan:
You can change the base branch using the edit button at the top, but I don't think you can change it to a different fork.

shamisp:
Discussion moved here https://github.com/jdinan/specification/pull/9

shamisp:
Opened new pr on your tree

PR #372, opened 2020-02-12
Section 9.4-9.5: Contexts/Teams

jdinan:
Please post section edits to jdinan:sec/ctx-teams


jdinan:
@nspark I seem to have hit this issue on my PR https://github.com/actions/checkout/issues/23. There's a solution buried in that thread to retrieve the SHA. Wonder if it's worth trying?

nspark:
@jdinan Looks like the fix is to "upgrade" to actions/checkout@v2.

PR #398, opened 2020-03-17
Update example formatting

### LINKED ISSUE #397: Make examples easier to read
IMO, the examples are not so easy to read with the current choice of color stylings. Ideas to make it easier?

- Change the background color to 80% or 90% white instead of the current ~50% white.
- ...

nspark:
- Remove background color
- Add frame to examples

Closes #397

### DIFF
--- a/utils/defs.tex
+++ b/utils/defs.tex
@@ -153,6 +153,8 @@
 
 \newcommand{\ListingsParamStyle}{\bfseries\itshape\color{CadetBlue}}
 
+\definecolor{DarkGreen}{rgb}{0, 0.25, 0}
+
 \lstdefinestyle{SourceListingsDefault}{
   language={C},
   breaklines=true,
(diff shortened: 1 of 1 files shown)


BryantLam:
Would it be worthwhile to make the green and blue coloring darker and/or bolder? The text is readable as is, but it would stand out more if the coloring was more pronounced like in the purple for `TYPENAME` and `TYPE`.

nspark:
Here's an example for which the colors were made darker.

Old:
ColorsOld

New:
ColorsNew

BryantLam:
Looks great!

jamesaross:
Since this is more of an aesthetics thing, it should probably be discussed at the next meeting.  I prefer the grey background, although lighter than in rc1 (like 97%) and a grey border rather than black.  The github color scheme is about right to me.  The border is necessary to identify code examples that cross page breaks. I also prefer that the API blocks be the lighter grey, but without a border in order to distinguish them from the code examples.

nspark:
@manjugv Could we have some time near the end of Friday's RCM to discuss this?

manjugv:
> @manjugv Could we have some time near the end of Friday's RCM to discuss this?

👍 I have added this to the [agenda](https://github.com/openshmem-org/specification/wiki/March-20th#miscellaneous).

jdinan:
Replaced by https://github.com/jdinan/specification/pull/16

PR #399, opened 2020-03-18
Improve use of "non" vs. "non-"

### LINKED ISSUE #166: nonblocking vs. non-blocking
Comments from RCM meeting for synchronization section (#125) indicated [we should prefer "nonblocking"](https://github.com/openshmem-org/specification/pull/125#discussion_r143017597), but that preference is not consistent with actual use of "non-blocking".

| Instances | nonblocking | non-blocking |
| -- | -- | -- |
| OpenSHMEM 1.3 | 5 | 7 |
| OpenSHMEM 1.4rc1 | 8 | 18 |
* Generated via PDF search

I personally prefer "non-blocking", therefore backmatter has been using that spelling without previous guidance.

nspark:
- [X] Remove hyphen from  non-{critical, existent, profit, standard, zero}
  - Rationale: All these are dictionary words without the hyphen.
- [X] Reword to avoid non-initializing
  - Rationale: Don't "solve" a problem you can avoid entirely.
- [ ] Fix "non-symmetric"
  - Plan: Reword use to avoid entirely

Closes #166


jdinan:
Replaced by https://github.com/jdinan/specification/pull/17

PR #410, opened 2020-03-31
Correct sets and optionals

BryantLam:
Change entries in the deprecation table to use `{ ... }` sets instead of `[ ... ]` optionals.


BryantLam:
Wrong base branch. :(

PR #420, opened 2020-04-27
OpenSHMEM 1.5 Doc Edits

jdinan:


nspark:
Notes on use of the em dash (—):

- [Grammar Girl](https://www.quickanddirtytips.com/education/grammar/em-dash-space):

  > Don't put a space between an em-dash and the adjacent words.
  > 
  > Every style guide I checked, except the AP Stylebook, stated there should be no spaces between an em-dash and the adjacent words. That means it is a style choice. If you're writing for a newspaper, magazine, or website that uses Associated Press style, put in the spaces. Otherwise, leave them out. 

- [The Punctuation Guide](https://www.thepunctuationguide.com/em-dash.html)

  > The em dash is typically used without spaces on either side, and that is the style used in this guide. Most newspapers, however, set the em dash off with a single space on each side.

- [Grammarly](https://www.grammarly.com/blog/why-you-should-love-the-em-dash/)

  > Putting spaces before and after an em dash is a matter of preference; just be consistent.

jdinan:
@nspark Until recently, I used two spaces after a period, so I am happy to follow your lead and remove the extra spaces that I added. :) There is [one instance in backmatter.tex](https://github.com/openshmem-org/specification/blob/master/content/backmatter.tex#L615) that I didn't change. Could you take a look at that one?

content/backmatter.tex
```diff
@@ -672,98 +705,109 @@ \section{Version 1.5}
 \item Added team-based communication-management functions:
   \FUNC{shmem\_team\_create\_ctx} and
```
jdinan, line 706:
```suggestion
  \FUNC{shmem\_team\_create\_ctx} and \\
```

content/backmatter.tex
```diff
@@ -672,98 +705,109 @@ \section{Version 1.5}
...
+  subsec:shmem_wait_until_any,
+  subsec:shmem_wait_until_some,
+  subsec:shmem_test_all,
+  subsec:shmem_test_any,
+  subsec:shmem_test_some}%
 %
 \item Added support for vectorized comparison values in the multiple-element
   point-to-point synchronization API with the functions:
   \FUNC{shmem\_wait\_until\_all\_vector}, \FUNC{shmem\_wait\_until\_any\_vector},
   \FUNC{shmem\_wait\_until\_some\_vector}, \\
```
jdinan, line 775:
```suggestion
  \FUNC{shmem\_wait\_until\_some\_vector},
```

content/backmatter.tex
```diff
@@ -829,72 +873,72 @@ \section{Version 1.4}
...
 %
 \item Clarified API description for \FUNC{shmem\_info\_get\_name}.
-\\See Section \ref{subsec:shmem_info_get_name}.
+\ChangelogRef{subsec:shmem_info_get_name}%
 %
 \item Expanded the type support for RMA, AMO, and point-to-point
     synchronization operations.
-\\ See Tables \ref{stdrmatypes}, \ref{stdamotypes}, \ref{extamotypes}, and
-    \ref{p2psynctypes}
+\ChangelogRef{stdrmatypes,, stdamotypes,, extamotypes,, p2psynctypes}%
```
jdinan, line 922: Why `,,`?

BryantLam: These are Table references. By default, cleveref will sort and compress the references, but you can use `,,` to separate the list items to tell cleveref to *not* compress. The effect will be something akin to:
`See Tables 1, 2, 3 and 4.`
instead of
`See Tables 1 to 4.`
I could have done it either way, but since I felt that the Table hyperref links were useful, I didn't want to compress them away.

BryantLam: I will add a note here as a reminder.

content/backmatter.tex
BryantLam: Do these `\FUNC` bleed into the margin? If so, the same also happens with "Added support for a multiple-element point-to-point synchronization API with the functions". I would prefer to not adjust the line breaks on this one.

content/threads_intro.tex
```diff
@@ -21,7 +21,11 @@
 interfaces.
 \end{description}
 
-\noindent The following semantics apply to the usage of these models:
+\sloppypar
+\noindent The thread level constants must have increasing integer values, i.e.
```
jdinan, line 25:
```suggestion
\noindent The thread level constants must have increasing integer values; i.e.,
```

content/backmatter.tex
BryantLam, line 922:
```suggestion
%% cleveref will compress a list of references by default. It is better to not                                                                                                                             
%% compress this list of *table* references because the clickable hyperref
%% links are useful. You can tell cleveref to not compress the LHS and RHS by
%% inserting an empty item between them; i.e., `,,`.
\ChangelogRef{stdrmatypes,, stdamotypes,, extamotypes,, p2psynctypes}%
```

jdinan: Thanks!

content/backmatter.tex
jdinan: I will double check before merging.

jdinan: Adding the line break here didn't change what landed on each line, but it fixed some bad whitespace that LaTeX had inserted to fully justify the line containing `shmem_team_create_ctx`.

Screen Shot 2020-04-28 at 4 10 11 PM

PR #425, opened 2020-04-27
Reduce barriall and sync figure alignements

naveen-rn:
Reduced the size of the figures for barrierall and sync - the size
is similar to all other figures and it fits well within the page
limit.

This is better than recreating the images and making some extensive changes.

Hope this is fine.


jdinan:
@naveen-rn Apologies for the push, I had made the changes locally and wanted to save you the time. Using `\linewidth` should get us the right width without any tweaking. Please review and let me know what you think.

naveen-rn:
Looks fine with me.

PR #428, opened 2020-04-28
Add permission to copy statement

jdinan:
We reviewed the proposed copyright and license updates at the April 27 RCM and concluded that more time is needed to adopt these changes. In the meantime, we felt it would be appropriate to adopt this permission to copy statement that is used by MPI.


manjugv:
Requesting Steve's review. @swpoole

PR #439, opened 2020-06-09
OpenSHMEM 1.5 Doc Edits

jdinan:


jdinan:
Need two committee members to review and approve. Suggest reviewing individual commits to see the rationale for each set of changes.

Issue #365, opened 2020-02-07
Improve Strided API


jdinan:
## Issue

The current interleaved communication routines in OpenSHMEM (`shmem_iput/iget`) transfer single element chunks that are a fixed stride apart (source and destination can have different strides). This is inefficient for applications that wish to transfer larger array sections.

### Block Interleaved API
Extend the existing SHMEM interleaved APIs (e.g. `shmem_iput`) to include a block size:
```
void shmem_ibput(TYPE *dest, const TYPE *source, ptrdiff_t dst, ptrdiff_t sst,
                 size_t blocksize, size_t nelems, int pe);
```

### Strided APIs
Something similar to the ARMCI strided APIs could be used.
```
int ARMCI_PutS(void *src_ptr, int src_stride_ar[/*stride_levels*/],
               void *dst_ptr, int dst_stride_ar[/*stride_levels*/], 
               int count[/*stride_levels+1*/], int stride_levels, int proc);
```

### Subarray APIs
Similar to the MPI subarray datatype:
```
void shmem_TYPE_put/get[_warp](
        shmem_ctx_t ctx,
        TYPE *dest, size_t dest_ndim, size_t dest_dims[],
        size_t dest_ul_dim[], size_t dest_lr_dim[],
        TYPE *src, size_t src_ndim, size_t src_dims[],
        size_t src_ul_dim[], size_t src_lr_dim[],
        int pe);
```
The user specifies the full dimensions of the source and destination matrices and indicates the pointer to the zero'th element. The upper left and lower right indices of the source and destination slices are given to indicate the source/dest buffer.

This API has the advantage of being very easy for users to use (versus strided APIs, which require thinking about the linearization of the matrix). However, because data can be reshaped during the transfer, it also requires more work on the part of implementations.

naveen-rn:
Related:
[Extending Strided Communication Interfaces in OpenSHMEM](https://www.researchgate.net/publication/297734932_Extending_the_Strided_Communication_Interface_in_OpenSHMEM)
[Towards Matrix Oriented Strides in OpenSHMEM](https://dl.acm.org/doi/10.1145/2676870.2676888)

jdinan:
From 7/2/2020 meeting, WG prefers the block interleaved API.

jeffhammond:
Regarding the more general strided APIs...

From https://github.com/jeffhammond/oshmpi/blob/master/docs/oug2014_resubmission-acm_4.pdf:

[...]

PR #448, opened 2020-07-23
Add Block-Strided RMA Operations

jdinan:
Extend the SHMEM strided APIs (e.g. `shmem_iput`) to include a block size, e.g.:

```
void shmem_ibput(TYPE *dest, const TYPE *source, ptrdiff_t dst, ptrdiff_t sst,
                 size_t bsize, size_t nblocks, int pe);
```

See #365 for details.

### DIFF
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -653,6 +653,17 @@ standard \ac{AMO} types in Table~\ref{stdamotypes}.
 
 \chapter{Changes to this Document}\label{sec:changelog}
 
+\section{Version 1.6}
+Major changes in \openshmem[1.6] include \dots
+
+\begin{itemize}
+%
+\item Added interleaved block transfer APIs \FUNC{shmem\_ibget} and
+    \FUNC{shmem\_ibput}.
+\ChangelogRef{subsec:shmem_ibput, subsec:shmem_ibget}%
+%
+\end{itemize}
+
 \section{Version 1.5}
 Major changes in \openshmem[1.5] include the addition of new team-based
 collective functions, \OPR{put-with-signal} functions, nonblocking \ac{AMO}
--- a/content/shmem_ibget.tex
+++ b/content/shmem_ibget.tex
@@ -56,4 +56,10 @@ where \SIZE{} is one of \CONST{8, 16, 32, 64, 128}.
     None.
 }
 
+\apinotes{
+    The \FUNC{shmem\_ibget} API provides a more general purpose interleaved
+    transfer API than \FUNC{shmem\_iget}. Calling \FUNC{shmem\_ibget} with a
+    block size of 1 is equivalent to the \FUNC{shmem\_iget} API.
+}
+
 \end{apidefinition}
--- a/content/shmem_ibget.tex
+++ b/content/shmem_ibget.tex
@@ -30,18 +30,18 @@ where \SIZE{} is one of \CONST{8, 16, 32, 64, 128}.
         The type of \dest{} should match that implied in the SYNOPSIS section.}
     \apiargument{IN}{source}{Symmetric address of the source array data object.
         The type of \source{} should match that implied in the SYNOPSIS section.}
-    \apiargument{IN}{dst}{The stride between consecutive elements of the \dest{}
-        array. The stride must be greater than or equal to \CONST{1} and is
-        scaled by the element size of the \dest{} array. A value of \CONST{1}
+    \apiargument{IN}{dst}{The stride between consecutive blocks of the \dest{}
+        array. The stride must be greater than or equal to \VAR{bsize} and is
+        scaled by the element size of the \dest{} array. A value of \VAR{bsize}
         indicates contiguous data.}
-    \apiargument{IN}{sst}{The stride between consecutive elements of the \source{}
-        array. The stride must be greater than or equal to \CONST{1} and is
-        scaled by the element size of the \source{} array. A value of \CONST{1}
+    \apiargument{IN}{sst}{The stride between consecutive blocks of the \source{}
+        array. The stride must be greater than or equal to \VAR{bsize} and is
+        scaled by the element size of the \source{} array. A value of \VAR{bsize}
         indicates contiguous data.}
     \apiargument{IN}{bsize}{Number of elements per block in the \dest{} and \source{}
         arrays.}
-    \apiargument{IN}{nblocks}{Number of blocks in the \dest{} and \source{}
-        arrays.}
+    \apiargument{IN}{nblocks}{Number of blocks to be copied from the \source{} array
+        to the \dest{} array.}
     \apiargument{IN}{pe}{\ac{PE} number of the remote \ac{PE}.}
 \end{apiarguments}
 
--- a/content/shmem_ibget.tex
+++ b/content/shmem_ibget.tex
@@ -1,0 +1,59 @@
+\apisummary{
+    Copies blocks of strided data from a specified \ac{PE}.
+}
+
+\begin{apidefinition}
+
+\begin{C11synopsis}
+void @\FuncDecl{shmem\_ibget}@(TYPE *dest, const TYPE *source, ptrdiff_t dst, ptrdiff_t sst, size_t bsize, size_t nblocks, int pe);
+void @\FuncDecl{shmem\_ibget}@(shmem_ctx_t ctx, TYPE *dest, const TYPE *source, ptrdiff_t dst, ptrdiff_t sst, size_t bsize, size_t nblocks, int pe);
+\end{C11synopsis}
+where \TYPE{} is one of the standard \ac{RMA} types specified by Table \ref{stdrmatypes}.
+
+\begin{Csynopsis}
+void @\FuncDecl{shmem\_\FuncParam{TYPENAME}\_ibget}@(TYPE *dest, const TYPE *source, ptrdiff_t dst, ptrdiff_t sst, size_t bsize, size_t nblocks, int pe);
+void @\FuncDecl{shmem\_ctx\_\FuncParam{TYPENAME}\_ibget}@(shmem_ctx_t ctx, TYPE *dest, const TYPE *source, ptrdiff_t dst, ptrdiff_t sst, size_t bsize, size_t nblocks, int pe);
+\end{Csynopsis}
+where \TYPE{} is one of the standard \ac{RMA} types and has a corresponding \TYPENAME{} specified by Table \ref{stdrmatypes}.
+
+\begin{CsynopsisCol}
+void @\FuncDecl{shmem\_ibget\FuncParam{SIZE}}@(void *dest, const void *source, ptrdiff_t dst, ptrdiff_t sst, size_t bsize, size_t nblocks, int pe);
+void @\FuncDecl{shmem\_ctx\_ibget\FuncParam{SIZE}}@(shmem_ctx_t ctx, void *dest, const void *source, ptrdiff_t dst, ptrdiff_t sst, size_t bsize, size_t nblocks, int pe);
+\end{CsynopsisCol}
+where \SIZE{} is one of \CONST{8, 16, 32, 64, 128}.
+
+\begin{apiarguments}
+    \apiargument{IN}{ctx}{A context handle specifying the context on which to perform the operation.
+        When this argument is not provided, the operation is performed on
+        the default context.}
+    \apiargument{OUT}{dest}{Local address of the array to be updated.
+        The type of \dest{} should match that implied in the SYNOPSIS section.}
+    \apiargument{IN}{source}{Symmetric address of the source array data object.
+        The type of \source{} should match that implied in the SYNOPSIS section.}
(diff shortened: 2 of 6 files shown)


jamesaross:
The proposed block-strided RMA interfaces _ibput_/_ibget_  are close to being useful to support movement of arbitrary data structures.  They're close enough that I would expect the interfaces would be abused in order to make it work by padding structs and casting pointers to an appropriate data type.  So why not support a single interface for arbitrary data structures and block sizes and drop the typed interfaces?

The proposal requires _dst_ and _sst_ strides to have a value greater than or equal to 1, not just for the new block-strided RMA operations, but for the existing strided interfaces.  It appears this requirement was placed on the previous strided interfaces _iput_ but not _iget_ (the current specification seems to be inconsistent here).  Having negative strides enabled applications to perform things like inverting data order.  What is the reason for requiring the stride to be >= 1?  Do we deprecate the current _iget_ strided interface in this case because this proposal reduces functionality?  I believe the >= 1 stride requirement isn't necessary for any of the interfaces.

jdinan:
@jamesaross Appreciate your feedback on the proposal. Interested to hear your thoughts on the other options under discussion in #365. The working group felt block-strided would be an obvious extension of what we already have, so we decided to bring it forward to start a discussion.

It looks like OpenSHMEM added the stride greater or equal to 1 restriction in OpenSHMEM 1.1.

jamesaross:
@jdinan The possible solution of _shmem_iputmem_/_shmem_igetmem_ in #365 appears to generalize the datatype so that it would support arbitrary data structures and all of the typed routines in this proposal.

There doesn't appear to be a requirement within for strides greater or equal to 1 for _shmem_iget_.  This is possibly an oversight, but adding it now would hobble prior capability.

davidozog:
FWIW, I do like the idea of adding a `shmem_ibputmem/ibgetmem`... it only adds 2 more symbols to the spec and quite a bit of flexibility.

I'm not so sure about supporting negative strides - `iget` looks like an oversight to me.  Is it useful?  I'm not sure I understand how it would work.  If the source array is {1, 2, 3, 4, ..., 10}, and you ask for sst=-2, would it it send {1, 9, 7, 5, 3} or something else like {10, 8, 6, 4, 2}?  Seems a little strange, but maybe there's a use case I'm missing.

In SOS, it looks like we assume `dst` and `sst` are positive and return an error if not (if the user does "error checking", but the operation goes out of bounds otherwise...).

jamesaross:
@davidozog 

In the case of sst=-2, you would need to send it an appropriate source address so the operation could go out of bounds (i.e. `&source[9]`).  I don't think wrapping around the data is correct since the function doesn't know the length of the array.  Going out of bounds is a potential issue even without negative strides.

I'm not strongly attached to the need for negative strides, but it removes error checking from library code, makes it an application/user issue, and potentially provides a capability to minimize extra data manipulation in some cases.

So I think we're looking to add `ibputmem/ibgetmem`.

content/shmem_ibput.tex
```diff
@@ -0,0 +1,62 @@
...
+    \apiargument{IN}{bsize}{Number of elements per block in the \dest{} and \source{}
+        arrays.}
+    \apiargument{IN}{nblocks}{Number of blocks in the \dest{} and \source{}
+        arrays.}
+    \apiargument{IN}{pe}{\ac{PE} number of the remote \ac{PE}.}
+\end{apiarguments}
+
+
+\apidescription{
+    The \FUNC{shmem\_ibput} routines provide a method  for  copying strided data
```
jdinan, line 50: First sentence is awkward (ibput and ibget). Updated here and in iput/iget.

content/shmem_ibput.tex
```diff
@@ -0,0 +1,62 @@
...
+        The type of \source{} should match that implied in the SYNOPSIS section.}
+    \apiargument{IN}{dst}{The stride between consecutive elements of the \dest{}
+        array. The stride must be greater than or equal to \CONST{1} and is
+        scaled by the element size of the \dest{} array. A value of \CONST{1}
+        indicates contiguous data.}
+    \apiargument{IN}{sst}{The stride between consecutive elements of the \source{}
+        array. The stride must be greater than or equal to \CONST{1} and is
+        scaled by the element size of the \source{} array. A value of \CONST{1}
+        indicates contiguous data.}
+    \apiargument{IN}{bsize}{Number of elements per block in the \dest{} and \source{}
```
jdinan, line 41: @minsii Suggested bsize -> nelem

jdinan: iput/iget already have an nelems argument

content/shmem_iget.tex
```diff
@@ -1,5 +1,5 @@
 \apisummary{
-    Copies strided data from a specified \ac{PE}.
+    Copies strided data blocks from a specified \ac{PE}.
```
jdinan, line 2:
```suggestion
    Copies strided data from a specified \ac{PE}.
```

jdinan:
Name suggestion from @BryantLam `shmem_iblock_put/get`.

jdinan:
@swpoole To provide feedback on usage models in some apps.

jeffhammond:
I hear dense matrix computations are an application domain of interest in HPC 😄 
[oug2014slides.pdf](https://github.com/openshmem-org/specification/files/5219339/oug2014slides.pdf)

jdinan:
@yfguo Please review.

jdinan:
@davidozog Pointed out that we don't have nonblocking versions of the proposed API and are also missing nonblocking versions of the existing interleaved APIs.

jdinan:
This was discussed at the 11/29/21 RMA WG meeting and we decided to move forward with the block-interleaved proposal, while considering a future subarray API if need arises.

jdinan:
Scheduling this proposal for a re-reading at the December 2, 2022 meeting as it's been two years since the last reading.

content/shmem_ibget.tex
```diff
@@ -0,0 +1,59 @@
...
+    \apiargument{IN}{source}{Symmetric address of the source array data object.
+        The type of \source{} should match that implied in the SYNOPSIS section.}
+    \apiargument{IN}{dst}{The stride between consecutive elements of the \dest{}
+        array. The stride must be greater than or equal to \CONST{1} and is
+        scaled by the element size of the \dest{} array. A value of \CONST{1}
+        indicates contiguous data.}
+    \apiargument{IN}{sst}{The stride between consecutive elements of the \source{}
+        array. The stride must be greater than or equal to \CONST{1} and is
+        scaled by the element size of the \source{} array. A value of \CONST{1}
+        indicates contiguous data.}
```
jdinan, line 40:
```suggestion
    \apiargument{IN}{sst}{The stride between consecutive blocks of the \source{}
        array. The stride must be greater than or equal to \CONST{bsize} and is
        scaled by the element size of the \source{} array. A value of \CONST{bsize}
        indicates contiguous data.}
```

content/shmem_ibget.tex
```diff
@@ -0,0 +1,59 @@
...
+        When this argument is not provided, the operation is performed on
+        the default context.}
+    \apiargument{OUT}{dest}{Local address of the array to be updated.
+        The type of \dest{} should match that implied in the SYNOPSIS section.}
+    \apiargument{IN}{source}{Symmetric address of the source array data object.
+        The type of \source{} should match that implied in the SYNOPSIS section.}
+    \apiargument{IN}{dst}{The stride between consecutive elements of the \dest{}
+        array. The stride must be greater than or equal to \CONST{1} and is
+        scaled by the element size of the \dest{} array. A value of \CONST{1}
+        indicates contiguous data.}
```
jdinan, line 36:
```suggestion
    \apiargument{IN}{dst}{The stride between consecutive blocks of the \dest{}
        array. The stride must be greater than or equal to \CONST{bsize} and is
        scaled by the element size of the \dest{} array. A value of \CONST{bsize}
        indicates contiguous data.}
```

content/shmem_ibget.tex
```diff
@@ -0,0 +1,59 @@
...
+    The \FUNC{shmem\_ibget} routines provide a method for copying strided data blocks from
+    a symmetric array from a specified remote \ac{PE} to strided locations on a
+    local array.  The routines return when the data has been copied into the local
+    \VAR{dest} array.
+}
+
+\apireturnvalues{
+    None.
+}
+
```
jdinan, line 58: \apinotes{
    The \FUNC{shmem_ibget} API provides a more general purpose interleaved transfer API than \FUNC{shmem_iget}. Calling this routine with a block size of 1 is equivalent to the \FUNC{shmem_iget} API.
}

content/shmem_ibput.tex
```diff
@@ -0,0 +1,62 @@
...
+        When this argument is not provided, the operation is performed on
+        the default context.}
+    \apiargument{OUT}{dest}{Symmetric address of the destination array data object.
+        The type of \dest{} should match that implied in the SYNOPSIS section.}
+    \apiargument{IN}{source}{Local address of the array containing the data to be copied.
+        The type of \source{} should match that implied in the SYNOPSIS section.}
+    \apiargument{IN}{dst}{The stride between consecutive elements of the \dest{}
+        array. The stride must be greater than or equal to \CONST{1} and is
+        scaled by the element size of the \dest{} array. A value of \CONST{1}
+        indicates contiguous data.}
```
jdinan, line 36:
```suggestion
    \apiargument{IN}{dst}{The stride between consecutive blocks of the \dest{}
        array. The stride must be greater than or equal to \ARG{bsize} and is
        scaled by the element size of the \dest{} array. A value of \ARG{bsize}
        indicates contiguous data.}
```

content/shmem_ibput.tex
```diff
@@ -0,0 +1,62 @@
...
+    \apiargument{IN}{sst}{The stride between consecutive elements of the \source{}
+        array. The stride must be greater than or equal to \CONST{1} and is
+        scaled by the element size of the \source{} array. A value of \CONST{1}
+        indicates contiguous data.}
```
jdinan, line 40:
```suggestion
    \apiargument{IN}{sst}{The stride between consecutive blocks of the \source{}
        array. The stride must be greater than or equal to \CONST{bsize} and is
        scaled by the element size of the \source{} array. A value of \CONST{bsize}
        indicates contiguous data.}
```

content/shmem_ibput.tex
```diff
@@ -0,0 +1,62 @@
...
+    \apiargument{IN}{bsize}{Number of elements per block in the \dest{} and \source{}
+        arrays.}
+    \apiargument{IN}{nblocks}{Number of blocks in the \dest{} and \source{}
+        arrays.}
```
jdinan, line 44:
```suggestion
    \apiargument{IN}{nblocks}{Number of blocks to be copied from the \source{} array to the \dest{}
        array.}
```

content/shmem_ibput.tex
```diff
@@ -0,0 +1,62 @@
...
+    \apiargument{IN}{pe}{\ac{PE} number of the remote \ac{PE}.}
+\end{apiarguments}
+
+
+\apidescription{
+    The \FUNC{shmem\_ibput} routines provide a method  for  copying strided data
+    elements (specified by \VAR{sst}) of an array from a \source{} array on the
```
jdinan, line 51:
```suggestion
    blocks (specified by \VAR{sst}) of an array from a \source{} array on the
```

content/shmem_ibget.tex
```diff
@@ -0,0 +1,59 @@
...
+        scaled by the element size of the \dest{} array. A value of \CONST{1}
+        indicates contiguous data.}
+    \apiargument{IN}{sst}{The stride between consecutive elements of the \source{}
+        array. The stride must be greater than or equal to \CONST{1} and is
+        scaled by the element size of the \source{} array. A value of \CONST{1}
+        indicates contiguous data.}
+    \apiargument{IN}{bsize}{Number of elements per block in the \dest{} and \source{}
+        arrays.}
+    \apiargument{IN}{nblocks}{Number of blocks in the \dest{} and \source{}
+        arrays.}
```
jdinan, line 44:
```suggestion
    \apiargument{IN}{nblocks}{Number of blocks to be copied from the \source{} array
        to the \dest{} array.}
```

content/shmem_ibget.tex
jdinan: Resolved by 5714500b

jdinan:
Special Ballot: https://github.com/openshmem-org/specification/pull/448/files/0d78753cb887cb8edcc5fa0412bbfcc7a8cd9b4a..5714500b808c27058c5dd93066aa0460711d3055

content/backmatter.tex
```diff
@@ -686,6 +691,7 @@ \section{Version 1.6}
     2 to enable profiling with profile library defined effects and
     additional arguments.
 \ChangelogRef{subsec:shmem_pcontrol}
+>>>>>>> master
```
manjugv, line 694: @jdinan Please fix it!

PR #452, opened 2020-08-25
Clarifying Team Collectives

jamesaross:
Some descriptions of collective routines did not specify support for OpenSHMEM teams, but rather only active sets.  These changes clarify that they support both teams and active sets (now deprecated).

### DIFF
--- a/content/shmem_reductions.tex
+++ b/content/shmem_reductions.tex
@@ -1,6 +1,6 @@
 \apisummary{
     The following functions perform reduction operations across all
-    \acp{PE} in a set of \acp{PE}.
+    \acp{PE} in an \openshmem team or active set of \acp{PE}.
 }
 
 \begin{apidefinition}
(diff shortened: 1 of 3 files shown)


content/shmem_reductions.tex
```diff
@@ -1,6 +1,6 @@
 \apisummary{
     The following functions perform reduction operations across all
-    \acp{PE} in a set of \acp{PE}.
+    \acp{PE} in an \openshmem team or active set of \acp{PE}.
```
jdinan, line 3: Would the sentence read better if we dropped "of \acp{PE}"?

jdinan: Also suggest dropping "of PEs" in the subsequent occurrences.

davidozog:
Most all these changes are apparently now covered by #511, but perhaps we should check with @kwaters4 before closing this PR.

@kwaters4 - the changes on this PR in `shmem_reductions.tex` are not in #511, but I think "a set of PEs" might be ok in that context... Otherwise, it looks like #511 can supplants the changes here, but please correct me if I'm wrong.

davidozog:
Closed by https://github.com/davidozog/openshmem-specification/pull/8

Issue #450, opened 2020-08-18
Behavior of non-parent PEs entering team split routines


davidozog:
Should we clarify the behavior of non-parent PEs that enter a team split routine, i.e when those PEs pass `parent_team=SHMEM_TEAM_INVALID`?

In other words, does this existing text (in `shmem_team_split_strided` and `shmem_team_split_2d`):
> This routine must be called by all PEs in the parent team.

imply this routine must _only_ be called by PEs in the parent team?  My guess is no, PEs that are not in the parent can pass `SHMEM_TEAM_INVALID`.  This semantic generated some confusion in an SOS unit test:

https://github.com/Sandia-OpenSHMEM/SOS/issues/969

and I'm thinking something like this (but better;) might help prevent confusion:
> This routine must be called by all PEs in the parent team, and any other PE calling the routine must set _parent_team_ to SHMEM_TEAM_INVALID.

There's also a slight bit of confusion with this existing text:
> If the new team cannot be created or an invalid PE triplet is provided, then _new_team_ will be assigned the value SHMEM_TEAM_INVALID and _shmem_team_split_strided_ will return a nonzero value on all PEs in the parent team.

in combination with:

> If parent_team compares equal to SHMEM_TEAM_INVALID, then no new team will be created and _new_team_ will be assigned the value SHMEM_TEAM_INVALID [...]   **Return Values** Zero on successful creation of new_team; otherwise, nonzero.

So it seems to be implied that PEs passing `SHMEM_TEAM_INVALID` should also return non-zero, but this may not be explicitly stated, unless I'm missing something.  I might suggest:
> If parent_team compares equal to SHMEM_TEAM_INVALID, then no new team will be created, _new_team_ will be assigned the value SHMEM_TEAM_INVALID, and the routine will return nonzero.

Sorry, lots of words to describe  a pretty minor problem... :wink:

jdinan:
You can have multiple simultaneous split operations, for example consider two splits where parents teams are even PEs and odd PEs. There is no semantic relationship between these splits even though they can be happening simultaneously. I think the same is true of a split involving SHMEM_TEAM_INVALID. There's no relationship to any other split that's happening. Given that it's impossible for a PE to have a handle for a team is does not belong to, I'm not sure there is anything to define for PEs outside of a given team. Whatever parent team argument they pass will define the requirements of that call. 

Since a split on SHMEM_TEAM_INVALID does not create a team, our existing semantics imply that the return value should be nonzero.

[...]

PR #461, opened 2021-02-03
Clarify return value for team split on SHMEM_TEAM_INVALID

davidozog:
This PR is meant to address #450 (as well as https://github.com/Sandia-OpenSHMEM/SOS/issues/969 and https://github.com/Sandia-OpenSHMEM/SOS/issues/972).

I don't believe https://github.com/openshmem-org/specification/commit/8b44a29f6075a666aa3fcb3a9d123d1e808c0ec8 is a semantic change - just a clarification that a split operation on a SHMEM_TEAM_INVALID parent is indeed considered "unsuccessful" and the routine should return a nonzero value.

I also included a small change to the 2D team split example code (https://github.com/openshmem-org/specification/commit/ec66ff256fe78c9dba04afdb90b23bd02e66f22b).  This has no effect on the example, but leaving these variables uninitialized commonly causes a compiler warning.  Please let me know if you prefer this in a separate PR...

### DIFF
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -664,6 +664,10 @@ The following list describes the specific changes in \openshmem[1.6]:
   \FUNC{shmem\_team\_ptr}.
 \ChangelogRef{subsec:shmem_team_ptr}%
 %
+\item Clarified that \FUNC{shmem\_team\_split\_strided} and
+    \FUNC{shmem\_team\_split\_strided} return a nonzero value when the parent
+    team compares equal to \LibConstRef{SHMEM\_TEAM\_INVALID}.
+\ChangelogRef{subsec:shmem_team_split_strided, subsec:shmem_team_split_2d}%
 \end{itemize}
 
 \section{Version 1.5}
--- a/content/shmem_team_split_2d.tex
+++ b/content/shmem_team_split_2d.tex
@@ -95,9 +95,10 @@ See Section~\ref{subsec:shmem_team_config_t} for field mask names and
 default configuration parameters.
 
 If \VAR{parent\_team} compares equal to
-\LibConstRef{SHMEM\_TEAM\_INVALID}, then no new teams will be created
-and both \VAR{xaxis\_team} and \VAR{yaxis\_team} will be assigned the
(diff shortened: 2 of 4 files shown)


davidozog:
@jdinan and @manjugv - should this ticket have an official reading?  I think it's not needed, but am happy to do it.

I also just realized the https://github.com/openshmem-org/specification/commit/ec66ff256fe78c9dba04afdb90b23bd02e66f22b commit is already approved in #449 but was never merged.

jdinan:
@davidozog Thanks for keeping track of this ticket. This feels more significant than a doc edit to me. Given that we have the time, I would prefer to give the committee a chance to weigh in on whether the case that's being clarified is one where the function should return nonzero.

davidozog:
@jdinan - sure.  Maybe I can do an unofficial reading at the upcoming meeting to see if it generates any concerns.  It's too late to announce an official reading this time...

davidozog:
Note - this should go into errata (and/or pending errata?)

Also need a changelog entry.

davidozog:
@jdinan - I was about to cherry-pick 503317b96356cca794538516d604b526bb026aea from @kholland-intel's PR, which adds a v1.6 change-log that mentions `shmem_team_ptr`.  Is that ok, or would you prefer a separate change-log commit on this PR that might conflict later?

jdinan:
@davidozog Please go ahead and add a changelog entry to the effect of "Clarified the return value of team split operations when the parent team is SHMEM_TEAM_INVALID". In the future, I'd prefer to include the changelog entry with the reading since it will help with consistency.

Please don't cherry pick from another PR. I'll deal with any merge conflicts as they come up.

jdinan:
@davidozog The changelog heading in question was merged into master. Please merge from master and add your changelog entry.

jdinan:
@davidozog Could you please review the changelog entry in 75559f8?

davidozog:
@jdinan - 75559f8 looks great, and thanks for merging this with master.

davidozog:
@jdinan - I believe this PR is ready to merge.  Is there anything pending on your end?

(This should close https://github.com/Sandia-OpenSHMEM/SOS/issues/969)

PR #466, opened 2021-05-04
Fix inconsistency in return value for shmem_test_all{_vector}

### LINKED ISSUE #442: Contradicting test_all semantics
In the API description of ```shmem_test_all```, we have the following statement:
> If all elements in status are nonzero or nelems is 0, the test set is empty and this routine returns 0.

And in the return values - we have the following:
> shmem_test_all returns 1 if all variables in ivars satisfy the test condition or if nelems is 0, otherwise this routine returns 0.

So, when the nelems is 0 - should we return 0 or 1?

nspark:
Fixes inconsistency in return value for `shmem_test_all{_vector}`

Closes #442

### DIFF
--- a/content/shmem_test_all.tex
+++ b/content/shmem_test_all.tex
@@ -52,7 +52,7 @@ corresponding \TYPENAME{} specified by Table \ref{stdamotypes}.
     the element is excluded from the test set.  Elements of \VAR{status} set to
     0 will be included in the test set, and elements set to a nonzero value will be ignored.  If all elements
     in \VAR{status} are nonzero or \VAR{nelems} is 0, the test set is empty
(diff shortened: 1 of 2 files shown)


naveen-rn (approved):

davidozog (approved):

nspark:
@jdinan Does this count as a doc edit?

Also, I tried to run the GitHub CI job, but it wouldn't go through. I'm willing to bet whatever I set up was a little brittle.

jdinan:
@davidozog Will create a PR with the missing changelog entry for the erratum captured here: https://github.com/openshmem-org/specification/wiki/OpenSHMEM-1.5-Errata

PR #470, opened 2021-06-25
Add shmem_{initialized, finalized}

### LINKED ISSUE #457: Add API for querying if OpenSHMEM is currently initialized
Unless I'm missing it, there is currently no API for querying if an OpenSHMEM implementation has been initialized yet (i.e. whether we are in the part of a program between shmem_init and shmem_finalize). Theoretically, this could be something like:

```
bool shmem_is_initialized();
```

In general this seems like a useful API, and I've just run in to a concrete case where I "need" it. I'm writing a library in C++ that uses OpenSHMEM. Stack-allocated objects can collectively allocate and deallocate symmetric objects. If a stack-allocated object is created in main(), C++ will not call its destructor until we exit main() which is likely after the programmer has already called shmem_finalize() and cleaned up the symmetric heap. My library calling shmem_free in the destructor of stack-allocated objects after shmem_finalize then causes the app to crash. I can workaround this in a couple of ways: (1) asking users to artificially introduce a new scope inside of main using {} so that C++ objects go out of scope before shmem_finalize, or (2) adding initialize/finalize routines to my library that in turn call shmem_init/finalize and then track that state myself. I'm currently doing (1) and I'd rather avoid (2) as the user may want to use SHMEM before/after my library is in use.

Question to implementors: is this something difficult to add? Is there any reason why this API doesn't already exist? Is there an equivalent in the spec already that I'm missing?

nspark:
This PR adds `shmem_initialized` and `shmem_finalized` to query the state of the library.

Closes #457

### DIFF
--- a/content/shmem_finalized.tex
+++ b/content/shmem_finalized.tex
@@ -1,0 +1,32 @@
+\apisummary{
+  Indicates whether the \openshmem library has been finalized.
+}
+
+\begin{apidefinition}
+
+\begin{Csynopsis}
+int @\FuncDecl{shmem\_finalized}@(void);
+\end{Csynopsis}
+
+\begin{apiarguments}
+  \apiargument{None}{}{}
+\end{apiarguments}
+
+\apidescription{
+  The \FUNC{shmem\_finalized} routine returns a value indicating
+  whether the \openshmem library has been finalized (i.e, a call to
+  \FUNC{shmem\_finalize} has completed successfully).
+  This routine may be called at any point in an \openshmem program,
+  including before \FUNC{shmem\_init[\_thread]} and after
+  \FUNC{shmem\_finalize}.
+  This routine may be called by any thread of execution in the
+  \ac{PE}, independent of the level of thread support provided by the
+  implementation.
+}
+
+\apireturnvalues{
+  Returns \CONST{1} if the \openshmem library has been finalized;
+  otherwise, returns \CONST{0}.
+}
+
+\end{apidefinition}
--- a/content/shmem_initialized.tex
+++ b/content/shmem_initialized.tex
@@ -1,0 +1,33 @@
+\apisummary{
+  Indicates whether the \openshmem library has been initialized.
+}
+
+\begin{apidefinition}
+
+\begin{Csynopsis}
+int @\FuncDecl{shmem\_initialized}@(void);
+\end{Csynopsis}
+
+\begin{apiarguments}
+  \apiargument{None}{}{}
+\end{apiarguments}
+
+\apidescription{
+  The \FUNC{shmem\_initialized} routine returns a value indicating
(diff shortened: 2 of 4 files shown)


nspark:
Suggestion from 8/16 WG: Add note clarifying that `shmem_initialized() == 0` is not necessarily sufficient to prevent multiple threads from racing to call `shmem_init`.

jdinan:
If we follow the MPI semantics, the guard for calling `shmem_init` would be `!shmem_initialized() && !shmem_finalized()`.

nspark:
Added the following note:

> Although `shmem_initialized` is thread-safe, its return value is not a sufficient guard to prevent multiple threads from racing to initialize the OpenSHMEM library concurrently, as `shmem_initialized` may return 0 to one thread while library initialization is in progress due to a call from another thread. Applications must ensure that only one call to `shmem_init[_thread]` is made to initialize the OpenSHMEM library.

jdinan:
@nspark The note is reasonable, but why not add reference counted init/finalize and require that init functions are thread safe?

nspark:
@jdinan IDK, that just seemed like a change of a larger scope. I'm happy to draft it, but I'd probably do it as a separate PR.

content/shmem_initialized.tex
```diff
@@ -0,0 +1,44 @@
...
+int @\FuncDecl{shmem\_initialized}@(void);
+\end{Csynopsis}
+
+\begin{apiarguments}
+  \apiargument{None}{}{}
+\end{apiarguments}
+
+\apidescription{
+  The \FUNC{shmem\_initialized} routine returns a value indicating
+  whether the \openshmem library has been initialized (i.e, a call to
```
jdinan, line 17: Do we want "has been" or "is currently"?

content/shmem_finalized.tex
```diff
@@ -0,0 +1,32 @@
...
+int @\FuncDecl{shmem\_finalized}@(void);
+\end{Csynopsis}
+
+\begin{apiarguments}
+  \apiargument{None}{}{}
+\end{apiarguments}
+
+\apidescription{
+  The \FUNC{shmem\_finalized} routine returns a value indicating
+  whether the \openshmem library has been finalized (i.e, a call to
```
jdinan, line 17: Do we want "has been" or "is currently"? The problem with "has been" is that it tells you whether the API was ever called. If we assume that an OpenSHMEM library can't be reinitialized, then this is enough information to tell you the library state. If an OpenSHMEM library supports reinitialization then "has been" is not enough information to know the state of the library. You need to know "is currently".

jdinan: For reference, here is the NVSHMEM library state query function: https://docs.nvidia.com/hpc-sdk/nvshmem/api/docs/gen/api/setup.html#nvshmemx-init-status

jdinan:
Superseded by #512

PR #472, opened 2021-06-25
Fix GitHub Actions CI

nspark:
- Updates to v2 for actions
- Adds `ca-certificates` dependence ([see ref](https://github.com/actions/checkout/issues/417#issuecomment-858749032))


nspark:
Despite referenced comment, I'm not sure the addition of `ca-certificates` is necessary, since those dependencies get installed _after_ the repository is checked out.

nspark:
@davidozog I think this fixes the CI issues we were having

@jdinan Okay to merge?

nspark:
Well, this is odd. #470 failed on the first commit but not the second. Both ran on `checkout@v1`. (The first commit failed during the checkout step.) Maybe failures are occasional with v1, but v2 is more robust? I don't think it's _bad_ to update to v2.

PR #477, opened 2021-09-08
WIP: Memory Model

nspark:
This is a **_very_** WIP draft for changes to define a memory model. It is in no way complete and almost certainly incorrect in the bit it attempts to specify. But, I invite and welcome all your feedback (You can see the remains of many previous attempts as commented-out LaTeX.)

The style of wording is heavily based on C11/C17 and cppreference.com (e.g., [atomic_thread_fence](https://en.cppreference.com/w/cpp/atomic/atomic_thread_fence)).

Part of the approach here is to add the notion of _PE-happens-before_ and _PE-synchronizes-with_, rather than overload/extend C's definitions of _happens-before_ and _synchronizes-with_. What has so far made this a complicated endeavor (both in wording and correctness) is that OpenSHMEM has blocking and nonblocking operations, fence vs. quiet.

### DIFF
--- a/content/memory_model.tex
+++ b/content/memory_model.tex
@@ -3,7 +3,9 @@
 \caption{\openshmem Memory Model}
 \label{fig:mem_model}
 \end{figure}
-%
+
+\subsection{Object Storage Model}
+
 An \openshmem program consists of data objects that are private to each \ac{PE}
 and data  objects that are remotely accessible by all \acp{PE}. Private data
 objects are stored in the local memory of each \ac{PE} and can only be accessed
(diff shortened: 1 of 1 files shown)


nspark:
Most of what's currently drafted... (see bullet 7 for one challenging area)

openshmem-memory-model-draft-20210908

nspark:
Some feedback points from 10/4:

- `shmem_fence` is really just a release fence; it doesn't order nonblocking gets and fetching AMOs. A standalone acquire fence is likely necessary for completeness.
- `shmem_fence` is a release fence for both blocking and nonblocking operations.
- Examples or litmus tests are likely to be very helpful and illustrative.
- Likely needs a complete definition of transitivity for *PE-synchronizes-with*.

Issue #476, opened 2021-09-04
Spec 1.5: ambiguous profile level information


tonycurtis:
On p. 141, the profile levels and their effects are listed.

Level 2 has differing definitions/effects: == 2 and >= 2

wrrobin:
Thanks for catching that @tonycurtis. Will post a fix.

kwaters4:
@wrrobin Can this be closed with the PR #480?

PR #480, opened 2021-09-24
Fix for profiling level value of 2

wrrobin:
Related to Issue #476 
Signed-off-by: rahmanmd <md.rahman@intel.com>


wrrobin:
@jdinan I added a bullet in the [OpenSHMEM-1.5 errata](https://github.com/openshmem-org/specification/wiki/OpenSHMEM-1.5-Errata) page. Should we also need the changelog entry? Please advise.

jdinan:
Yes, as a rule of thumb the note should be included in both the errata and changelog. That is, we would want to call attention to the change since it is subtle and significant to users/implementors.

jdinan (approved):

Issue #189, opened 2017-12-13
Operation Sessions Extension


jdinan:
## Problem

The OpenSHMEM library sees individual operations as they occur.  If the application will perform a series of operations (e.g. a loop of puts), it can be beneficial to inform the runtime.

## Proposed Solution

Allow users to provide a performance hint that they will issue a set of operations.  This information can be used to optimize the behavior of the OpenSHMEM runtime.  For example, sessions may allow the runtime to aggregate operations or completions.

## Proposed API

Straw man API:
```
void shmem_ctx_start_session(shmem_ctx_t ctx);
void shmem_ctx_end_session(shmem_ctx_t ctx);
```

The start session operation informs the OpenSHMEM library that operations will be issued on the given context.  The end session operation informs the OpenSHMEM library that the last operation in the session has been issued on the given context.  Ending a session does not guarantee completion or ordering, quiet/fence on the context must still be used to achieve these guarantees.

davidozog:
Here are the WIP slides from the RMA WG (Min's) discussion today:
https://docs.google.com/presentation/d/1rwONNPFlHMv36eLJSWm8TnGE6nrEvVwpLARaIFH1RmY/edit?usp=sharing

Key takeaways/questions for next time:
1) Do we want bundling to be more "implicit" (the OpenSHMEM library decides how to bundle operations) or "explicit" (the user says exactly which ops are bundled and/or when they are shipped).
2) Should bundles be associated with contexts?
3) Should collectives be supported for bundling?

davidozog:
Here are the slides we covered in the committee meeting today:
https://docs.google.com/presentation/d/1Eco0_yFRS__b-kHy9dM7vO4yDw951_Gc685divy_JZs/edit?usp=sharing

I created a gist that attempts to capture the ideas that came up today.  Feedback is welcome:
https://gist.github.com/davidozog/9be1e52dd425567f86ccb1b050446baf

nspark:
Personally, I like Option 2 from your slides. You present:

```c
shmem_ctx_t my_ctx;
shmem_bundle_start(my_ctx);

shmem_put(my_ctx, ...);
shmem_atomic_inc(my_ctx, ...);
/* ... more RMA/AMO ... */

shmem_bundle_stop(my_ctx);
```

The `shmem_bundle*_t` types seem unnecessary at the current stage of design. I'm hesitant to add structure types and fields if all we're (currently) specifying amounts to a boolean flag.

nspark:
[...]

Issue #382, opened 2020-02-27
Add Signal API


jdinan:
## Issue

Signaling is used for point-to-point synchronization. With OpenSHMEM 1.5 we removed support for `shmem_p` as a signal operation.

Current replacements for `shmem_p` are `shmem_set` (or another AMO) or zero-length `shmem_put_signal`. Atomic set must be atomic with respect to all other AMOs; signals can provide better performance (only require single-copy atomicity).

Put-with-signal can be used for only transmitting a signal update, but for this usage model the API is not very clean and there are additional overheads (arguments, branches, etc.) compared with a simple signal operation.

## Proposed Solution

Add a signal operation:

`void shmem_signal(uint64_t *sig_addr, uint64_t signal, int sig_op, int pe);`

naveen-rn:
I would assume there is no new implicit ordering guarantees introduced with this proposal. This is simply a special case PWS with 0-zero payload. If this is the intended semantics - I think this proposal is implementable.

Also, it would be better to introduce `shmem_signal_set` and `shmem_signal_add` as separate APIs.

PR #481, opened 2021-10-15
Proposal for sessions/bundles with a start/stop API and simple example

davidozog:
This PR addresses issue #189.

As always I appreciate feedback, thanks!

### DIFF
--- a/content/bundles_intro.tex
+++ b/content/bundles_intro.tex
@@ -1,0 +1,23 @@
+\newtext{The performance of \openshmem programs that issue many consecutive and
+small-sized communication routines might be improved by combining these
+operations into fewer messages.
+The \emph{bundling routines} provide a convenient interface for indicating to
+the \openshmem library that a series of operations on a communication context
+are eligible for bundling-related optimization.
+The \FUNC{shmem\_bundle\_start} routine indicates the beginning of a bundling
+phase, and the \FUNC{shmem\_bundle\_stop} routine indicates the end of a
+bundling phase.}
+
+\newtext{The bundling routines are \textit{hints} to the \openshmem library,
+and they do not affect the completion or ordering semantics of any \openshmem
+routines in the program.
+For this reason, routines such as non-blocking RMAs, non-blocking AMOs,
+non-blocking \OPR{put-with-signal}, blocking scalar \OPR{puts}, and blocking
+non-fetching AMOs are viable candidates for bundling optimizations.
+Other routines, such as blocking non-scalar \OPR{puts} and \OPR{gets}, blocking
+fetching AMOs, blocking scalar \OPR{gets}, and the memory ordering routines
+might require the library to enforce remote completion, reducing the potential
+benefit of bundling optimizations.
+Because bundling is performed with respect to an \openshmem communication
+context, routines not performed on a communication context are ineligible for
+bundling optimization.}
--- a/content/shmem_bundle_start.tex
+++ b/content/shmem_bundle_start.tex
@@ -1,0 +1,35 @@
+\apisummary{
+    Start a communication bundle.
+}
+
+\begin{apidefinition}
+
+\begin{Csynopsis}
+void @\FuncDecl{shmem\_bundle\_start}@(void);
+void @\FuncDecl{shmem\_ctx\_bundle\_start}@(shmem_ctx_t ctx);
+\end{Csynopsis}
+
+\begin{apiarguments}
+    \apiargument{IN}{ctx}{A context handle specifying the context on which to
+    perform the optimization. When this argument is not provided, the
+    optimization is performed on the default context.}
+\end{apiarguments}
+
+\apidescription{
+    The \FUNC{shmem\_bundle\_start} routine provides a hint to the \openshmem
+    library to being applying bundling-related optimizations.
+}
+
+\apireturnvalues{
+    None.
+}
+
+\apinotes{
+    The \FUNC{shmem\_bundle\_start} routine is a hint for improving
+    performance, and \openshmem implementations are not required to apply any
+    optimization.
+}
+
+\end{apidefinition}
+
+
--- a/content/shmem_bundle_stop.tex
+++ b/content/shmem_bundle_stop.tex
@@ -1,0 +1,40 @@
+\apisummary{
+    Stop a communication bundle.
+}
+
+\begin{apidefinition}
+
+\begin{Csynopsis}
+void @\FuncDecl{shmem\_bundle\_stop}@(void);
+void @\FuncDecl{shmem\_ctx\_bundle\_stop}@(shmem_ctx_t ctx);
+\end{Csynopsis}
+
+\begin{apiarguments}
+    \apiargument{IN}{ctx}{A context handle specifying the context on which to
+    perform the optimization. When this argument is not provided, the
+    optimization is performed on the default context.}
+\end{apiarguments}
+
+\apidescription{
+    The \FUNC{shmem\_bundle\_stop} routine provides a hint to the \openshmem
+    library to stop applying bundling-related optimizations.
+}
+
+\apireturnvalues{
+    None.
+}
+
+\begin{apiexamples}
+
+\apicexample
+    {The following example demonstrates the usage of
+    \FUNC{shmem\_bundle\_start} and \FUNC{shmem\_bundle\_stop} with a loop of
+    random atomic non-fetching XOR updates to a distributed table, similar to
+    the Giga-updates per second (GUPS) microbenchmark
+    \footnote{http://icl.cs.utk.edu/projectsfiles/hpcc/RandomAccess/}.}
+    {./example_code/shmem_bundle_example.c}
+    {}
+\end{apiexamples}
+
+\end{apidefinition}
+
--- a/example_code/shmem_bundle_example.c
+++ b/example_code/shmem_bundle_example.c
@@ -1,0 +1,29 @@
+#include <shmem.h>
+#include <stdlib.h>
+
+#define N_UPDATES (1 << 20)
+#define N_INDICES (1 << 10)
+#define N_VALUES  (1 << 31)
+
+int main(void) {
+
+  shmem_init();
+
+  uint64_t *table = shmem_malloc(N_INDICES * sizeof(uint64_t));
+
+  int mype = shmem_my_pe();
+  int npes = shmem_n_pes();
+
+  shmem_bundle_start();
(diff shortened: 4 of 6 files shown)


content/bundles_intro.tex
```diff
@@ -0,0 +1,23 @@
+\newtext{The performance of \openshmem programs that issue many consecutive and
```
naveen-rn, line 1: Irrespective of small/large-sized messages - bundles/sessions provides hints to the implementation to optimize. We need a better definition of what a bundle or session refers to.

content/shmem_bundle_start.tex
```diff
@@ -0,0 +1,35 @@
+\apisummary{
+    Start a communication bundle.
+}
+
+\begin{apidefinition}
+
+\begin{Csynopsis}
+void @\FuncDecl{shmem\_bundle\_start}@(void);
```
naveen-rn, line 8: why do we need a separate API - I don't see why someone can't use the super-set API with ctx.

content/shmem_bundle_stop.tex
```diff
@@ -0,0 +1,40 @@
...
+\end{Csynopsis}
+
+\begin{apiarguments}
+    \apiargument{IN}{ctx}{A context handle specifying the context on which to
+    perform the optimization. When this argument is not provided, the
+    optimization is performed on the default context.}
+\end{apiarguments}
+
+\apidescription{
+    The \FUNC{shmem\_bundle\_stop} routine provides a hint to the \openshmem
```
naveen-rn, line 19: We are not defining the exit condition for this routine correctly - I'm not sure whether the intention is to have an implicit quiet as part of this routine. 
Based on the example - it looks like there is an implicit quiet. Not sure - why this is defined in such a way.

content/shmem_bundle_start.tex
```diff
@@ -0,0 +1,35 @@
...
+void @\FuncDecl{shmem\_ctx\_bundle\_start}@(shmem_ctx_t ctx);
```
naveen-rn, line 9: I think - we are missing two things:
1. There should a hint parameter for this routine - apart from the context. Hints should help implementations to support different options.
2. Should there be shmem_team_t as part of the argument? I'm not entirely sure about this requirement. **My initial thinking is we should support - I'm open to suggestions.** Adding teams argument - doesn't make it a collective call. Though, we can also retrieve this information from the ctx argument - it is useful to have teams argument along with or instead of teams. Proposed API:
```C++
vous shmem_bundle_start(shmem_team_t team, shmem_ctx_t ctx, int hints)
```
Hint can be used to determine whether team or ctx is passed. If multiple CTX or all CTX from a team is expected to create separate bundles - it would be optimal to use a single internal bundle.

content/shmem_bundle_stop.tex
```diff
@@ -0,0 +1,40 @@
...
+    library to stop applying bundling-related optimizations.
+}
```
naveen-rn, line 21: We are not specifically making it mandatory for users to call a stop after a start is called. I think, we should make it mandatory to call a stop after a start.

content/shmem_bundle_start.tex
```diff
@@ -0,0 +1,35 @@
...
+\begin{apiarguments}
+    \apiargument{IN}{ctx}{A context handle specifying the context on which to
+    perform the optimization. When this argument is not provided, the
+    optimization is performed on the default context.}
+\end{apiarguments}
+
+\apidescription{
+    The \FUNC{shmem\_bundle\_start} routine provides a hint to the \openshmem
+    library to being applying bundling-related optimizations.
+}
```
naveen-rn, line 21: Do we need to handle a scenario - when multiple starts are called recursively without breaking the bundle?
```c++
shmem_start_bundle(ctx1)
shmem_start_bundle(ctx1)
shmem_start_bundle(ctx1)
shmem_stop_bundle(ctx1)
```

content/shmem_bundle_start.tex
davidozog: I don't think we need it, but I included for consistency with fence, quiet, PE-specific quiet, and the RMAs/AMOs.  It's just a convenience routine - I'm happy to remove if a majority find it unnecessary.

content/shmem_bundle_stop.tex
davidozog: As discussed in the WG's, there is no implicit quiet at the end of a bundle (but there is within the `shmem_free`, right?  I don't think the example needs a quiet, but maybe the example could be improved in this regard).

My intent was for the statement:
> [bundling routines] do not affect the completion or ordering semantics of any OpenSHMEM
 routines in the program

to cover this.  But perhaps it would help to include something like this in the API description or a note?

content/shmem_bundle_start.tex
davidozog: I'm confused on this one - doesn't the context directly imply a team?  I don't see why we would allow bundling on a team other than the context it's associated with...?

naveen-rn: Nevertheless - we need hints argument irrespective of whether we plan to add team argument or not.

naveen-rn: hint would say - SHMEM_BUNDLE_TEAM or SHMEM_BUNDLE_CTX

content/shmem_bundle_stop.tex
bcernohous: And some advice that it's a bad idea  to go into non-shmem code (like a long computation loop) without a stop?

content/shmem_bundle_start.tex
bcernohous: I'm not fond of nesting or disallowing.  Just a note that stop affects them all in that example?

content/shmem_bundle_start.tex
naveen-rn: I think we should have hints inside this routine and not in the context - I don't think we want to restrict the users to use a separate context for these sort of optimizations. They should be able to use the same context within the session and outside the session

naveen-rn: If they misuse them - say use a shared context within session and simulateously outside on another thread - it is their perf issue to handle.

jdinan: I agree. Sorry, my comment was a bit confusing. I was suggesting that the parameter to the proposed function could be called `long options` if we want to follow the convention set by the contexts API.

content/shmem_bundle_start.tex
```diff
@@ -0,0 +1,35 @@
...
+
+\begin{apiarguments}
+    \apiargument{IN}{ctx}{A context handle specifying the context on which to
+    perform the optimization. When this argument is not provided, the
+    optimization is performed on the default context.}
+\end{apiarguments}
+
+\apidescription{
+    The \FUNC{shmem\_bundle\_start} routine provides a hint to the \openshmem
+    library to being applying bundling-related optimizations.
```
nspark, line 20:
```suggestion
    The \FUNC{shmem\_bundle\_start} routine provides a hint to the \openshmem
    library to begin applying bundling-related optimizations.
```

example_code/shmem_bundle_example.c
```diff
@@ -0,0 +1,29 @@
+#include <shmem.h>
+#include <stdlib.h>
```
nspark, line 2:
```suggestion
#include <shmem.h>
#include <stdint.h>
#include <stdlib.h>
```

example_code/shmem_bundle_example.c
```diff
@@ -0,0 +1,29 @@
...
+
+#define N_UPDATES (1 << 20)
+#define N_INDICES (1 << 10)
+#define N_VALUES  (1 << 31)
```
nspark, line 6:
```suggestion
#define N_UPDATES (1lu << 20)
#define N_INDICES (1lu << 10)
#define N_VALUES  (1lu << 31)
```

example_code/shmem_bundle_example.c
```diff
@@ -0,0 +1,29 @@
...
+
+int main(void) {
+
+  shmem_init();
+
+  uint64_t *table = shmem_malloc(N_INDICES * sizeof(uint64_t));
```
nspark, line 12:
```suggestion
  uint64_t *table = shmem_calloc(N_INDICES, sizeof(uint64_t));
```

example_code/shmem_bundle_example.c
```diff
@@ -0,0 +1,29 @@
...
+  int mype = shmem_my_pe();
+  int npes = shmem_n_pes();
+
+  shmem_bundle_start();
+  for (size_t i = 0; i < N_UPDATES; i++) {
+      int random_pe = rand() % npes;
+      size_t random_idx = rand() % N_INDICES;
+      uint64_t random_val = rand() % N_VALUES;
+      shmem_uint64_atomic_xor(&table[random_idx], random_val, random_pe);
+  }
```
wrrobin, line 23: Minor: the indentations can be matched.

example_code/shmem_bundle_example.c
```diff
@@ -0,0 +1,29 @@
...
+
+int main(void) {
+
+  shmem_init();
+
+  uint64_t *table = shmem_malloc(N_INDICES * sizeof(uint64_t));
+
+  int mype = shmem_my_pe();
+  int npes = shmem_n_pes();
+
```
jdinan, line 16: Suggest adding `srand(mype);`

nspark:
@naveen-rn suggested making bundling a property that could be enabled/disabled dynamically (i.e., not only at context creation), which would generalize the API for potential future dynamic properties on contexts.

nspark:
@davidozog One case of interest to me is combining bundling with put-with-signal. I would like to think there's a good bit of room for optimization internal to an OpenSHMEM library.

For example, we often think of put-with-signal as roughly equivalent to:

```c
shmem_putmem(..., pe);
shmem_fence();
shmem_set(&sigword, signal, pe);
```

(Note, we lack, and need, `shmem_signal_set`; see #382.)

In my mind, bundling + put-with-signal could effectively transform:

```c
for (int i = 0; i < bound; i++)
  shmem_put_signal_nbi(..., &sigword, SHMEM_SIGNAL_SET, pe_targets[i]);
```

Into:

```c
for (int i = 0; i < bound; i++)
  shmem_put_nbi(..., pe_targets[i]);
shmem_fence();
for (int i = 0; i < bound; i++)
  shmem_signal_set(&sigword, signal, pe_targets[i]);
```

Not every library would necessarily want to make such a transformation. But, I imagine some might find it advantageous.

Issue #467, opened 2021-05-06
Reductions and NaN values


nspark:
## Description

Currently, the Specification does not specify the handling of NaN values in reductions over floating types.

Per C18 &sect;7.12.14-1, a NaN value is _unordered_ with respect to a numeric value or another NaN. For example, it is not clear what the result of `shmem_double_max_reduce` or `shmem_double_max_to_all` should be in the presence of NaN values.

In C, NaN values can be initially unintuitive; for example:

```c
#include <math.h>
#define MAX(a, b) ((a) > (b) ? (a) : (b))

MAX(1.0, NAN) == NAN;
MAX(NAN, 1.0) == 1.0;
```

C provides [`fmax`](https://en.cppreference.com/w/c/numeric/math/fmax) and [`fmin`](https://en.cppreference.com/w/c/numeric/math/fmin) to handle these situations gracefully:

```c
#include <math.h>
fmax(1.0, NAN) == 1.0;
fmax(NAN, 1.0) == 1.0;
fmax(NAN, NAN) == NAN;
```

In the tests I've performed on OpenSHMEM implementations readily accessible to me (certainly not all that exist), none handle min/max reductions correctly for NaN values. They did seem to handle sum reductions correctly.

## Suggestions

- MAX and MIN reductions with semantic equivalence to C's `fmax` and `fmin` functions.
  - Thus, such reductions will effectively drop NaN values unless *all* input values are NaN, in which case the result should be NaN.
- SUM and PROD reductions should produce a NaN value any input value is NaN.

## Considerations

- Some OpenSHMEM implementations leverage hardware-accelerated reductions. I am not aware of the state of NaN handling for such hardware.

nspark:
I expected MPI might have plenty to say about handling NaN values, but all I see is the following:

> According to IEEE specifications, the “NaN” (not a number) is system dependent. It should not be interpreted within MPI as anything other than “NaN.”
> > _Advice to implementors._ The MPI treatment of “NaN” is similar to the approach used in XDR (see ftp://ds.internic.net/rfc/rfc1832.txt). _(End of advice to implementors.)_

nspark:
To implementors/vendors: Are there performance concerns if the result (e.g., of a MAX reduction where some entries are NaN values) is implementation defined but required to be single-valued? That is, all PEs would be expected to return the same value. (This result is not currently the case on all implementations.)

jdinan:
[...]

PR #488, opened 2022-02-28
Add inclusive and exclusive scan (prefix sum) operations

nspark:
This PR adds inclusive and exclusive scan (prefix sum) operations as `shmem_sum_inscan` and `shmem_sum_exscan` operations. Some comments:

- MPI's inclusive scan is just called `MPI_Scan`.
- `MPI_Exscan` leaves the contents of the destination buffer on rank 0 undefined. The proposed `shmem_sum_exscan` explicitly zeros the destination buffer on PE 0. (Rationale: `MPI_Exscan` supports multiple operators; this PR only supports addition.)
- C++ uses the "full" names: [std::inclusive_scan](https://en.cppreference.com/w/cpp/algorithm/inclusive_scan), [std::exclusive_scan](https://en.cppreference.com/w/cpp/algorithm/exclusive_scan)
- Inclusive scan [seems somewhat more common](https://en.wikipedia.org/wiki/Prefix_sum#Inclusive_and_exclusive_scans), though I personally envision using exclusive scan more for indexing in gather-like operations.
- The example is a bit weak.
- I personally want value-based reductions/scans _way_ more than array-based reductions/scans.

### DIFF
--- a/content/shmem_scan.tex
+++ b/content/shmem_scan.tex
@@ -51,7 +51,7 @@ by Table \ref{teamreducetypes}.
 
   The \VAR{nreduce} argument determines the number of separate scan
   operations to perform.  The \source{} array on all \acp{PE}
-  participating in the opreation provides one element for each scan.
+  participating in the operation provides one element for each scan.
   The results of the scan operations are placed in the \dest{} array
   on all \acp{PE} participating in the scan.
 
--- a/content/shmem_scan.tex
+++ b/content/shmem_scan.tex
@@ -1,0 +1,121 @@
+\apisummary {
+  Performs inclusive or exclusive prefix sum operations
+}
+
+\begin{apidefinition}
+
+%% C11
+\begin{C11synopsis}
+int @\FuncDecl{shmem\_sum\_inscan}@(shmem_team_t team, TYPE *dest, const TYPE *source, size_t nreduce);
+int @\FuncDecl{shmem\_sum\_exscan}@(shmem_team_t team, TYPE *dest, const TYPE *source, size_t nreduce);
+\end{C11synopsis}
+where \TYPE{} is one of the integer, real, or complex types supported
+for the SUM operation as specified by Table \ref{teamreducetypes}.
+
+%% C/C++
+\begin{Csynopsis}
+int @\FuncDecl{shmem\_\FuncParam{TYPENAME}\_sum\_inscan}@(shmem_team_t team, TYPE *dest, const TYPE *source, size_t nreduce);
+int @\FuncDecl{shmem\_\FuncParam{TYPENAME}\_sum\_exscan}@(shmem_team_t team, TYPE *dest, const TYPE *source, size_t nreduce);
+\end{Csynopsis}
+where \TYPE{} is one of the integer, real, or complex types supported
+for the SUM operation and has a corresponding \TYPENAME{} as specified
+by Table \ref{teamreducetypes}.
+
+\begin{apiarguments}
+  \apiargument{IN}{team}{
+    The team over which to perform the operation.
+  }
+  \apiargument{OUT}{dest}{
+    Symmetric address of an array, of length \VAR{nreduce} elements,
+    to receive the result of the scan routines.  The type of
+    \dest{} should match that implied in the SYNOPSIS section.
+  }
+  \apiargument{IN}{source}{
+    Symmetric address of an array, of length \VAR{nreduce} elements,
+    that contains one element for each separate scan routine.
+    The type of \source{} should match that implied in the SYNOPSIS
+    section.
+  }
+  \apiargument{IN}{nreduce}{
+    The number of elements in the \dest{} and \source{} arrays.
+  }
+\end{apiarguments}
+
+\apidescription{
+
+  The \FUNC{shmem\_sum\_inscan} and \FUNC{shmem\_sum\_exscan} routines
+  are collective routines over an \openshmem team that compute one or
+  more scan (or prefix sum) operations across symmetric arrays on
+  multiple \acp{PE}.  The scan operations are performed with the SUM
+  operator.
+
+  The \VAR{nreduce} argument determines the number of separate scan
+  operations to perform.  The \source{} array on all \acp{PE}
+  participating in the opreation provides one element for each scan.
+  The results of the scan operations are placed in the \dest{} array
+  on all \acp{PE} participating in the scan.
+
+  The \FUNC{shmem\_sum\_inscan} routine performs an inclusive scan
+  operation, while the \FUNC{shmem\_sum\_exscan} routine performs an
+  exclusive scan operation.  
+
+  For \FUNC{shmem\_sum\_inscan}, the value of the $j$-th element in
+  the \VAR{dest} array on \ac{PE}~$i$ is defined as:
+  \begin{equation*}
+    \textrm{dest}_{i,j} = \displaystyle\sum_{k=0}^{i} \textrm{source}_{k,j}
+  \end{equation*}
+
+  For \FUNC{shmem\_sum\_exscan}, the value of the $j$-th element in
+  the \VAR{dest} array on \ac{PE}~$i$ is defined as:
+  \begin{equation*}
+    \textrm{dest}_{i,j} =
+    \begin{cases}
+    \displaystyle\sum_{k=0}^{i-1} \textrm{source}_{k,j}, & \text{if} \; i \neq 0 \\
+    0,  & \text{if} \; i = 0
+    \end{cases}
+  \end{equation*}
+
+  The \source{} and \dest{} arguments must either be the same
+  symmetric address, or two different symmetric addresses
+  corresponding to buffers that do not overlap in memory. That is,
+  they must be completely overlapping or completely disjoint.
+
+  Team-based scan routines operate over all \acp{PE} in the provided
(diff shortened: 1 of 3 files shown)


nspark:
Has anyone seen a scan operation that generalizes on the order in which entries are accumulated (e.g., bottom-up, as drafted, vs. top-down)?

Alternatively, can an OpenSHMEM team be created that reverses the PE order? For example, is the following valid...?

```c
shmem_team_t world_reversed;
shmem_team_split_strided(SHMEM_TEAM_WORLD, shmem_n_pes() - 1, -1, shmem_n_pes(), NULL, 0, &world_reversed);
```

AFAICT, there's nothing _precluding_ the creation of such a team. I'd guess a lot of first implementations of OpenSHMEM 1.5 might break on this, though.

jdinan:
I think the MPI answer to this would be to create a communicator that renumbers the MPI processes. I don't think that a negative stride is forbidden by the OpenSHMEM spec, but I'm also not sure whether it's something that we intended to support. The legacy collectives didn't support a negative stride, since the stride argument was treated as 2^(stride). Can anyone remember better than me? @davidozog or @manjugv?

nspark:
Making a note to consider the behavior of scan operations and NaN values (cf. #467); for example:

```c
static double dst = 0;
static double src = 0;
src = (shmem_my_pe() == shmem_n_pes() / 2) ? NAN : shmem_my_pe();
shmem_sum_inscan(SHMEM_TEAM_WORLD, &dst, &src, 1);
if (shmem_my_pe() >= shmem_n_pes())
  assert(isnan(dst));
```

wrrobin:
Are we keeping the `src` single element data type for now and add the support for arrays later? Or, is the example above a specific one? For arrays, do we need to provide the indices similar to what C++ has?

content/shmem_scan.tex
```diff
@@ -0,0 +1,121 @@
...
+where \TYPE{} is one of the integer, real, or complex types supported
+for the SUM operation as specified by Table \ref{teamreducetypes}.
+
+%% C/C++
+\begin{Csynopsis}
+int @\FuncDecl{shmem\_\FuncParam{TYPENAME}\_sum\_inscan}@(shmem_team_t team, TYPE *dest, const TYPE *source, size_t nreduce);
+int @\FuncDecl{shmem\_\FuncParam{TYPENAME}\_sum\_exscan}@(shmem_team_t team, TYPE *dest, const TYPE *source, size_t nreduce);
+\end{Csynopsis}
+where \TYPE{} is one of the integer, real, or complex types supported
+for the SUM operation and has a corresponding \TYPENAME{} as specified
```
jdinan, line 21: Can we include a reference to the reduction table, and also update the caption for the table to be inclusive of scan operations?

content/shmem_scan.tex
```diff
@@ -0,0 +1,121 @@
...
+
+  The \FUNC{shmem\_sum\_inscan} and \FUNC{shmem\_sum\_exscan} routines
+  are collective routines over an \openshmem team that compute one or
+  more scan (or prefix sum) operations across symmetric arrays on
+  multiple \acp{PE}.  The scan operations are performed with the SUM
+  operator.
+
+  The \VAR{nreduce} argument determines the number of separate scan
+  operations to perform.  The \source{} array on all \acp{PE}
+  participating in the opreation provides one element for each scan.
```
nspark, line 54:
```suggestion
  participating in the operation provides one element for each scan.
```

content/shmem_scan.tex
```diff
@@ -0,0 +1,121 @@
...
+    \begin{cases}
+    \displaystyle\sum_{k=0}^{i-1} \textrm{source}_{k,j}, & \text{if} \; i \neq 0 \\
+    0,  & \text{if} \; i = 0
+    \end{cases}
+  \end{equation*}
+
+  The \source{} and \dest{} arguments must either be the same
+  symmetric address, or two different symmetric addresses
+  corresponding to buffers that do not overlap in memory. That is,
+  they must be completely overlapping or completely disjoint.
```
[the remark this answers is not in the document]
nspark: Is #290 the right reference here? I don't see how that applies here.

nspark:
> Are we keeping the `src` single element data type for now and add the support for arrays later? Or, is the example above a specific one? For arrays, do we need to provide the indices similar to what C++ has?

@wrrobin This API accepts arrays. The example just happens to only use a single element.

content/shmem_scan.tex
nspark: > Can we include a reference to the reduction table

This next line references the team-based reduction table of types, names, and operations. Is there some other table you want referenced?

> and also update the caption for the table to be inclusive of scan operations?

I'm hesitant to caption the table "~Reduction~ Types, Names, and Supporting Operations for Team-Based Reductions **and Scans"** because the scan support only applies to the SUM operation. Maybe that's okay, but it feels to me like it adds ambiguity not reduces it. Would such a caption seem to imply scan supports more than just SUM operations?

kwaters4:
@naveen-rn  Add me to this issue

content/shmem_scan.tex
```diff
@@ -0,0 +1,121 @@
+\apisummary {
+  Performs inclusive or exclusive prefix sum operations
+}
+
+\begin{apidefinition}
+
+%% C11
+\begin{C11synopsis}
+int @\FuncDecl{shmem\_sum\_inscan}@(shmem_team_t team, TYPE *dest, const TYPE *source, size_t nreduce);
+int @\FuncDecl{shmem\_sum\_exscan}@(shmem_team_t team, TYPE *dest, const TYPE *source, size_t nreduce);
```
jdinan, line 10: Should the last parameter be named something like `nelem` instead of `nreduce`?

example_code/shmem_scan_example.c
```diff
@@ -0,0 +1,12 @@
+#include <shmem.h>
+
+int collect_at(shmem_team_t team, void *dest, const void *source, size_t nbytes, int who) {
+  static size_t sym_nbytes;
+  sym_nbytes = nbytes;
+  shmem_team_sync(team);
```
BKP, line 6: Can we get rid of this sync if the src and dest buffers are different and dest is statically initialized?

If not, we may need to address this case in the section matter above.

content/shmem_scan.tex
```diff
@@ -0,0 +1,121 @@
...
+  corresponding to buffers that do not overlap in memory. That is,
+  they must be completely overlapping or completely disjoint.
+
+  Team-based scan routines operate over all \acp{PE} in the provided
+  team argument. All \acp{PE} in the provided team must participate in
+  the scan operation.  If \VAR{team} compares equal to
+  \LibConstRef{SHMEM\_TEAM\_INVALID} or is otherwise invalid, the
+  behavior is undefined.
+
+  Before any \ac{PE} calls a scan routine, the \dest{} array on all
```
BKP, line 89: In response to the message in the example section below:
By omission, this is saying that the src arrays on all pes does not need to be ready.

content/shmem_scan.tex
```diff
@@ -0,0 +1,121 @@
...
+for the SUM operation and has a corresponding \TYPENAME{} as specified
+by Table \ref{teamreducetypes}.
+
+\begin{apiarguments}
+  \apiargument{IN}{team}{
+    The team over which to perform the operation.
+  }
+  \apiargument{OUT}{dest}{
+    Symmetric address of an array, of length \VAR{nreduce} elements,
+    to receive the result of the scan routines.  The type of
```
jdinan, line 30:
```suggestion
    to receive the result of the scan operations.  The type of
```

content/shmem_scan.tex
```diff
@@ -0,0 +1,121 @@
...
+    \dest{} should match that implied in the SYNOPSIS section.
+  }
+  \apiargument{IN}{source}{
+    Symmetric address of an array, of length \VAR{nreduce} elements,
+    that contains one element for each separate scan routine.
```
jdinan, line 35:
```suggestion
    that contains one element for each separate scan operation.
```

content/shmem_scan.tex
```diff
@@ -0,0 +1,121 @@
...
+
+  Team-based scan routines operate over all \acp{PE} in the provided
+  team argument. All \acp{PE} in the provided team must participate in
+  the scan operation.  If \VAR{team} compares equal to
+  \LibConstRef{SHMEM\_TEAM\_INVALID} or is otherwise invalid, the
+  behavior is undefined.
+
+  Before any \ac{PE} calls a scan routine, the \dest{} array on all
+  \acp{PE} participating in the operation must be ready to accept the
+  results of the operation.  Otherwise, the behavior is undefined.
```
jdinan, line 91: @davidozog Proposed text to add here (and other collectives):
```
The \source{} buffer at the local \ac{PE} must be ready to be read by any PE in the team. The application does not need to synchronize to ensure that the \source{} buffer is ready across all \acp{PE} prior to calling this routine.
```

jdinan: Please feel free to bikeshed this and improve the text. :)

content/shmem_scan.tex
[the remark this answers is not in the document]
jdinan: 🤦 It was #490. Please incorporate that (minor) change to the reductions text here.

jdinan:
Collectives section committee -- there are several minor wording changes mentioned on this PR. Please incorporate those changes as section edits.

PR #489, opened 2022-03-23
Add shmem_signal_{add,set}

### LINKED ISSUE #382: Add Signal API

nspark:
- Adds `shmem_signal_{add,set}`
- Revises the signaling operations introduction and atomicity sections accordingly

Closes #382

### DIFF
--- a/content/shmem_signal_add.tex
+++ b/content/shmem_signal_add.tex
@@ -1,0 +1,47 @@
+\apisummary{
+  Adds to a signal value of a remote data object.
+}
+
+\begin{apidefinition}
+
+\begin{C11synopsis}
+void @\FuncDecl{shmem\_signal\_add}@(shmem_ctx_t ctx, const uint64_t *sig_addr, uint64_t signal, int pe);
+\end{C11synopsis}
+
+\begin{Csynopsis}
+void @\FuncDecl{shmem\_signal\_add}@(const uint64_t *sig_addr, uint64_t signal, int pe);
+void @\FuncDecl{shmem\_ctx\_signal\_add}@(shmem_ctx_t ctx, const uint64_t *sig_addr, uint64_t signal, int pe);
+\end{Csynopsis}
+
+\begin{apiarguments}
+  \apiargument{IN}{ctx}{
+    A context handle specifying the context on which to perform the
+    operation. When this argument is not provided, the operation is
+    performed on the default context.
+  }
+  \apiargument{OUT}{sig\_addr}{
+    Symmetric address of the signal data object to be updated on the
+    remote \ac{PE}.
+  }
+  \apiargument{IN}{signal}{
+    Unsigned 64-bit value that is used for updating the remote
+    \VAR{sig\_addr} signal data object.
+  }
+  \apiargument{IN}{pe}{
+    \ac{PE} number of the remote \ac{PE}.
+  }
+\end{apiarguments}
+
+\apidescription{
+  \FUNC{shmem\_signal\_add} adds \VAR{value} to the signal data
+  object pointed to by \VAR{sig\_addr} on \ac{PE}~\VAR{pe}.
+  The update to the \VAR{sig\_addr} signal object at the calling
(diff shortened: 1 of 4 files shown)


nspark:
Yes. AMOs are only guaranteed atomicity of updates w.r.t. other AMOs of the same data type (but for all operation types; e.g., set, add, xor). Signal operations are only guaranteed atomicity of updates w.r.t. other signal operations of the same operation type (i.e., add or set).

I think the intent was that, for performance reasons, `shmem_signal_set` and `shmem_put_signal(SET)` can provide (for example) single-copy atomic writes without needing the full strength of atomicity required by AMOs.

shamisp:
I think this adds quite a bit API bloat, while the semantical difference is very nuanced. I would suggest to add a parameter to shmalloc that lets you define desired AMO semantics, so you can reuse all existing AMO operations.

content/shmem_signal_add.tex
```diff
@@ -0,0 +1,47 @@
+\apisummary{
+  Adds to a signal value of a remote data object.
+}
+
+\begin{apidefinition}
+
+\begin{C11synopsis}
+void @\FuncDecl{shmem\_signal\_add}@(shmem_ctx_t ctx, const uint64_t *sig_addr, uint64_t signal, int pe);
```
nspark, line 8:
```suggestion
void @\FuncDecl{shmem\_signal\_add}@(shmem_ctx_t ctx, uint64_t *sig_addr, uint64_t signal, int pe);
```

content/shmem_signal_set.tex
```diff
@@ -0,0 +1,47 @@
+\apisummary{
+  Sets the signal value of a remote data object.
+}
+
+\begin{apidefinition}
+
+\begin{C11synopsis}
+void @\FuncDecl{shmem\_signal\_set}@(shmem_ctx_t ctx, const uint64_t *sig_addr, uint64_t signal, int pe);
```
nspark, line 8:
```suggestion
void @\FuncDecl{shmem\_signal\_set}@(shmem_ctx_t ctx, uint64_t *sig_addr, uint64_t signal, int pe);
```

content/shmem_signal_set.tex
```diff
@@ -0,0 +1,47 @@
...
+\end{C11synopsis}
+
+\begin{Csynopsis}
+void @\FuncDecl{shmem\_signal\_set}@(const uint64_t *sig_addr, uint64_t signal, int pe);
+void @\FuncDecl{shmem\_ctx\_signal\_set}@(shmem_ctx_t ctx, const uint64_t *sig_addr, uint64_t signal, int pe);
```
nspark, line 13:
```suggestion
void @\FuncDecl{shmem\_signal\_set}@(uint64_t *sig_addr, uint64_t signal, int pe);
void @\FuncDecl{shmem\_ctx\_signal\_set}@(shmem_ctx_t ctx, uint64_t *sig_addr, uint64_t signal, int pe);
```

nspark (commented):
Remove erroneous `const` type qualifiers

manjugv:
May spec meeting: Please add change log entry

jdinan:
Special ballot: [1c254ce](https://github.com/openshmem-org/specification/pull/489/commits/1c254cea3d29ac3fec357eb2b9fbddd896152994)

PR #493, opened 2022-09-28
OpenSHMEM Sessions (ctx hints) API

davidozog:
This is a continuation of the Bundles proposal, so closes the #481 pull request.  Sessions are a generalization of Bundles, so users can pass hints for context optimization that go beyond just chaining/bundling.  There is more discussion on this API design in the original ticket, #189.

As usual, I appreciate any feedback - thanks in advance!


davidozog:
Also, for those interested, an example of a bare-minimum implementation of this proposed API in Sandia OpenSHMEM here:
https://github.com/Sandia-OpenSHMEM/SOS/pull/1065

It simply ignores all hints...

davidozog:
I address some key suggestions from the unofficial reading on Sept. 30th with commit https://github.com/openshmem-org/specification/pull/493/commits/93f510e33d3158619fc7687bed8b7bb46c26e86d.  Thanks for the feedback!

Here are a few other notes from that meeting:

-  There was another request for an interface to provide _values_ to hints, such as the number of upcoming (chainable) RMA operations.  Something like an "epoch" sub-session or a "get/set" API for existing library constants.  This was suggested by Brandon Potter at AMD (but don't know his Github handle... does @khamidouche?).  Thoughts?
- Concern was raised that some implementations will not be able to support sufficiently long chains ("postlists") to see any benefit from the chaining hint in sessions.  Did I capture that correctly @manjugv / @naveen-rn?

If anybody feels I missed something, please let me know!  I unfortunately lost my PDF notes in a laptop crash, so I'm going by memory...

content/backmatter.tex
```diff
@@ -674,6 +674,10 @@ \section{Version 1.6}
     operations for team-based reductions.
 \ChangelogRef{teamreducetypes}%
 %
+\item Added the session routines, \FUNC{shmem\_session\_start} and
+    \FUNC{shmem\_session\_stop}, which allow users to pass hints to the
+    \openshmem library to apply runtime optimizations.
+\ChangelogRef{subsec:session}%
```
davidozog, line 680: Broken reference.
```suggestion
\ChangelogRef{subsec:sessions}%
```

content/sessions_intro.tex
```diff
@@ -0,0 +1,23 @@
...
+The \FUNC{shmem\_session\_start} routine indicates the beginning of a session,
+and the \FUNC{shmem\_session\_stop} routine indicates the end of a session.
+The \LibConstRef{SHMEM\_SESSION\_*} options (Table~\ref{session_opts}) indicate
+which patterns of \openshmem RMA and AMO routines will occur within a session.
+These options serve only as \textit{hints} to the library; it is up to the
+implementation whether or not to apply any optimizations within a session.
+
+Usage of the \openshmem session APIs on a particular context must comply with
+the requirements of all options set on that context.
+Starting and stoping \openshmem sessions should not affect the completion or
```
jdinan, line 16:
```suggestion
Starting and stopping \openshmem sessions should not affect the completion or
```

davidozog:
Special ballot diff:
[aa84edeb..e22a3f15](https://github.com/openshmem-org/specification/pull/493/files/aa84edeb9cba7b8ce80c21885e115e5b48146ba9..e22a3f1503d9ff97487888900522c86a67a27342)

content/backmatter.tex
```diff
@@ -674,6 +674,10 @@ \section{Version 1.6}
     operations for team-based reductions.
 \ChangelogRef{teamreducetypes}%
 %
+\item Added the session routines, \FUNC{shmem\_session\_start} and
+    \FUNC{shmem\_session\_stop}, which allow users to pass hints to the
```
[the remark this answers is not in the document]
davidozog: Oops, yes; I think these should have been renamed just like every other instance below.

How shall we handle this @jdinan?  Do you think it's eligible for a doc edit?

PR #496, opened 2023-01-20
Clarify fence orders only within same context

jdinan:
Clarify that Fence operations only guarantee ordering for operations that are performed on the same context.


manjugv:
Add errata, please @jdinan

Issue #491, opened 2022-03-25
Should team strides be positive?


nspark:
The current (1.5) definition of `shmem_team_split_strided` does not constrain the `stride` argument to be positive. As written, it may be negative or even zero, (potentially) allowing for the creation of the following teams:

```c
shmem_team_t world_reversed;
shmem_team_split_strided(SHMEM_TEAM_WORLD,
                         /* start */ shmem_n_pes() - 1, /* stride */ -1, /* size */ shmem_n_pes(),
                         NULL, 0, &world_reversed);

shmem_team_t zero_stride;
shmem_team_split_strided(SHMEM_TEAM_WORLD,
                         /* start */ shmem_my_pe(), /* stride */ 0, /* size */ 4,
                         NULL, 0, &zero_stride); 
```

For the former, should a negative-stride team be permissible? This would allow for creating teams with reversed PE indices. (See also https://github.com/openshmem-org/specification/pull/488#issuecomment-1063408888.)

For the latter, what does a zero-stride team imply? Per 1.5, it seems that the only PE(s) in the team would be the starting PE, but what value is the `size` argument here? We can't create a team with, say, four instances of the same PE in it. If `stride` is zero, does `size` only serve to distinguish zero vs. nonzero?

jdinan:
Closed by #519

PR #519, opened 2024-07-11
Teams: clarify behavior of split ops with strides <= 0

davidozog:
## Summary of changes
This addresses #491 by clarifying the behavior of team split operations when `stride` is zero or negative.

## Proposal Checklist
- [x] Link to issue(s)
- [x] Changelog entry
- [ ] Reviewed for changes to front matter
- [ ] Reviewed for changes to back matter

### DIFF
--- a/content/backmatter.tex
+++ b/content/backmatter.tex
@@ -703,6 +703,10 @@ The following list describes the specific changes in \openshmem[1.6]:
     additional arguments.
 \ChangelogRef{subsec:shmem_pcontrol}
 %
+\item Clarified the behavior of \FUNC{shmem\_team\_split\_strided} when the
+    stride argument is 0 or negative.
+\ChangelogRef{subsec:shmem_team_split_strided}
+%
 \end{itemize}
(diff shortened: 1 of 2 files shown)


jdinan:
@davidozog Start description needs to be changed in order to allow negative strides:
```
\apiargument{IN}{start}{The lowest \ac{PE} number of the subset of \acp{PE} from
the parent team that will form the new team.}
```
With a negative stride, `start` is no longer the lowest PE index of the members in the parent team.

davidozog:
> @davidozog Start description needs to be changed in order to allow negative strides...

Agreed!

How about "`the first`" PE number...?  Any suggestions on how to handle this fix?

jdinan:
I can't see a small edit that will make this clear. I'd suggest another sentence or two to explain it clearly.

jdinan:
Adding a note that "lowest" is doing some work here by forbidding "wrap-arounds". The updated text should specify "lowest/highest".

Issue #542, opened 2024-09-09
Vector synchronization APIs need some const attributes


stewartl318:
# Problem Statement

Parameters which are only read but which are passed as pointers should be marked "const" so that the compiler can (a) enforce the read-only and (b) perhaps perform extra optimizations.

credit to Phillip Marshal for noticing this.

# Proposed Changes

Mark the "cmp_value" parameters of the vector synchronization APIs as const.

# Impact on Implementations

Implementations will have to add const to their .h and .c files.

# Impact on Users

none

# References and Pull Requests

davidozog:
Closed by #543

PR #543, opened 2024-09-09
Add const to cmp_value parameters of vector synchronization apis

stewartl318:
# Summary of changes
Add const to *cmp_values in vector synchronization apis

# Proposal Checklist
- [x] Link to issue(s)
- [x] Changelog entry
- [x] Reviewed for changes to front matter
- [x] Reviewed for changes to back matter

### DIFF
--- a/content/shmem_test_all_vector.tex
+++ b/content/shmem_test_all_vector.tex
@@ -7,14 +7,14 @@
 
 \begin{C11synopsis}
 int @\FuncDecl{shmem\_test\_all\_vector}@(TYPE *ivars, size_t nelems, const int *status, int cmp,
-    TYPE *cmp_values);
+    const TYPE *cmp_values);
 \end{C11synopsis}
 where \TYPE{} is one of the standard \ac{AMO} types specified by
 Table \ref{stdamotypes}.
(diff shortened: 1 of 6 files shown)


stewartl318:
Issue #542

stewartl318:
PR should be rebased onto rc1

davidozog:
Question from WG: add this as new function signature for v1.6 - or add as an errata for v1.5?

jdinan:
This reference may be help with regard to C backwards compatibility: https://stackoverflow.com/questions/5083765/does-changing-fmystruct-a-to-fconst-mystruct-a-breaks-api-abi-in-c/5083976#5083976

As discussed, it may break tools that aren't updated with const as the call to the tool's binding for the function can potentially discard const from the argument passed by the application, which causes a compile time error.

A potentially bigger issue is with type-generic C++ bindings (which SOS has). The function name mangling will be different for the updated function and this will cause ABI breakage. Type-generic C++ bindings would need to add both const and non-const versions of the function. This isn't an issue in C because generic type selection simply chooses an ordinary C function name for one of the existing non-generic functions.

PR #558, opened 2024-11-01
Fix argument formatting

jdinan:


jdinan:
Math mode formatting was correct.

PR #570, opened 2026-01-26
Add formal API-level memory consistency model

joalsop:
Add a section formally defining memory ordering guaranteed by the OpenSHMEM API. This formalization section describes how a formal PL-level memory model like C++ can be extended to define the API-level memory semantics provided by OpenSHMEM. This adds clarity and reduces ambiguity about how OpenSHMEM interacts with the memory system, which is increasingly important as the API is implemented on accelerators with SW managed coherence

# Summary of changes

# Proposal Checklist
- [ ] Link to issue(s)
- [ ] Changelog entry
- [ ] Reviewed for changes to front matter
- [ ] Reviewed for changes to back matter


content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+    \item At least one access is a write.
+    \item At least one access is not atomic.
+    \item $A$ and $B$ are not related by \textit{hb}.
+\end{itemize}
+
+Although the axiomatic model places no requirements on when each event must physically update or access physical memory, it prohibits executions that don't adhere to a set of constraints, and by extension it disallows any architecture or compilation scheme that can result in such an execution. Specifically, C++ constrains candidate execution with the following axioms\footnote{When using SC memory order, there is an additional total order over all SC atomics that must similarly be consistent with happens-before}:
+
+\textbf{Axioms}
+
+\begin{itemize}
```
joalsop, line 32: These axioms are drawn from the axiomatic formalization of the C++ model as described  by Batty et al. in "Overhauling SC Atomics in C11 and OpenCL", 2016 (https://johnwickerson.github.io/papers/openclmm.pdf) - not sure if citation is warranted, and if so, how best to cite.

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+    \item $A$ targets the PE of the operation event that issues it.
+    \item The operation event that issues $A$ is \textit{hb} ordered before $B$, or is \textit{hb} ordered before an operation event that issues $B$.
+\end{itemize}
+
+\subsubsubsection{Remote Delivery Order (rdo)}
+
+Remote delivery order relates accesses separated by an \openshmem fence operation. However, this relation does not establish full connectivity; it only forms a connection between observable accesses to the same PE from operations \textit{hb} ordered before and after the fence, and between \textit{hb} ordered accesses before a fence and observable accesses from operations \textit{hb} ordered after a fence. Specifically, $(A, B) \in rdo$ only if there is an \openshmem fence operation $Fop$ and any of the following conditions are true:
+\begin{itemize}
+    \item $A$ is a memory access that is \textit{hb} ordered before $Fop$, $B$ is an observable access issued by an operation that is \textit{hb} ordered after $Fop$, and the context associated with $Fop$ does NOT have the $OPENSHMEM\_CTX\_NOSTORE$ option enabled.
```
joalsop: Note: As currently defined, a fence is required to establish "release" ordering semantics for a synchronizing remote write (e.g., signal_set), established via rdo. In contrast, "acquire" ordering semantics are implicit for synchronizing local reads (e.g., wait_until), established via lco.

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+\begin{itemize}
+    \item $A$ is a memory access that is \textit{hb} ordered before $QBop$, $B$ is an access that is \textit{hb} ordered after $QBop$, and the context associated with $QBop$ does NOT have the $OPENSHMEM\_CTX\_NOSTORE$ option enabled.
+    \item $A$ is a memory access that is \textit{hb} ordered before $QBop$, $B$ is an observable access issued by an operation that is \textit{hb} ordered after $QBop$, and the context associated with $QBop$ does NOT have the $OPENSHMEM\_CTX\_NOSTORE$ option enabled.
+    \item $A$ is an observable access issued by an operation that is \textit{hb} ordered before $QBop$ and $B$ is an access that is \textit{hb} ordered after $QBop$.
+    \item $A$ is an observable memory access issued by a qb-ordered operation (defined below) that is \textit{hb} ordered before $QBop$ and $B$ is an observable access issued by an operation that is \textit{hb} ordered after $QBop$.
+\end{itemize}
+
+\textbf{qb-ordered operations}: blocking or nonblocking put, nonblocking get, blocking or nonblocking atomic memory operation, blocking or nonblocking put\_signal.
+
+\subsubsubsection{API Synchronizes-With (asw)}
```
joalsop: Note: As currently defined, asw is only established between observable operation accesses. Thus, we can't use PL-level local atomics to establish happens-before ordering between PEs (i.e., a signal_set() cannot synchronize with an atomicLD(memory_order=acquire).

content/api_mem_model.tex
joalsop: Not specified in the spec, but in order to enable sensible release-style semantics for synchronizing openshmem remote stores, fences should order all prior memory accesses (not just stores)

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+\begin{itemize}
+    \item $A$ is a memory access that is \textit{hb} ordered before $Fop$, $B$ is an observable access issued by an operation that is \textit{hb} ordered after $Fop$, and the context associated with $Fop$ does NOT have the $OPENSHMEM\_CTX\_NOSTORE$ option enabled.
+    \item $A$ is an observable access issued by a fence-ordered operation (defined below) that is \textit{hb} ordered before $Fop$, $B$ is an observable access issued by an operation that is \textit{hb} ordered after $Fop$, and $A$ and $B$ target the same PE.
+\end{itemize}
+
+\textbf{fence-ordered operations}: blocking or nonblocking put, blocking or nonblocking atomic memory operation, blocking or nonblocking put\_signal.
```
joalsop, line 119: For symmetry, it might be worth defining fences to establish rdo between the remote accesses of all operations, including nonblocking gets. I currently don't see the benefit of preventing rdo ordering for the remote accesses of a nonblocking get (please chime in if you're aware of a pattern or implementation that would benefit). Therefore I suggest we update this to apply to all remote ops.

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+\subsubsection{Updating the PL-Level Model to an API-Level Model}\label{subsubsec:api_integrate}
+
+We leverage the newly defined observable accesses and ordering relations to define the API-level  memory model, which is based on the C++ model described in Section~\ref{subsubsec:api_pl_baseline}. 
+We first define a new relation, API happens-before (\textit{api\_hb}), to be the transitive closure of the C++ \textit{hb} and all API-level ordering relations introduced in Section~\ref{subsubsec:api_relations}:
+
+\begin{itemize}
+    \item $api\_hb := (hb \cup ilv \cup lco \cup rdo \cup rco \cup asw)^+$
+\end{itemize}
+
+In addition, an \textbf{API data race} is defined to relate any two accesses $A$ and $B$ where all of the following are true:
```
joalsop, line 173: TODO: define data races between synchronizing PL-level atomics and synchronizing shmem operation observable accesses (shmem can only synchronize with shmem atomics, PL can only synchronize with PL atomics)

edit: the above is completed

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+    \begin{itemize}
+        \item $B$ takes a value derived from $A$ as an operand.
+        \item $B$ has a control dependency on the result of $A$ such that $A$'s evaluation determines whether $B$ will occur.
+        \item $B$ has an ordering dependency $A$. This type of dependency currently exists only in put\_signal[\_nbi] where the final store to $sig_addr$ is dependent on all prior observable accesses.
+    \end{itemize} 
+Many operations have no visible accesses (e.g., memory ordering routines, whose semantics are defined by the API-level relations described in Section~\ref{subsubsec:api_relations}). 
+
+\input{content/api_observable}
```
joalsop, line 80: Currently, observable accesses are detailed in a large table in this section, but this might not be the best option. Possible changes:

- Keep it as a table in this section, but figure out how to format it so it looks nicer
- Include the observable accesses within the later sections describing each individual operation
- Include as a table in an appendix

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+
+
+
+The C++ memory model is defined axiomatically in the C++ specification. Although many prior works aim to mathematize or improve this formalism, here we simply try to provide sufficient understanding to build upon it with the \openshmem API model.
+
+Under the PL-level C++ memory model, events consist of reads (\textit{R}), writes (\textit{W}), atomic read-modify-writes (\textit{RMW}, which perform both a read and write to memory), and fences (\textit{F}). Events that access memory may be atomic or non-atomic. Atomic events are associated with a memory order which can be acquire, release, acquire-release, or relaxed (we leave out consume and sequential consistent memory orders here for simplicity, but this does not impact how the PL-level semantics are extended for the API-level model). The following relations are defined over the events in a candidate execution, and will be used to derive API-level relations in Section~\ref{subsubsec:api_relations}:
+
+\begin{itemize}
+    \item \textbf{Sequenced before (\textit{sb}):} Relates memory events from the same thread based on the order they are evaluated (also called program order in some contexts)
```
joalsop, line 15: Would it be more readable to expand these wherever used? same goes for newly introduced relation abbreviations. Can still set these terms apart from less formal prose using italics.

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+    \item \textbf{Modification order (\textit{mo}):} A total ordering between all writes to the same address.
+    \item \textbf{Reads-from (\textit{rf}):} Relates a write \textit{W} and any read \textit{R} to the same address such that \textit{R} returns the value written by \textit{W}.
+    \item \textbf{Synchronizes-with (\textit{sw}):} Same as \textit{rf}, but also requires that \textit{W} and \textit{R} are atomic, \textit{W} has release or acquire-release memory order, and \textit{R} has acquire or acquire-release memory order (this relation is also called synchronization order in some contexts).
```
abouteiller, line 18: I know this term is used in C++ but It overloads the synchronize term in shmem, maybe a different term would be preferred, e.g. acquire-release pair?

abouteiller: In addition, there is no definition for acquire-release memory orders, a short reminder may be useful.

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+
+After defining how \openshmem extends the C++ memory model, we describe how the addition of scoped synchronization in GPU memory models impacts the API-level model.
+
+\subsubsection{PEs and Operation Events}\label{subsubsec:api_pes_operations}
+
+When defining behavior at the \openshmem API level, we must first consider two key differences in program execution relative to a PL-level model. First, \openshmem applications span multiple PEs, each of which has its own local memory space. Therefore, the PL-level definition of memory address is extended to include both the PE-local address and the PE ID. Normal memory accesses can only access addresses on the PE local to the thread that issued it, which is constant for the life of a thread. \openshmem operations may trigger accesses to the local PE’s memory and/or the memory of remote PEs. Dynamic memory accesses in an execution are only considered to access the same address if both the address and PE ID attributes match (impacting the definitions of of \textit{mo} and \textit{rf}).
```
abouteiller:
```
Normal memory accesses can only access addresses on the PE local to the thread that issued it, which is constant for the life of a thread. 
```
The thread may access all local memory (there is no notion of thread-local memory in shmem afaik), but coherency of accesses must be enforced by PL level memory model.

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+Implicit local visibility relates local observable accesses from any \openshmem operation and prior hb-ordered memory accesses from the initiating PE. This ensures, for example, that a put operation will load an up-to-date version of data from the source buffer. Specifically, $(A, B) \in ilv$ only if all of the following are true:
+\begin{itemize}
+    \item $A$ is \textit{hb} ordered before the operation event that issues $B$.
+    \item $A$ and $B$ target the same PE (the PE of the caller thread).
+\end{itemize}
+
+\subsubsubsection{Local Completion Order (lco)}
+Local completion order relates local observable accesses from a blocking \openshmem operation and subsequent \textit{hb} ordered accesses and operations from the initiating PE. This ensures, for example, that a blocking put completes its accesses to the source buffer before a subsequent store from the initiating thread overwrites the data. Specifically, $(A, B) \in lco$ only if all of the following are true:
+\begin{itemize}
+    \item $A$ is an observable access issued from a blocking operation event (i.e., an operation that does not have nbi in its name).
```
abouteiller, line 106: some operations are implicit nbi (e.g. nonfetching atomics) so the statement about not having nbi in the name is imprecise (but helpful for comprehension).

content/api_mem_model.tex
abouteiller: I don't think that is an expected behavior in this case one should use barrier_all (If I undertand the case you discuss correctly: e.g. 
0: LD(A); fence(); signal_set(pe=1, val, 1)
1: wait_until(val, 1); put(pe=0, A, newval)
I don't think this implies LD(A) is protected from reading newval, but unsure.

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+
+Second, since the \openshmem API does not define the underlying PL-level implementation, API operations do not resolve to a function consisting of PL-level events. Thus, these operations must be treated as a new type of primitive in the API-level memory model: the operation event. An operation event specifies its type, and each type defines the set of minimum observable accesses triggered by the operation (described next). Operation events participate in \textit{sb} and derived relations, similar to normal thread access events. Although operation events may be associated with separate memory access events (see Section~\ref{subsubsec:api_observable}), the operation events themselves do not participate directly in access-based relations like \textit{mo}, \textit{rf}, and \textit{sw}.
+
+\subsubsection{Per-Operation Observable Accesses}\label{subsubsec:api_observable}
```
joalsop, line 70: is there a better term for this? possibly "internal accesses"?

content/api_mem_model.tex
```diff
@@ -182,9 +182,9 @@ \subsubsection{Updating the PL-Level Model to an API-Level Model}\label{subsubse
 
 The following axioms are similarly updated to use \textit{api\_hb} rather than \textit{hb}, and to extend RMW atomicity to \openshmem atomic memory operations:
 \begin{itemize}
-    \item $acyclic(api\_hb)$: \textit{hb} cannot have any cycles.
-    \item $irreflexive(rf;api\_hb)$: \textit{hb} must be consistent with the \textbf{rf} relation.
-    \item $irreflexive((rf^{-1})^{?};mo;rf^?;api\_hb)$: \textit{hb} must be consistent with the \textit{mo} order of writes, as well as any reads that read from writes in mo order (often called coherence requirement).
+    \item $acyclic(api\_hb)$: \textit{api_hb} cannot have any cycles.
```
markbrown314: I cannot build specification without the escape for the underscore in {api_hb}

{api_hb} -> {api\\_hb}

content/api_mem_model.tex
```diff
@@ -182,9 +182,9 @@ \subsubsection{Updating the PL-Level Model to an API-Level Model}\label{subsubse
...
+    \item $irreflexive(rf;api\_hb)$: \textit{api_hb} must be consistent with the \textbf{rf} relation.
```
markbrown314: {api_hb} -> {api\\_hb}

content/api_mem_model.tex
```diff
@@ -182,9 +182,9 @@ \subsubsection{Updating the PL-Level Model to an API-Level Model}\label{subsubse
...
+    \item $irreflexive((rf^{-1})^{?};mo;rf^?;api\_hb)$: \textit{api_hb} must be consistent with the \textit{mo} order of writes, as well as any reads that read from writes in mo order (often called coherence requirement).
```
markbrown314: {api_hb} -> {api\\_hb}

markbrown314 (commented):
Added a comment about escapes for api_hb.

content/api_mem_model.tex
joalsop: > I know this term is used in C++ but It overloads the synchronize term in shmem, maybe a different term would be preferred, e.g. acquire-release pair?

I see that "synchronize" is used in the current spec for defining bulk synchronize patterns (e.g., barriers), but I'm hesitant to redefine the existing teminology in C++ when describing the baseline memory model - I'm worried that will add confusion. I did not find the specific term "synchronize[s] with" anywhere in the existing spec - maybe we can explicitly note that this term refers to the C++ definition when used in this section and set it out with italics?

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+
+Although \openshmem may be implemented in multiple programming languages, this section describes how \openshmem extends the C++ memory consistency model. C++ is chosen as an example because of its generality, and because it forms the basis for common GPU programming languages (CUDA, hip). We describe how the guarantees described here can be adapted for GPU memory models in Section~\ref{subsubsec:api_scoped}.
+
+\subsubsection{Baseline PL-level Axiomatic Memory Model (C++)}\label{subsubsec:api_pl_baseline}
```
joalsop: Can explicitly cite C++ standard here - similar to in 9.6 page 55 in spec version 1.6. Should maybe more explicitly say this is not meant to be a substitute for the full specification.

We could also reference/base the memory model on C11 instead of C++ since the rest of the spec relies on this - however, C++ is useful because GPU languages are based on it. Curious what others think.

content/api_mem_model.tex
joalsop: here I was just trying to precisely specify "local memory" - but maybe overly verbose - if it's precise enough to say "local memory", I can make that change

content/api_mem_model.tex
joalsop: non-fetching atomics don't access local memory, so lco does not apply anyways - are there any operations that are non-blocking, access local memory, and don't have nbi in the name?

content/api_mem_model.tex
joalsop: thanks - clearly I neglected to test that last fix commit - will apply that shortly

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+
+\subsubsection{Fused Collectives}\label{subsubsec:api_integrate}
+
+\openshmem defines multiple collective operations that synchronize across multiple PEs in a team, in some cases without any observable accesses to memory (e.g., as is the case with barrier). 
+There are multiple requirements on the use of these operations. In particular, each dynamic call to a collective operation must be paired with a matching call from every other PE in the team.
+The API-level memory model fuses these matching calls into a single event in the candidate execution. 
+This fused operation and the associated observable accesses (if any) are considered to be executed collectively by all nodes in the synchronizing set of PEs, allowing us to establish ordering relations across these PEs (defined in Section~\ref{subsubsec:api_relations}).
+This does not mean the observable accesses must occur atomically, and it does not constrain how the operation is implemented (i.e., any thread or PE in the team may perform each observable access).
+
+\subsubsection{API-Level Relations}\label{subsubsec:api_relations}
```
joalsop, line 90: It's been noted that currently the api-level relations are grouped based on the operation that generates them, but this leads to some non-intuitive implications (e.g., local completion order currently establishes ordering only for blocking ops, but local completion can also be established with a quiet). We could reorganize relations based on the type of ordering established (e.g., include quiet-established local completion as part of lco) or alternatively think of a different naming scheme...

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+
+\subsubsection{Per-Operation Observable Accesses}\label{subsubsec:api_observable}
+
+For the purposes of the API memory model, each \openshmem operation type defines the minimum set of observable memory accesses performed by the operation. These accesses are considered to be "issued" by the associated. These accesses must occur as part of the operation, however there is no guarantee that they will be executed by the initiating thread, or that they will be the only accesses performed by the API operation implementation. Observable accesses have the same semantics as normal accesses and take part in the same relations, with one exception: there is no \textit{sb} relationship between observable accesses and the initiating thread of an API operation. Accesses performed by an implementation that are not included in the observable set must not modify data that may be observed from application code external to \openshmem operation implementations.
```
gonzalobg: > These accesses are considered to be "issued" by the associated **[something is missing]**.

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+
+
+
+The C++ memory model is defined axiomatically in the C++ specification. Although many prior works aim to mathematize or improve this formalism, here we simply try to provide sufficient understanding to build upon it with the \openshmem API model.
+
+Under the PL-level C++ memory model, events consist of reads (\textit{R}), writes (\textit{W}), atomic read-modify-writes (\textit{RMW}, which perform both a read and write to memory), and fences (\textit{F}). Events that access memory may be atomic or non-atomic. Atomic events are associated with a memory order which can be acquire, release, acquire-release, or relaxed (we leave out consume and sequential consistent memory orders here for simplicity, but this does not impact how the PL-level semantics are extended for the API-level model). The following relations are defined over the events in a candidate execution, and will be used to derive API-level relations in Section~\ref{subsubsec:api_relations}:
```
lstewart: What does PL-level mean?

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+\end{itemize}
+
+Although the axiomatic model places no requirements on when each event must physically update or access physical memory, it prohibits executions that don't adhere to a set of constraints, and by extension it disallows any architecture or compilation scheme that can result in such an execution. Specifically, C++ constrains candidate execution with the following axioms\footnote{When using SC memory order, there is an additional total order over all SC atomics that must similarly be consistent with happens-before}:
+
+\textbf{Axioms}
+
+\begin{itemize}
+    \item $acyclic(hb)$: \textit{hb} cannot have any cycles.
+    \item $irreflexive(rf;hb)$: \textit{hb} must be consistent with the \textbf{rf} relation.
+    \item $irreflexive((rf^{-1})^{?};mo;rf^?;hb)$: \textit{hb} must be consistent with the \textit{mo} order of writes, as well as any reads that read from writes in mo order (often called coherence requirement).
```
lstewart: It would likely be useful to know who is reading or writing.  Are these relations always referring to different PEs or are they referring to different threads in the same PE or what?  In the GPU case, are we covering relations between CPU threads and GPU threads?  CPU-CPU only? GPU-GPU only?

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+
+Although \openshmem may be implemented in multiple programming languages, this section describes how \openshmem extends the C++ memory consistency model. C++ is chosen as an example because of its generality, and because it forms the basis for common GPU programming languages (CUDA, hip). We describe how the guarantees described here can be adapted for GPU memory models in Section~\ref{subsubsec:api_scoped}.
```
lstewart: Maybe either mention all the GPU languages (SYCL, Vulkan, OpenCL) or none of them, or just say "for example"

content/api_observable.tex
```diff
@@ -0,0 +1,436 @@
...
+    \tabularnewline\cline{2-2}
+    & \begin{verbatim}shmem_atomic_compare_swap_nbi(fetch, dest, cond, 
+value, pe):
+  R1 = LD dest, pe
+  if R1 == cond:
+    ST value -> dest, pe
+  ST R1 -> fetch, my_pe
+    \end{verbatim}
+    \tabularnewline\cline{2-2}
+    & \begin{verbatim}shmem_atomic_swap_nbi(fetch, dest, cond, value,
```
lstewart: Does swap_nbi really have a cond argument?

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+
+\subsubsubsection{Local Completion Order (lco)}
+Local completion order relates local observable accesses from a blocking \openshmem operation and subsequent \textit{hb} ordered accesses and operations from the initiating PE. This ensures, for example, that a blocking put completes its accesses to the source buffer before a subsequent store from the initiating thread overwrites the data. Specifically, $(A, B) \in lco$ only if all of the following are true:
+\begin{itemize}
+    \item $A$ is an observable access issued from a blocking operation event (i.e., an operation that does not have nbi in its name).
+    \item $A$ targets the PE of the operation event that issues it.
+    \item The operation event that issues $A$ is \textit{hb} ordered before $B$, or is \textit{hb} ordered before an operation event that issues $B$.
+\end{itemize}
+
+\subsubsubsection{Remote Delivery Order (rdo)}
```
lstewart, line 111: rfo doesn't apply to put_signal?  Or is that defined elsewhere.  I naively think put, fence, put is equivalent to put_signal

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+\begin{itemize}
+    \item $A$ is a memory access that is \textit{hb} ordered before $Fop$, $B$ is an observable access issued by an operation that is \textit{hb} ordered after $Fop$, and the context associated with $Fop$ does NOT have the $SHMEM\_CTX\_NOSTORE$ option enabled.
+    \item $A$ is an observable access issued by a fence-ordered operation (defined below) that is \textit{hb} ordered before $Fop$, $B$ is an observable access issued by an operation that is \textit{hb} ordered after $Fop$, and $A$ and $B$ target the same PE.
+\end{itemize}
+
+\textbf{fence-ordered operations}: blocking or nonblocking put, blocking or nonblocking atomic memory operation, blocking or nonblocking put\_signal.
```
lstewart, line 119: In GPU implementations, there are API calls made on the GPU, and API calls made on the host, and they are not completely ordered w.r.t. each other,  Often, ordering is established only at kernel boundaries.  I think it is technically feasible to have a GPU initiated FENCE affect host behavior in current implementations, but I am not sure that host FENCE can affect GPU behavior.
There are proposals to make host initiated and gpu initiated be different PEs, or sub-PEs or something, but this is not resolved.

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+    \item $A$ is an observable access issued by an operation that is \textit{hb} ordered before $QBop$ and $B$ is an access that is \textit{hb} ordered after $QBop$.
+    \item $A$ is an observable memory access issued by a qb-ordered operation (defined below) that is \textit{hb} ordered before $QBop$ and $B$ is an observable access issued by an operation that is \textit{hb} ordered after $QBop$.
+\end{itemize}
+
+\textbf{qb-ordered operations}: blocking or nonblocking put, nonblocking get, blocking or nonblocking atomic memory operation, blocking or nonblocking put\_signal.
+
+\subsubsubsection{API Synchronizes-With (asw)}\label{subsubsubsec:api_relations_asw}
+
+API synchronizes-with is the API-level analog of the PL-level \textit{sw} relation. It relates observable accesses from legally synchronizing \openshmem operations that form a \textit{rf} relationship. Specifically, $(A, B) \in asw$ only if $(A, B) \in rf$ and any of the following are true:
```
lstewart, line 135: This is the only occurance of "rf$" I can find.  What should it be? I also see one asw$ so this is some markup thing I don't understand?

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+\begin{itemize}
+    \item $A$ is an observable read or read-modify-write access issued by a synchronizing \openshmem operation.
+    \item $B$ is an observable write or read-modify-write access issued by a synchronizing \openshmem operation.
+\end{itemize}
+
+In the above, "synchronizing \openshmem operations" are defined as any operation from the following categories: atomic memory operations, signaling operations, point-to-point synchronization routings, or distributed locking routines.
```
lstewart: Do you mean to exclude shmem_sync here? or quiet + sync? or barrier?

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+\end{itemize}
+
+The following axioms are similarly updated to use \textit{api\_hb} rather than \textit{hb}, and to extend RMW atomicity to \openshmem atomic memory operations:
+\begin{itemize}
+    \item $acyclic(api\_hb)$: \textit{api_hb} cannot have any cycles.
+    \item $irreflexive(rf;api\_hb)$: \textit{api_hb} must be consistent with the \textbf{rf} relation.
+    \item $irreflexive((rf^{-1})^{?};mo;rf^?;api\_hb)$: \textit{api_hb} must be consistent with the \textit{mo} order of writes, as well as any reads that read from writes in mo order (often called coherence requirement).
+    \item $irreflexive(rf;(sb^{amo\_op})^? \cup mo;mo;(;sb^{amo\_op})^?;rf^-1 \cup mo;rf;(sb^{amo\_op})^?$ where $sb^{amo\_op}$ relates $sb$ ordered observable accesses issued by the same \openshmem atomic memory operation: The read of an atomic read-modify-write access must always read the last value (in the modification order) written before the write of the read-modify-write access. Similarly, the observable read of an \openshmem atomic memory operation must always read the last value (in the modification order) written before the observable write of the atomic memory operation.
```
lstewart: According to the spec, shmem atomics are only atomic w.r.t. other shmem atomics.  This statement is far stricter, and doesn't match up with implementations using NIC atomics.

content/api_observable.tex
```diff
@@ -0,0 +1,436 @@
...
+    \end{verbatim}
+    \tabularnewline\cline{2-2}
+    & \begin{verbatim}shmem_atomic_compare_swap(dest, cond, value, pe):
+  R1 = LD dest, pe
+  if R1 == cond:
+    ST value -> dest, pe
+  return R1
+    \end{verbatim}
+    \tabularnewline\cline{2-2}
+    & \begin{verbatim}shmem_atomic_swap(dest, cond, value, pe):
```
lstewart: Doesn't have cond

content/api_mem_model.tex
```diff
@@ -0,0 +1,200 @@
...
+\begin{itemize}
+\item \textit{ilv} orders the store of $a$ in PE0 (and all prior hb-ordered accesses) with the load of $a$ from the put operation.
+\item \textit{lco} orders the load of $a$ in PE1 (and all subsequent hb-ordered accesses) with the load of $b$ from the (blocking) wait\_until operation.
+\item \textit{rdo} orders the bulk store of a from the put operation in PE0 with the store of $b$ from the fence-separated atomic\_set operation.
+\item \textit{rco} orders all \textit{hb} ordered accesses and the accesses of \textit{hb} ordered \openshmem operations before and after the fused barrier event spanning PE0 and PE1.
+\item \textit{asw} orders the store of $b$ from the atomic\_set operation in PE0 with the consuming (\textit{rf} related) load of $b$ from the wait\_until operation in PE1.
+\end{itemize}
+
+
+\subsubsection{Updating the PL-Level Model to an API-Level Model}\label{subsubsec:api_integrate}
```
wfaderhold21, line 164: Isn't this label already defined on line 84?

content/api_mem_model.tex
joalsop: the programming language-level C11 memory model

content/api_mem_model.tex
joalsop: as described here, this applies to any threads that are using the C11 memory model. However, a later section describes how it is applicable to scoped memory models like hip, CUDA, HSA, etc. which may have CPU and/or GPU thread. These memory models are based on C11, but add scopes. The distinction of which devices are synchronizing impacts which scopes may be used for legal synchronization, but does not impact the semantics of the API-level model (any scoped synchronization will occur transparently within the implementation of the API call).

content/api_observable.tex
joalsop: nope - thanks for catching that

content/api_mem_model.tex
joalsop: currently the ordering between the put and the signal in put_signal is defined as an sb relationship between every put access and the final signal store (see the conditions for sb in the observable accesses section) - so it will be slightly different from put fence signal_set in that it only establishes ordering between the accesses of the put_signal operation, and doesn't establish any ordering between all prior and successive accesses like a normal fence (e.g., if you have put, put, put_signal, only the final put will have guaranteed ordering with the signal op). However, this was one area where the existing spec was not clear - I'm open to defining it differently.

content/api_mem_model.tex
joalsop: The intent here is to allow a fence on a GPU to order operations on the CPU if they have established hb order between the threads in some way, and they use the same context. this may be valuable at some point - but if a programmer doesn't want that, then using separate contexts or separate PEs as you say would be a way to get around it I think.

content/api_mem_model.tex
joalsop: this refers to the reads-from relation in the underyling PL-level model - the $ is just part of the math formatting in latex

content/api_mem_model.tex
joalsop: Yes - these do not have observable accesses - so they can't form a synchronizing relationship as defined here

content/api_mem_model.tex
joalsop: Currently we avoid interactions/guarantees of atomicity between shmem atomics and "normal" PL-level atomics by defining them as a data race in the API-level model. That is, if shmem atomics and normal atomics race, that forms a data race. Does that address this concern? This is an area that I'm less familiar with, so open to alternative solutions.

content/api_mem_model.tex
joalsop: yes - fixed, thanks

content/api_mem_model.tex
```diff
@@ -0,0 +1,198 @@
...
+
+This section describes how \openshmem extends the C11 memory consistency model. C11 is chosen as an example because of its generality and ubiquity in the CPU domain, and because it forms the basis for common GPU programming languages (for example, CUDA, hip). Section~\ref{subsubsec:api_pl_baseline} summarizes the baseline C11 memory model, Sections~\ref{subsubsec:api_model_overview}-\ref{subsubsec:api_integrate} describe how this programming language-level model is extended to provide API-level semantics for \openshmem, and Section~\ref{subsubsec:api_scoped} describes how the guarantees of the API-level model can be adapted for GPU memory models.
```
wfaderhold21, line 4: This section - > This Section 
?

content/api_mem_model.tex
```diff
@@ -0,0 +1,198 @@
...
+    \item $irreflexive(rf \cup (mo;mo;rf^{-1}) \cup (mo;rf))$: The read of an atomic read-modify-write access must always read the last value (in \textit{mo}) written before the write of the read-modify-write access.
+\end{itemize}
+
+
+Finally, the C11 specification states that, if a data race exists in a candidate execution that satisfies the above axioms, then the program's behavior is undefined. If a data race does not exist, then the program's behavior must conform to one of the conforming candidate executions.
+
+\subsubsection{API-Level Memory Model Overview}\label{subsubsec:api_model_overview}
+
+The \openshmem API-level memory model extends PL-level memory models like C11 to define how \openshmem operations interact with memory. While PL-level models operate on fine-grain loads, stores, and fences, the API-level memory model must additionally operate on coarse-grain API operations like get, put, barrier, etc. that can execute asynchronously and involve multiple PEs.
```
wfaderhold21, line 44: Aren't there macros for get / put ? also PEs -> \acp{PE}

content/api_mem_model.tex
lstewart: My understanding is that SHMEM users really should not mix atomic and non atomic references to the same addresses.

content/api_observable.tex
```diff
@@ -0,0 +1,540 @@
...
+Point-to-Point Synchronization Routine
+    & \begin{BVerbatim}
+shmem_test_some(ivars, nelems, indices, status,
+cmp, cmp_value):
+  num_found = 0
+  for i in nelems:
+    R1 = LD ivar, my_pe
+    if cmp.test(R1, cmp_value)
+      return i
+      ST i -> indices[num_found], my_pe
```
wfaderhold21, line 450: dead code?

content/api_mem_model.tex
```diff
@@ -0,0 +1,226 @@
...
+
+\subsubsection{Per-Operation Observable Accesses}\label{subsubsec:api_observable}
+
+For the purposes of the API memory model, each \openshmem operation type defines the minimum set of observable memory accesses performed by the operation. These accesses are considered to be ``issued'' by the associated operation. These accesses must occur as part of the operation, however the API-level model places no constraints on the thread that executes observable accesses (i.e., they may or may not be executed by the initiating thread), and it does not prevent an implementation from performing other accesses (e.g., for implementation-internal synchronization, buffering, etc. -- although data accessed in this way should never be accessed by code external to the implementation.)
```
wfaderhold21: nit: when we have text that is written like the following: `accesses (e.g., ... implementation.)` let's instead put the period outside of the parathensis.

content/api_mem_model.tex
```diff
@@ -0,0 +1,226 @@
...
+    \begin{itemize}
+        \item $B$ takes a value derived from $A$ as an operand.
+        \item $B$ has a control dependency on the result of $A$ such that $A$'s evaluation determines whether $B$ will occur.
+	\item $B$ has an ordering dependency $A$. This type of dependency currently exists only in blocking and non-blocking \emph{put-with-signal} operations where the final store to $sig\_addr$ is ordered after all prior observable accesses to source and dest.
+    \end{itemize} 
+Many operations have no visible accesses (e.g., memory ordering routines, whose semantics are defined by the API-level relations described in Section~\ref{subsubsec:api_relations}). 
+In addition, synchronization semantics for observable accesses in the API model are defined at a higher level than in the PL model. Rather than associating atomic semantics and acquire/release ordering for each observable accesses that may be involved in synchronization, we instead associate semantics at the level of the issuing operation. Specifically, API-level relations are defined based on
```
wfaderhold21: from reading. sentence dropped off?

joalsop:
Here are some supplementary slides to help understand the need for formalism, how formalism is defined, some litmus tests, and some reasoning behind why the synchronization semantics were chosen/possible future extensions.

[openshmem_formal_mem_model_supplementary_slides.pptx](https://github.com/user-attachments/files/28407696/openshmem_formal_mem_model_supplementary_slides.pptx)

jeffhammond:
The following definitions of `SHMEM_CTX_NOSTORE` are not consistent.

Page 23:
> The context creation option which specifies that quiet and fence operations performed on the given context are not required to enforce completion and ordering of _memory store operations_.

Page 59:
> Quiet and fence operations performed on the given context are not required to enforce completion and ordering of _memory accesses_ performed by the application.

jeffhammond:
Where is the definition of "normal memory access"?  That is not an obvious term of art for me.

content/api_mem_model.tex
```diff
@@ -0,0 +1,226 @@
...
+\end{enumerate}
+
+After defining how \openshmem extends the C11 memory model, Section~\ref{subsubsec:api_scoped} describes how the addition of scoped synchronization in GPU memory models impacts the API-level model.
+
+\subsubsection{PEs and Operation Events}\label{subsubsec:api_pes_operations}
+
+When defining behavior at the \openshmem API level, there are two key differences in program representation relative to a PL-level program. First, \openshmem applications span multiple \acp{PE}, each of which has its own local memory space. Therefore, the PL-level definition of memory location is extended to include both the \ac{PE}-local memory address and the \ac{PE} ID. Normal memory accesses can access memory locations on the local \ac{PE}, or on a remote \ac{PE} if \FUNC{shmem\_ptr} is used. \openshmem operations may trigger accesses to the local \ac{PE}’s memory and/or the memory of remote \acp{PE}. Dynamic memory accesses in an execution are only considered to access the same memory location if both the local memory address and \ac{PE} ID attributes match (impacting the definitions of of \relMO{} and \relRF{}).
```
naveen-rn: typo - of of

content/api_mem_model.tex
```diff
@@ -0,0 +1,226 @@
...
+
+\subsubsection{Per-Operation Observable Accesses}\label{subsubsec:api_observable}
+
+For the purposes of the API memory model, each \openshmem operation type defines the minimum set of observable memory accesses performed by the operation. These accesses are considered to be ``issued'' by the associated operation. These accesses must occur as part of the operation, however the API-level model places no constraints on the thread that executes observable accesses (i.e., they may or may not be executed by the initiating thread), and it does not prevent an implementation from performing other accesses (e.g., for implementation-internal synchronization, buffering, etc. -- although data accessed in this way should never be accessed by code external to the implementation).
```
naveen-rn: > should never be accessed

The spec's normative vocabulary is must / may / undefined. "should never be accessed" reads as a requirement — please try to use must not for a hard constraint (or move the clause to a Note if it's advisory).

content/api_mem_model.tex
```diff
@@ -0,0 +1,226 @@
...
+Observable accesses have the same semantics as normal accesses and take part in the same relations, with two exceptions: 
+\begin{enumerate}
+  \item There is no \relSB{} relationship between observable accesses issued by an API operation and any other event in the thread that initiated the operation. Instead, Section~\ref{subsubsec:api_relations} defines the relations that constrain ordering at the API level. Note, however, that the operation event that issues these observable accesses \textit{does} form a \relSB{} relationship with other events issued by the initiating thread.
+  \item There are no synchronization semantics (atomic, acquire/release ordering) associated with observable accesses. Although the accesses may be lowered to PL-level accesses with these semantics, at the API level, ordering relations and data race conditions are defined a the level of the operation that issues them (see Sections~\ref{subsubsec:api_relations} and~\ref{subsubsec:api_integrate}, respectively).
```
naveen-rn: typo - defined at the level - not defined a level

content/api_mem_model.tex
```diff
@@ -0,0 +1,226 @@
...
+However, this single-threaded control flow does not constrain the actual implementation and does not necessarily imply a total \relSB{} ordering between observable accesses. Within an operation, \relSB{} is defined to exist only between observable accesses when there is a true control or data dependency between two accesses (i.e., no \relSB{} relation across independent iterations of a loop). 
+Specifically, $(A, B) \in sb$ only if $A$ and $B$ are observable accesses issued by the same operation and one of the following is true:
+    \begin{itemize}
+        \item $B$ takes a value derived from $A$ as an operand.
+        \item $B$ has a control dependency on the result of $A$ such that $A$'s evaluation determines whether $B$ will occur.
+	\item $B$ has an ordering dependency $A$. This type of dependency currently exists only in blocking and non-blocking \emph{put-with-signal} operations where the final store to $sig\_addr$ is ordered after all prior observable accesses to source and dest.
```
naveen-rn: dependency "on" $A$ - missing

content/api_mem_model.tex
```diff
@@ -0,0 +1,226 @@
...
+    \item $irreflexive((rf^{-1})^{?};mo;rf^?;api\_hb)$: \relAHB{} must be consistent with \relMO{} between writes, and it must also be consistent with the order of reads that read from writes in \relMO{} (often called coherence requirement).
+    \item $irreflexive(rf \cup (mo;mo;rf^{-1}) \cup (mo;rf))$: The read of an atomic read-modify-write access must always read the last value (in \relMO{}) written before the write of the read-modify-write access (often called RMW atomicity).
+    \item $irreflexive(rf;sb^{amo\_op} \cup (mo;mo;(sb^{amo\_op})^{-1};rf^{-1}) \cup (mo;rf))$: The observable read of an \openshmem \ac{AMO} must always read the last value (in \relMO{}) written before the observable write of the \ac{AMO} (this is the \ac{AMO} analog of the PL-level RMW atomicity axiom). In this axiom, $sb^{amo\_op}$ is defined as the $sb$ relation between the observable read and write accesses issued by an \ac{AMO}.
+    \item $irreflexive(rf \cup (mo;mo;rf^{-1}) \cup (mo;rf))$: The read of an atomic read-modify-write access must always read the last value (in \relMO{}) written before the write of the read-modify-write access.
```
naveen-rn: duplicate axiom - it is a verbatim of line 212 - unless I'm missing something

content/api_mem_model.tex
```diff
@@ -0,0 +1,226 @@
...
+\begin{itemize}
+    \item $acyclic(api\_hb)$: \relAHB{} cannot have any cycles.
+    \item $irreflexive(rf;api\_hb)$: \relAHB{} must be consistent with the \relRF{} relation.
+    \item $irreflexive((rf^{-1})^{?};mo;rf^?;api\_hb)$: \relAHB{} must be consistent with \relMO{} between writes, and it must also be consistent with the order of reads that read from writes in \relMO{} (often called coherence requirement).
+    \item $irreflexive(rf \cup (mo;mo;rf^{-1}) \cup (mo;rf))$: The read of an atomic read-modify-write access must always read the last value (in \relMO{}) written before the write of the read-modify-write access (often called RMW atomicity).
+    \item $irreflexive(rf;sb^{amo\_op} \cup (mo;mo;(sb^{amo\_op})^{-1};rf^{-1}) \cup (mo;rf))$: The observable read of an \openshmem \ac{AMO} must always read the last value (in \relMO{}) written before the observable write of the \ac{AMO} (this is the \ac{AMO} analog of the PL-level RMW atomicity axiom). In this axiom, $sb^{amo\_op}$ is defined as the $sb$ relation between the observable read and write accesses issued by an \ac{AMO}.
```
naveen-rn: As written, the disjuncts 1–2 are bridged but the disjunct 3 is not. This is internally inconsistent with loop closure argument and appears to leave the AMO atomicity not fully constrained. 

Either fix the 3rd disjunct to be symmetric with 1 and 2, or please state if this is caught somewhere in the coherence axiom - as it is now it is vague.

content/api_mem_model.tex
```diff
@@ -0,0 +1,226 @@
...
+    \item At least one access is a modifying access (write or read-modify-write).
+    \item Both $A$ and $B$ are normal accesses and at least one of them is not atomic, or both $A$ and $B$ are observable memory accesses issued by \openshmem operations and at least one of them is issued by a non-synchronizing \openshmem operation (defined in Section~\ref{subsubsec:api_relations}), or one of the accesses is a normal memory access and the other is an observable memory access issued by an \openshmem operation.
+    \item $A$ and $B$ are not related by \relAHB{}.
+\end{itemize}
+
+The following axioms are similarly updated to use \relAHB{} rather than \relHB{}, and to extend RMW atomicity to \openshmem \ac{AMO}s:
+\begin{itemize}
+    \item $acyclic(api\_hb)$: \relAHB{} cannot have any cycles.
+    \item $irreflexive(rf;api\_hb)$: \relAHB{} must be consistent with the \relRF{} relation.
+    \item $irreflexive((rf^{-1})^{?};mo;rf^?;api\_hb)$: \relAHB{} must be consistent with \relMO{} between writes, and it must also be consistent with the order of reads that read from writes in \relMO{} (often called coherence requirement).
```
naveen-rn, line 211: No pun intended - but the consistency axiom and the coherence axioms are not verbatim - even though they refer to the same.

line 36 (as well as any reads that read from writes) is different than what is here in this item

content/api_observable.tex
```diff
@@ -0,0 +1,540 @@
...
+shmem_get[_nbi](dest, source, nelems, pe):
+  for i in nelems:
+    R1 = LD source[i], pe
+    ST R1 -> dest[i], my_pe
+    \end{BVerbatim}
+    \tabularnewline\hline
+Remote Access Routine
+    & \begin{BVerbatim}
+shmem_g(source, pe):
+  return LD source
```
naveen-rn: missing target PE

content/api_observable.tex
```diff
@@ -0,0 +1,540 @@
...
+  return R1
+    \end{BVerbatim}
+    \tabularnewline\hline
+Atomic Memory Operation
+    & \begin{BVerbatim}
+shmem_atomic_inc(dest, pe):
+  R1 = LD dest, pe
+  R2 = R1 + 1
+  ST R2 -> dest, pe
+  return R1
```
naveen-rn: looks like a typo - remove the return - there is no return in atomic_inc - the atomic_add below is correct

content/api_observable.tex
```diff
@@ -0,0 +1,540 @@
...
+    & \begin{BVerbatim}
+shmem_atomic_swap_nbi(fetch, dest, value, pe):
+  R1 = LD dest, pe
+  ST value -> dest, pe
+  ST R1 -> fetch, my_pe
+    \end{BVerbatim}
+    \tabularnewline\hline
+Atomic Memory Operation
+    & \begin{BVerbatim}
+shmem_atomic_fetch_inc_nbi(dest, pe):
```
naveen-rn: there is no fetch in the argument list - this needs fix
that said, the body uses fetch correctly

content/api_observable.tex
```diff
@@ -0,0 +1,540 @@
...
+  R1 = LD dest, pe
+  R2 = R1 + 1
+  ST R2 -> dest, pe
+  ST R1 -> fetch, my_pe
+    \end{BVerbatim}
+    \tabularnewline\hline
+Atomic Memory Operation
+    & \begin{BVerbatim}
+shmem_atomic_fetch_add_nbi(dest, value, pe):
```
naveen-rn: same fetch issue as above

content/api_observable.tex
```diff
@@ -0,0 +1,540 @@
...
+  R1 = LD dest, pe
+  R2 = R1 + value
+  ST R2 -> dest, pe
+  ST R1 -> fetch, my_pe
+    \end{BVerbatim}
+    \tabularnewline\hline
+Atomic Memory Operation
+    & \begin{BVerbatim}
+shmem_atomic_fetch_and_nbi(dest, value, pe):
```
naveen-rn: fetch missing

content/api_observable.tex
```diff
@@ -0,0 +1,540 @@
...
+  R1 = LD dest, pe
+  R2 = R1 & value
+  ST R2 -> dest
```
naveen-rn: missing pe

content/api_observable.tex
```diff
@@ -0,0 +1,540 @@
...
+  ST R1 -> fetch, my_pe
+    \end{BVerbatim}
+    \tabularnewline\hline
+Atomic Memory Operation
+    & \begin{BVerbatim}
+shmem_atomic_fetch_or_nbi(dest, value, pe):
```
naveen-rn: same fetch and pe comment from before

content/api_observable.tex
```diff
@@ -0,0 +1,540 @@
...
+  R1 = LD dest, pe
+  R2 = R1 | value
+  ST R2 -> dest
+  ST R1 -> fetch, my_pe
+    \end{BVerbatim}
+    \tabularnewline\hline
+Atomic Memory Operation
+    & \begin{BVerbatim}
+shmem_atomic_fetch_xor_nbi(dest, value, pe):
```
naveen-rn: same fetch and pe comment from before

content/api_observable.tex
```diff
@@ -0,0 +1,540 @@
...
+shmem_put_signal[_nbi](dest, source, nelems,
+sig_addr, signal, sig_op, pe):
+  for i in nelems:
+    R1 = LD source[i], my_pe
+    ST R1 -> dest[i], pe
+  switch sig_op:
+  case SHMEM_SIGNAL_SET:
+    ST signal -> sig_addr, pe
+  case SHMEM_SIGNAL_ADD:
+    R2 = R1 + signal
```
naveen-rn: isn't this adding a wrong base

content/api_observable.tex
```diff
@@ -0,0 +1,540 @@
...
+nelems):
+  chunk_size = nelems/size(team)
+  for pe_src in team:
+    for pe_dst in team
+      for i in chunk_size:
+        src_idx = sst*(chunk_size*pe_dst+i)
+        dst_idx = dst*(chunk_size*pe_src+i)
+        R1 = LD source[src_idx], pe_src
+        ST R1 -> dest[dst_idx], pe_dst
+  for i in nelems:
```
naveen-rn, line 305: looks like a copy paste error - looks like bcast is somehow copied here - unless i'm missing something

naveen-rn: that said - it looks like bcast is missing from this table - i tried to look through this list - didnt see the rendered doc though

content/api_observable.tex
```diff
@@ -0,0 +1,540 @@
...
+Collective Routine
+    & \begin{BVerbatim}
+shmem_<op>_reduce(team, dest, source, nreduce):
+  for i in nreduce:
+    R1 = op.init
+    for pe in team:
+      R2 = LD source[i], pe
+      R1 = op.func(R1, R2)
+    for pe in team:
+      ST R1 -> dest[i], p
```
naveen-rn: typo - p -> pe

content/api_observable.tex
```diff
@@ -0,0 +1,540 @@
...
+    \end{BVerbatim}
+    \tabularnewline\hline
+Point-to-Point Synchronization Routine
+    & \begin{BVerbatim}
+shmem_wait_until_all(ivars, nelems, status,
+cmp, cmp_value):
+  for i in nelems:
+    R1 = LD ivars[i], my_pe
+    while !cmp.test(R1, cmp_value):
+      R1 = LD ivar, my_pe
```
naveen-rn: there is no ivar - it should be ivar[i]

content/api_observable.tex
```diff
@@ -0,0 +1,540 @@
...
+    \tabularnewline\hline
+Point-to-Point Synchronization Routine
+    & \begin{BVerbatim}
+shmem_wait_until_all_vector(ivars, nelems, status,
+cmp, cmp_values):
+  for i in nelems:
+    R1 = LD ivars[i], my_pe
+    R2 = LD cmp_values[i], my_pe
+    while !cmp.test(R1, R2):
+      R1 = LD ivar, my_pe
```
naveen-rn: same ivar issue

naveen-rn: looks like few other places have the same array index issue

content/api_observable.tex
```diff
@@ -0,0 +1,540 @@
...
+    while !cmp.test(R1, cmp_value):
+      R1 = LD ivar, my_pe
+    \end{BVerbatim}
+    \tabularnewline\hline
+Point-to-Point Synchronization Routine
+    & \begin{BVerbatim}
+shmem_wait_until_any(ivars, nelems, status, cmp,
+cmp_value):
+  for i in nelems:
+    R1 = LD ivars[i], my_pe
```
naveen-rn: isnt this blocking - atleast one should pass right?

content/api_observable.tex
naveen-rn: still a dead code - needs fix

content/shmem_fence.tex
```diff
@@ -36,8 +36,10 @@
 }
 
 \apinotes{
-    \FUNC{shmem\_fence} only provides per-\ac{PE} ordering guarantees and does not
-    guarantee completion of delivery.
+    \FUNC{shmem\_fence} orders prior remote operations with subsequent remote operations to the same \ac{PE}.
+    It also orders all prior normal memory accesses with subsequent remote observable accesses to any \ac{PE}.
```
naveen-rn, line 40: Same comment as @jeffhammond normal memory access isn't defined - please either define these terms or reword as prior local ld and st, or subsequent remote writes

naveen-rn (changes requested):
Overall, this is a much needed direction - beyond the editorial changes requested - there are some open semantic items. Adding them:

1. Undefined foundational terms - memory access and normal memory access - this is same as @jeffhammond's comment - this needs fix.
2. It looks like fence and quiet are now ordering loads - this needs to be called out in the changelog. I think - we need a larger discussion if this is not the intended change. Same as @jeffhammond 's `NOSTORE` comment - same issue as there but different context.
3. `SHMEM_CTX_NOSTORE` meaning is changed. Need discussion.
4. The text says the model is defined normative of C11 as an example - but we seem to omit comsume/seq_cst - this is an open for discussion - do we need ot mention why?
5. This is a confusion - i know we introduced changes in the spec that rejected using puts as a signal for operations and compatibility with wait/test removed. Now with asw - we are strictly enforcing this semantics. This is critical enough to be captured and made sure none of the examples are using the old semantics.
6. The model makes mixing SHMEM atomics and PL‑level atomics on the same location UB - a reseonable change, but it should be stated prominently as normative.
7. Collective needs more coverage - bcast missing from table, collect is modeled after uniform nelems - this needs fix.

naveen-rn:
@joalsop  Sorry missed a editorial blocker before - we need a graffle source for the images - we dont want to stick with just svg/png - this is a blocker - we have previously created graffle for all the images. Need some sort of source.

content/api_observable.tex
joalsop: argh - yes - it seems we were missing the broadcast header

content/api_observable.tex
joalsop: added while(1) at top

content/api_observable.tex
joalsop: don't know how I missed this earlier, sorry - replaced return line with num_found += 1

content/api_observable.tex
```diff
@@ -0,0 +1,540 @@
...
+Point-to-Point Synchronization Routine
+    & \begin{BVerbatim}
+shmem_test_some_vector(ivars, nelems, indices,
+status, cmp, cmp_values):
+  num_found =0
+  for i in nelems:
+    R1 = LD ivar, my_pe
+    R2 = LD cmp_values[i], my_pe
+    if cmp.test(R1, R2)
+      return i
```
joalsop: replace this also with num_found += 1

content/api_mem_model.tex
joalsop: good catch - amo_op should be added before the mo in that third disjunct

joalsop:
@naveen-rn really appreciate the detailed review - I'm in agreement about basically all the fixes you highlighted - a couple of additional responses below:

> 1. Undefined foundational terms - memory access and normal memory access - this is same as @jeffhammond's comment - this needs fix.

Now defined in PEs and operaiton events as follows: 

PL-level load, store, and RMW accesses (referred to here as "normal" memory accesses, to differentiate
from operation-generated observable accesses)...

> 2. It looks like fence and quiet are now ordering loads - this needs to be called out in the changelog. I think - we need a larger discussion if this is not the intended change. Same as @jeffhammond 's `NOSTORE` comment - same issue as there but different context.
> 3. `SHMEM_CTX_NOSTORE` meaning is changed. Need discussion.

This was discussed at general subcommittee meeting on 5/14 - a few points from that discussion:
* ordering isn't very clearly or consistently defined in the current spec, particularly for quiet
* ordering only stores rather than all memory accesses represents an incompatibility with the underlying PL-level memory model (extending it for the api would require significant changes breaking hb transitivity and/or differentiating the definition of sb for loads and stores)
* nobody was aware of any implementation that exploits this relaxation (I can't think of a way to lower it in a way that only orders prior stores in a c11 representation)
* the store-only description prevents an intuitive use case where pe0 loads some value, then syncs with pe1 (e.g., via a fence and atomic_set), then pe1 performs a put to the load location of pe0. it seems there would be no way to provide this ordering if quiet and fence only order stores.

Agreed that a changelog would be needed - should that be part of the PR? and if so, what version should be listed?

joalsop:
> @joalsop Sorry missed a editorial blocker before - we need a graffle source for the images - we dont want to stick with just svg/png - this is a blocker - we have previously created graffle for all the images. Need some sort of source.

@naveen-rn I think we updated to use svg and pdf images via draw.io based your prior feedback - I can look into converting svg to graffle - but let me know if you're aware of a good tool for this

manjugv:
@joalsop  @naveen-rn  @wfaderhold21  are these editorial level changes? or require special ballot ?

joalsop:
[REPORT_upload.md](https://github.com/user-attachments/files/31562093/REPORT_upload.md)

To highlight the importance of formal semantics for agentic coding, I asked an agent to implement a subset of openshmem calls in a couple of ways - and report on any areas of ambiguity that could lead to poor performance or errors just for these implementations. The attached is the result.

